PR #234, opened 2020-04-14
[sdf3] Changelog links to BitBucket backup

chapulina:


scpeters:
I wouldn't bother changing this branch since it is EOL

PR #233, opened 2020-04-13
Changelog links to BitBucket backup

chapulina:
Pull requests and commits point to the backup.

Issues point to this repo.

PR #240, opened 2020-04-14
[sdf9] Changelog links to BitBucket backup

chapulina:
Signed-off-by: Louise Poubel <louise@openrobotics.org>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-syntax: glob
 
 build
 build_*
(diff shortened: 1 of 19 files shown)


scpeters:
it looks like the formatting in this PR matches the changes you noted in https://github.com/osrf/sdformat/pull/233#issuecomment-613177960 that aren't yet included on master.

as I think about how to resolve these conflicts, it make me wonder if we are still going to merge branches forward? or will we use the cherry-pick strategy?

chapulina:
> it make me wonder if we are still going to merge branches forward? or will we use the cherry-pick strategy?

I don't have a strong preference. I've been merging forward `gazebo_ros_pkgs` because that's what I'm most comfortable with, but perhaps cherry-picking could be cleaner.

j-rivero:
@osrf-jenkins run tests

chapulina:
> it looks like the formatting in this PR matches the changes you noted in #233 (comment) that aren't yet included on master.

See #246

scpeters:
I just noticed an incorrect link in the API documentation and fixed it in c13f447

we should re-upload the API docs with this fix

PR #241, opened 2020-04-14
Port to TinyXML2

mjcarroll:
Resurrecting an older attempt to port SDFormat over to TinyXML2 as a basis for a new attempt.


mjcarroll:
This is going to take a bit more work to land on `master`, that's happening in #264

PR #243, opened 2020-04-15
README: Minor stylistic changes, add documentation links

EricCousineau-TRI:
* Discourages (or rather does not encourage) use of `sudo make [un]install` - it's a Bad Idea™
* Updates Markdown formatting
* Add explicit links to sdf_tutorials and sdformat.org

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# sdformat #
+# sdformat
 
 SDFormat is an XML file format that describes environments, objects, and robots
 in a manner suitable for robotic applications. SDFormat is capable of representing
@@ -7,37 +7,44 @@ or dynamic objects, and articulated robots with various sensors, and acutators.
 The format of SDFormat is also described by XML, which facilitates updates and
 allows conversion from previous versions.
 
-* SDFormat - The specification.
-    * SDF - Synonym for SDFormat, though SDFormat should be preferred, as "SDF"
-      is an acronym with other meanings.
-* libsdformat - The C++ parsing code contained within this repository,
+## Terminology
+
+* **SDFormat** - The specification.
+    * **SDF** - Synonym for SDFormat, though SDFormat should be preferred, as
+      "SDF" is an acronym with other meanings.
+* `libsdformat` - The C++ parsing code contained within this repository,
   which can be used to read SDFormat files and return a C++ interface.
 
-Test coverage:
+## Test coverage
 
 [![codecov](https://codecov.io/bb/osrf/sdformat/branch/default/graph/badge.svg)](https://codecov.io/bb/osrf/sdformat)
 
-
-## Installation ##
+## Installation
 
 Standard installation can be performed in UNIX systems using the following
 steps:
 
- - mkdir build/
- - cd build/
- - cmake ..
- - sudo make install
+```sh
+mkdir build/
+cd build/
+cmake ..  # Consider specifying -DCMAKE_INSTALL_PREFIX=...
+make install
+```
 
 sdformat supported cmake parameters at configuring time:
- - USE_INTERNAL_URDF (bool) [default False]
-   Use an internal copy of urdfdom 1.0.0 instead of look for one
-   installed in the system
- - USE_UPSTREAM_CFLAGS (bool) [default True]
-   Use the sdformat team compilation flags instead of the common set defined
-   by cmake.
 
-## Uninstallation ##
+* `USE_INTERNAL_URDF (bool) [default False]` <br/>
+  Use an internal copy of urdfdom 1.0.0 instead of look for one
+  installed in the system
+* `USE_UPSTREAM_CFLAGS (bool) [default True]` <br/>
+  Use the sdformat team compilation flags instead of the common set defined
+  by cmake.
+
+## Uninstallation
 
 To uninstall the software installed with the previous steps:
- - cd build/
- - sudo make uninstall
+
+```sh
+cd build/
+make uninstall
+```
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI:
Preview: https://github.com/EricCousineau-TRI/sdformat/blob/feature-2020-04-15-readme/README.md

README.md
```diff
@@ -7,6 +7,30 @@ or dynamic objects, and articulated robots with various sensors, and acutators.
...
 
+## Documentation
+
+See the [SDFormat Website](http://sdformat.org/) for:
+
+* [An outline of the SDFormat XML specification](http://sdformat.org/spec)
+* [Documentation on specification specifics](http://sdformat.org/tutorials?cat=specification)
+* [Proposals for modifications](http://sdformat.org/tutorials?cat=pose_semantics_docs)
+* [Developer instructions](http://sdformat.org/tutorials?tut=roadmap)
+  (e.g. installation and a roadmap)
```
chapulina, line 18: That page is a bit misleading to me, when I go to a roadmap, I'd expect to see a list of features planned for future releases. That page also doesn't have installation instructions, only version information. I think it should be called `releases`, similar to this one: https://ignitionrobotics.org/docs/all/releases

README.md
```diff
@@ -7,6 +7,30 @@ or dynamic objects, and articulated robots with various sensors, and acutators.
...
+
+<!--
+TODO(eric.cousineau): Add link to sdf_web_app once it's public.
+TODO(eric.cousineau): Add a link to API once later versions work.
+TODO(eric.cousineau): Add a link to Downloads once they're easier to navigate.
+TODO(eric.cousineau): Fix proposals to not use `pose_semantics_docs` as its
+name?
+TODO(eric.cousineau): Move installation instructions to sdf_tutorials, and link
+there?
+TODO(eric.cousineau): Move terminology section to sdf_tutorials?
```
chapulina, line 31: The last 3 sound reasonable to me, but maybe these TODOs should be point to issues on https://github.com/osrf/sdf_tutorials/ ?

README.md
EricCousineau-TRI: Aye, will remove it from this for now, and create an issue on `sdf_tutorials`.

EricCousineau-TRI:
What is the review policy for this repo? When do I squash? (I really like to make sure my commits are curated, esp. now that BitBucket is no longer being used.)

README.md
EricCousineau-TRI: Done.

README.md
EricCousineau-TRI: Done. Consolidated to namely direct towards

chapulina:
> What is the review policy for this repo? When do I squash?

We're still figuring these things out internally. Suggestions are very welcome.

EricCousineau-TRI:
I like the ROS2 guidelines (when using plain vanilla GitHub review), which is to always have fast-forward pushes for review, but then curate commits at the end, possibly after approval.
https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#pull-requests

Would that work for you?

EricCousineau-TRI:
Ooh, also, @scpeters and @azeey may try out Reviewable (example draft PR here: https://github.com/osrf/sdf_tutorials/pull/3). This is perhaps not actionable now, but I could let you know how that goes for us?

chapulina:
> always have fast-forward pushes for review, but then curate commits at the end, possibly after approval.

That sounds reasonable to me, we'll discuss this as a team next week.

> Reviewable

We're also planning to discuss this. My main concern is the recommendation to not mix the GitHub and Reviewable interfaces. Do you have any experience mixing both?

The entire team at Open Robotics should be able to open and review SDFormat PRs. We should also strive to have a consistent workflow for the entire Ignition family. So it sounds like we either use Reviewable for all projects or none.

EricCousineau-TRI:
> That sounds reasonable to me, we'll discuss this as a team next week.

Sounds good. For this PR, I will wait for Approval, then curate my commits.

> [...] Do you have any experience mixing both?

We just tried it out here: https://github.com/osrf/sdf_tutorials/pull/3#pullrequestreview-394909755
Basically, it works from GitHub to Reviewable (i.e. I see a new discussion in Reviewable, started by the GitHub item). However, once I responded in Reviewable, it did not go back to the GitHub discussion, as we've noted here for Drake:
https://drake.mit.edu/reviewable.html#github-integration
I think that can easily be solved by just referencing it.

The main issue we've had is mixing the request of review between GitHub (assigning reviewers in the sidebar) vs Reviewable (`+@username`). It does not always recognize GitHub assignment, so we generally stick to Reviewable assignments. [*]

So overall, the mixing between GitHub and Reviewable is not the most convenient, but may be workable.
Additionally, once we've become accustomed to Reviewable, it's wayyy easier to know that everything has been addressed in the review (and really easy to get the most de-cluttered view of it), with minimal effort of going back and looking through conversation history or what not in GitHub. The excessive client-side stuff is a bit annoying, but eh, overall I say worth it.

Understandably, this may not work for y'all, but I would recommend people taking it out for a test run. It's easy to point Reviewable to a PR to try it out on.

---

[*] We also have the following completion script, so it's unclear if this is due to our custom completion script (meant to handle labels and such):
https://github.com/RobotLocomotion/drake-ci/wiki/reviewable.io-review-completion-condition

EricCousineau-TRI:
Are there any action items remaining here?

EricCousineau-TRI:
Squashed to two curated commits

EricCousineau-TRI:
And thanks!

README.md
scpeters: it sounds like she was asking for a link to issues in the TODO comments?

ie: https://github.com/osrf/sdf_tutorials/issues/1

I don't think it hurts to have the terminology in both places. So I wouldn't "Move terminology section to sdf_tutorials" but rather copy it there. Duplicate info isn't great, it's short enough, and I think it's helpful in both places

PR #244, opened 2020-04-15
Make parsing of values syntactically more strict with bad values generating an error

### LINKED ISSUE #228: Diagnostics for invalid numeric input
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/228)) by Jeremy Nimmer (Bitbucket: [Jeremy Nimmer](https://bitbucket.org/%7B127cef5b-add8-4be7-9175-2c7214f56bcf%7D/)).**

\# Summary

When a user makes a mistake and provides invalid data, e.g., for a \(link pose\)\[[http://sdformat.org/spec?ver=1.6&elem=link#link\_pose](http://sdformat.org/spec?ver=1.6&elem=link#link_pose)\], the sdformat parser should provide a diagnostic.

\# Motivation

In many cases \`\*.sdf\` files are authored by hand, possibly even with xacro or related tools.  It is easy to accidentally make a typo and end up with a file that appears to be valid, but in fact will behave poorly in simulation for some unknown reason, e.g., if `1.O1e2` is parsed as "1." instead of "101.".

\# Describe alternatives you've considered

I would appreciate if the parser provided an API channel for errors, but I believe this isn't available yet – even for mistakes that it does provide diagnostics for.  Starting out with log printouts for numeric typos would be a step in the right direction.

\# Additional context

Here is an example stanza that does not produce any diagnostics as of tag/sdformat9\_9.1.0:

```xml
      <collision name='link1_collision1'>
        <pose>foo foo foo foo foo foo foo foo -1e899999</pose>
        <geometry>
          <plane>
            <normal>1.0 2.0 3.0 foo</normal>
          </plane>
        </geometry>
        <surface>
          <friction>
            <ode>
              <mu>0.8 gravitons</mu>
              <mu2>0.3 muons</mu2>
            </ode>
          </friction>
        </surface>
      </collision>
```

‌

azeey:
Currently, libsdformat silently ignores syntax errors when parsing values. For example,  `<pose>bad 0 0 0 0 0</pose>` or `<pose>0.1 0.2 0.3 0.4 0.5</pose>` are both bad values for the `<pose>` tag, but the parsed values are `<pose>0 0 0 0 0 0</pose>` and `<pose>0.1 0.2 0.3 0.4 0.5 0</pose>` respectively. With this PR, libsdformat will generate an error for such values.

The plan is to make a similar PR for libsdformat9, but printing a warning message instead.

This resolves #228

### DIFF
--- a/src/Param.cc
+++ b/src/Param.cc
@@ -379,7 +379,7 @@ bool Param::ValueFromString(const std::string &_value)
         StringStreamClassicLocale ss(tmp);
         ss >> colortmp;
       }
-      catch(const std::ios_base::failure &_e)
+      catch(const std::ios_base::failure &)
       {
         StringStreamClassicLocale ss(tmp + " " + std::to_string(colortmp.A()));
         ss >> colortmp;
@@ -464,7 +464,7 @@ bool Param::ValueFromString(const std::string &_value)
            << this->dataPtr->key << "].\n";
     return false;
   }
-  catch(const std::ios_base::failure &_e)
+  catch(const std::ios_base::failure &)
   {
     sdferr << "Unknown error. Unable to set value ["
            << _value << " ] for key["
--- a/src/Param.cc
+++ b/src/Param.cc
@@ -45,14 +45,12 @@ namespace sdf
     public: explicit StringStreamClassicLocale()
     {
       this->imbue(std::locale::classic());
-      this->exceptions(std::stringstream::failbit);
     }
 
     public: explicit StringStreamClassicLocale(const std::string& str)
       : std::stringstream(str)
     {
       this->imbue(std::locale::classic());
-      this->exceptions(std::stringstream::failbit);
     }
   };
   }
@@ -275,6 +273,29 @@ std::string Param::GetDefaultAsString() const
 }
 
 //////////////////////////////////////////////////
+/// \brief Helper function for Param::ValueFromString
+/// \param[in] _input Input string.
+/// \param[in] _key Key of the parameter, used for error message.
+/// \param[out] _value This will be set with the parsed value.
+/// \return True if parsing succeeded.
+template <typename T>
+bool ParseUsingStringStream(const std::string &_input, const std::string &_key,
+                            ParamPrivate::ParamVariant &_value)
+{
+  StringStreamClassicLocale ss(_input);
+  T _val;
+  ss >> _val;
+  if (ss.fail())
+  {
+    sdferr << "Unknown error. Unable to set value [" << _input << " ] for key["
+           << _key << "]\n";
+    return false;
+  }
+  _value = _val;
+  return true;
+}
+
+//////////////////////////////////////////////////
 bool Param::ValueFromString(const std::string &_value)
 {
   // Under some circumstances, latin locales (es_ES or pt_BR) will return a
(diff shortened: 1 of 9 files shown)


j-rivero:
@osrf-jenkins run tests

src/Param.cc
```diff
@@ -440,6 +464,13 @@ bool Param::ValueFromString(const std::string &_value)
            << this->dataPtr->key << "].\n";
     return false;
   }
+  catch(const std::ios_base::failure &_e)
```
EricCousineau-TRI, line 467: nit Also possible to handle using `ss.fail()`?

EricCousineau-TRI: OK Per convo with Addisu, it's throwing exceptions, so nah, can't do this.

EricCousineau-TRI:
meta comment: https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

scpeters:
I think there are some test files missing that should start with `bad_syntax_*`

src/Param.cc
```diff
@@ -403,11 +417,21 @@ bool Param::ValueFromString(const std::string &_value)
...
-      ss >> posetmp;
-      this->dataPtr->value = posetmp;
+      if (!tmp.empty() || !std::all_of(tmp.begin(), tmp.end(), isWhiteSpace))
+      {
+        StringStreamClassicLocale ss(tmp);
+        ignition::math::Pose3d posetmp;
+
+        ss >> posetmp;
+        this->dataPtr->value = posetmp;
+      }
```
scpeters, line 434: I think the changes to this block are not needed because [Param::SetFromString](https://github.com/osrf/sdformat/blob/stricter_parsing_10/src/Param.cc#L481-L493) already calls `sdf::trim` to remove whitespace and then sets a default value if the trimmed string is empty. I've added tests to confirm this in 9daad39e603fa77987689b03070167b924fd0e3a of branch `scpeters/stricter_parsing_10` after reverting the changes to this block in 8f6120c

azeey:
> I think there are some test files missing that should start with `bad_syntax_*`

Oops! I've added them in 2b42034

scpeters:
there are test failures on macOS and windows; I've investigated the macOS failure, and it seems that the parsing the bad pose file doesn't generate an exception with the CI version of clang / libc++

src/Param.cc
azeey: Sounds good. I see that `ValueFromString` is also called form the constructor without `sdf::trim`, but that should be okay because it's using the `default` string from the spec description.
https://github.com/osrf/sdformat/blob/b80e070dbed613e972acfb81e66b24a248eefd6c/src/Param.cc#L70

azeey: I just tried it and it looks like `UNIT_Utils_TEST` fails with that change.

scpeters: ok, it sounds like this change is a good idea. Perhaps we could generalize it so it applies to types other than Pose?

src/Param.cc
azeey: Using `fail()` now because macOS doesn't throw an exception

src/Param.cc
azeey: I added an overload for `sdf::trim` that takes a `std::string` in https://github.com/osrf/sdformat/pull/244/commits/fdecf77b19ee59c062602b2649737b4da2d9a7e4, and used that to trim input values in https://github.com/osrf/sdformat/pull/244/commits/32b3d84ef305ff1f20e09eeab6672dbe6e2d0f68

scpeters:
do you want to target this at `sdf10`?

azeey:
> do you want to target this at `sdf10`?

Done.

PR #246, opened 2020-04-17
[master] Update BitBucket links

chapulina:
Follow-up to #233, improved the script to also update files other than the changelog.

There are a couple of links to `osrf/gazebo` which won't work until that's migrated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+
 build
 build_*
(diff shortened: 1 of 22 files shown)


scpeters:
I'm going to rebase and merge without squashing since is bringing commits for #251 and #268 forward from sdf9

PR #235, opened 2020-04-14
[sdf4] Changelog links to BitBucket backup

chapulina:
Signed-off-by: Louise Poubel <louise@openrobotics.org>

PR #248, opened 2020-04-17
[sdf4] Update BitBucket links

chapulina:
Follow up to #235.

The `osrf/gazebo` links won't work until that migration has been completed.


scpeters:
@osrf-jenkins run tests

Issue #230, opened 2020-04-09
Return positive infinity instead of -1 for boundless positive values


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/230)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

----------------------------------------
## Summary

This was initially submitted as [pull request #458](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/458) by @mxgrey, and is a worthwhile change that we'd like to merge in libsdformat10, but it needs some tests. We can also add some tests to libsdformat9 that confirm that `inf` / `Inf` / `INF` and +- variants can be parsed correctly in numerical fields. Due to the incoming bitbucket migration, the pull request was closed and should be re-opened when tests are available.

## Original PR description:

SDFormat has a convention of using -1 to represent positive infinity. If users of the DOM API are not careful and cognizant of this convention, it can result in confusing bugs that are hard to track down. My dartsim plugin was exhibiting subtle non-physical behavior after loading structures from SDF, because it had effort and velocity limits of +1 (minimum) and -1 (maximum).

This PR is a suggested tweak to the API behavior so that the API returns positive infinity instead of -1 to represent boundlessly positive values. Hopefully this might save other SDF users from long, confusing debugging sessions.

However, this does deviate from the exact convention of the SDF document specification, so I’d understand if this PR is declined.

EricCousineau-TRI:
Relates #169

scpeters:
this was merged to `sdf10` in https://github.com/osrf/sdformat/pull/357

PR #249, opened 2020-04-18
Return positive infinity instead of -1 for boundless positive values

scpeters:
Revived from [bitbucket PR 458](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/458) per #230. Tests are still needed.

## Original description:

SDFormat has a convention of using -1 to represent positive infinity. If users of the DOM API are not careful and cognizant of this convention, it can result in confusing bugs that are hard to track down. My dartsim plugin was exhibiting subtle non-physical behavior after loading structures from SDF, because it had effort and velocity limits of +1 (minimum) and -1 (maximum).

This PR is a suggested tweak to the API behavior so that the API returns positive infinity instead of -1 to represent boundlessly positive values. Hopefully this might save other SDF users from long, confusing debugging sessions.

However, this does deviate from the exact convention of the SDF document specification, so I’d understand if this PR is declined.


scpeters:
targeting this to `sdf10` instead in #357

PR #254, opened 2020-04-24
Add github action CI for uploading to codecov

scpeters:


scpeters:
I was hoping this would start an action, but maybe it has to be merged to master first?

scpeters:
I'll submit this from `origin` instead of my fork

PR #255, opened 2020-04-25
Add github action CI for uploading to codecov

scpeters:
Follow-up to #254 submitted from `origin`


scpeters:
ruby is inserting the following text into all printouts from the `ign` tool, which causes a `UNIT_ign_TEST` failure:

~~~
/usr/lib/ruby/vendor_ruby/rubygems/defaults/operating_system.rb:10: warning: constant Gem::ConfigMap is deprecated
~~~

scpeters:
~~~
lcov: ERROR: no valid records found in tracefile coverage.info
~~~

scpeters:
I'm able to reproduce the problem with `lcov` after upgrading gcov to 7.5.0

scpeters:
ok I think this is finally working and ready for review

scpeters:
after this is merged, we can consider adding coverage checks to the build status

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,76 @@
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ '*' ]
```
chapulina, line 5: I *think* the `branches` are optional in this case.

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,76 @@
+name: C/C++ CI
```
chapulina, line 1: Consider adding Bionic / Linux to the name

scpeters: fb13a87836cc668df9275cd7b17e3baf3d289c07

.github/workflows/c-cpp.yml
scpeters: yep, I simplified the syntax in fb13a87836cc668df9275cd7b17e3baf3d289c07

chapulina:
Oh I just realized we should remove this file:

https://github.com/osrf/sdformat/blob/c6dac91c7845b0c45611e4db338e240bd86049d8/codecov.yml#L1-L2

It was only needed with mercurial because the main branch defaults to `master`.

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,72 @@
+name: Ubuntu Bionic / Linux
```
chapulina, line 1: The name of this file, `c-cpp.yml` could also be more descriptive, to differentiate well from other actions that may come to this folder.

chapulina:
While porting this to Ignition, I thought it would be nice to reduce duplication, so I came up with an alternative approach on #266. Let me know what you think.

.github/workflows/c-cpp.yml
scpeters: 19e4896

scpeters:
> Oh I just realized we should remove this file:

removed in a4fc1d7

PR #257, opened 2020-04-28
Merge sdf4 -> sdf6

scpeters:
Includes changes from #248.

I'm trying out the workflow of merging changes forward.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-syntax: glob
 
 build
 build_*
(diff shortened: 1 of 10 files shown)


scpeters:
> How did you do it? I see there's only a single commit authored by you. When we start porting lots of changes, it would be nice to preserve some of the history.

yeah, that's a good point. I did the following:

~~~
git checkout sdf4
git checkout sdf6
git merge sdf4
# then used the following to resolve conflicts
git mergetool
~~~

I'll make a parallel PR using cherry-pick to see if that's any better

scpeters:
see #258 for an alternative approach

scpeters:
merged #258 instead

PR #259, opened 2020-04-29
Enforce minimum/maximum values specified in SDFormat description files

azeey:
Currently, some SDFormat description files contain `min` and `max` values, but the parser does not enforce those values in any way. This PR adds a validation step in the parser that checks if scalar values are within the allowed range. Note that this PR addresses only element values, not attributes.

### DIFF
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -174,14 +174,29 @@ namespace sdf
                               const std::string &_description="");
 
     /// \brief Add a value to this Element.
-    /// \param[in] _type Type of data the attribute will hold.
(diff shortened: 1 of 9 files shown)


EricCousineau-TRI:
Per f2f, @azeey will retarget to libsdformat10.

azeey:
I rebased onto sdf10 and changed the base of this PR to sdf10. Not sure why Github is saying there are conflicting files.

EricCousineau-TRI:
Per VC, DCO and GitHub are being weird about Addisu having changed the target branch from `sdf9` to `sdf10`.

Addisu may try to just open a new PR.

PR #264, opened 2020-04-30
Migrate to using TinyXML2

mjcarroll:
This is the forward port of #241 for the current master branch `sdf10`.


scpeters:
there are some conflicts here, partially due to deprecation of the API's in parser_urdf.hh in the following bitbucket PR:

* https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/658/page/1

@azeey, we marked it as deprecated as of 9.2, and I think we are planning to remove it in 10.0? let's make a separate PR to move that header file from `include` to `src`, and then it will clarify the API impact of this PR

scpeters:
PR to not install parser_urdf.hh: https://github.com/osrf/sdformat/pull/276

EricCousineau-TRI:
@mjcarroll Do you know if you need a stable release for Ignition Dome (`libsdformat10`), or will y'all be fine with using `master`?

Asking in the context of #278, whether we should target `libsdformat11` instead of `libsdformat10`.

mjcarroll:
@EricCousineau-TRI I would like to target `dome`.

EricCousineau-TRI:
OK, thanks!

scpeters:
do you mind rebasing on top of `sdf10` and changing the target of this PR?

* https://github.com/osrf/sdformat/commits/sdf10

mjcarroll:
> do you mind rebasing on top of sdf10 and changing the target of this PR?

Done! Thanks for getting that set up.

scpeters:
this PR is huge, but I'm not sure if there's a way to split up the changes. once you start parsing with tinyxml2, all the functions are going to need to read those data structures

mjcarroll:
> this PR is huge, but I'm not sure if there's a way to split up the changes.

One thing to reduce the size could be to do the conversion, and then the removal of the vendored TinyXML in a followup PR

src/parser.cc
```diff
@@ -754,16 +755,16 @@ std::string getModelFilePath(const std::string &_modelDirPath)
     }
   }
 
-  TiXmlDocument configFileDoc;
-  if (!configFileDoc.LoadFile(configFilePath))
+  tinyxml2::XMLDocument configFileDoc;
+  if (!configFileDoc.LoadFile(configFilePath.c_str()))
```
azeey, line 759: This should actually be the opposite logic. Consider comparing the return value against tinyxml2::XML_SUCCESS. Changing this to
```suggestion
  if (tinyxml2::XML_SUCCESS != configFileDoc.LoadFile(configFilePath.c_str()))
```
fixes all the test failures.

src/XmlUtils.hh
```diff
@@ -0,0 +1,42 @@
...
+  /// \param[in] _doc Document in which to place the copied node
+  /// \param[in] _src The node to deep copy
+  /// \returns The newly copied node
+  tinyxml2::XMLNode* DeepClone(tinyxml2::XMLDocument *_doc,
+                               const tinyxml2::XMLNode *_src);
+
+  /// \brief Trim whitespace from a string
+  /// \param[in] _s String to trim
+  /// \returns String _s with whitespace trimmed from both ends
+  std::string TrimString(const std::string &_s);
```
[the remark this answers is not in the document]
mjcarroll: It can, when modified to remove newlines `\n` as well.  Done in 63318.

mjcarroll:
@scpeters How do I go about updating `sdformat-ci-pr_any-ubuntu_auto-amd64` to install tinyxml2?

mjcarroll:
It appears that the example is failing due to a trailing quote in the `simple.sdf` file: https://github.com/osrf/sdformat/blob/master/examples/simple.sdf#L6

Is this intentional, or is it that TinyXML2 happens to be a bit more strict?

scpeters:
> How do I go about updating `sdformat-ci-pr_any-ubuntu_auto-amd64` to install tinyxml2?

the logic for the ubuntu build is in release-tools:

* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/lib/dependencies_archive.sh#L55-L107

for homebrew, we'll need a new formula with different dependencies; I'll work on that

scpeters:
> for homebrew, we'll need a new formula with different dependencies; I'll work on that

https://github.com/osrf/homebrew-simulation/pull/1054

scpeters:
> example is failing due to a trailing quote in the `simple.sdf` file

that was definitely not intentional; thanks for fixing it

scpeters:
> > for homebrew, we'll need a new formula with different dependencies; I'll work on that
> 
> [osrf/homebrew-simulation#1054](https://github.com/osrf/homebrew-simulation/pull/1054)

osrf/homebrew-simulation#1054 is merged

scpeters:
@osrf-jenkins run tests

src/parser_urdf.cc
```diff
@@ -2908,9 +2931,9 @@ void CreateInertial(TiXmlElement *_elem,
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-void CreateJoint(TiXmlElement *_root,
+void CreateJoint(tinyxml2::XMLElement *_root,
                  urdf::LinkConstSharedPtr _link,
-                 ignition::math::Pose3d &/*_currentTransform*/)
+                 ignition::math::Pose3d &_currentTransform)
```
scpeters, line 2936: I would revert the change on this line to keep `_currentTransform` commented out since it is generating an unused variable clang warning on the homebrew CI

scpeters: [Build Status](https://build.osrfoundation.org/job/sdformat-ci-pr_any-homebrew-amd64/1312/) https://build.osrfoundation.org/job/sdformat-ci-pr_any-homebrew-amd64/1312/warnings10Result/

mjcarroll: Done

scpeters:
the homebrew CI is working, but it's failing to compile on Ubuntu, any ideas?

mjcarroll:
> the homebrew CI is working, but it's failing to compile on Ubuntu, any ideas?

I believe it's because we force using the internal urdfdom on the Jenkins CI, I'll get that fixed up.

src/XmlUtils.cc
```diff
@@ -0,0 +1,60 @@
...
+inline namespace SDF_VERSION_NAMESPACE {
+
+/////////////////////////////////////////////////
+tinyxml2::XMLNode* DeepClone(tinyxml2::XMLDocument *_doc,
+                             const tinyxml2::XMLNode *_src)
+{
+  if (_src == nullptr)
+  {
+    sdferr << "Pointer to XML node _src is NULL\n";
+    return NULL;
```
scpeters, line 34: nit: `NULL` -> `nullptr`

src/XmlUtils.hh
```diff
@@ -0,0 +1,40 @@
...
+
+namespace sdf
+{
+  // Inline bracket to help doxygen filtering.
+  inline namespace SDF_VERSION_NAMESPACE {
+
+  /// \brief Perform a deep copy of an XML Node
+  /// \param[in] _doc Document in which to place the copied node
+  /// \param[in] _src The node to deep copy
+  /// \returns The newly copied node
```
scpeters, line 35: discuss error behavior? could add " or nullptr if an error occurs" or something similar

src/urdf/urdf_parser/joint.cpp
```diff
@@ -575,71 +575,71 @@ bool parseJoint(Joint &joint, TiXmlElement* config)
...
   }
   return true;
 }
 
-bool exportJoint(Joint &joint, TiXmlElement* xml)
+bool exportJoint(Joint &joint, tinyxml2::XMLElement* xml)
 {
-  TiXmlElement * joint_xml = new TiXmlElement("joint");
-  joint_xml->SetAttribute("name", joint.name);
+  tinyxml2::XMLElement * joint_xml = xml->GetDocument()->NewElement("joinrt");
```
scpeters, line 641: typo: `joinrt` should be `joint`

scpeters: this makes me question our test coverage

src/parser_urdf.cc
```diff
@@ -1355,46 +1369,46 @@ void URDF2SDF::ParseSDFExtension(TiXmlDocument &_urdfXml)
...
         sdf->minDepth = std::stod(GetKeyValueAsString(childElem));
       }
-      else if (childElem->ValueStr() == "mu1")
+      else if (strcmp(childElem->Name(), "mu1") == 0)
       {
         sdf->isMu1 = true;
         sdf->mu1 = std::stod(GetKeyValueAsString(childElem));
       }
-      else if (childElem->ValueStr() == "mu2")
+      else if (strcmp(childElem->Value(), "mu2") == 0)
```
scpeters, line 1392: similar to other comment, use `Name()` instead of `Value()` here (and in the rest of this function)?

src/parser_urdf.cc
```diff
@@ -2568,17 +2593,14 @@ void ReduceSDFExtensionFrameReplace(SDFExtensionPtr _ge,
     ReduceSDFExtensionProjectorFrameReplace(blobIt, _link);
     ReduceSDFExtensionGripperFrameReplace(blobIt, _link);
     ReduceSDFExtensionJointFrameReplace(blobIt, _link);
-
-    std::ostringstream debugStreamOut;
-    debugStreamOut << *(*blobIt);
```
scpeters, line 2573: well this wasn't doing anything

src/parser_urdf.cc
```diff
@@ -3267,44 +3299,41 @@ TiXmlDocument URDF2SDF::InitModelString(const std::string &_urdfStr,
...
     catch(...)
     {
-      delete sdf;
       throw;
     }
   }
   catch(...)
   {
-    delete robot;
     throw;
```
scpeters, line 3307: I see the `delete` is removed here and in the previous `catch` block; does this mean there will be memory leaks if either of these `catch` blocks  are executed?

src/parser_urdf.cc
```diff
@@ -3267,44 +3299,41 @@ TiXmlDocument URDF2SDF::InitModelString(const std::string &_urdfStr,
...
+void URDF2SDF::InitModelFile(const std::string &_filename,
+                             tinyxml2::XMLDocument *_sdfXmlDoc)
 {
-  TiXmlDocument xmlDoc;
-  if (xmlDoc.LoadFile(_filename))
+  tinyxml2::XMLDocument xmlDoc;
+  if (!xmlDoc.LoadFile(_filename.c_str()))
   {
-    return this->InitModelDoc(&xmlDoc);
+    return this->InitModelDoc(&xmlDoc, _sdfXmlDoc);
```
scpeters, line 3330: this is a `void` function; perhaps move `this->InitModelDoc(&xmlDoc, _sdfXmlDoc);` to the previous line?

src/XmlUtils.cc
mjcarroll: f21871e8

src/XmlUtils.hh
mjcarroll: f21871e8

src/urdf/urdf_parser/joint.cpp
mjcarroll: Based on a quick look, I don't think that we have any test coverage of the internal URDF parser in this direction (this is exporting).  Instead, we are only using the internal parser to read URDF into SDF.

src/urdf/urdf_parser/link.cpp
```diff
@@ -226,18 +226,18 @@ bool parseMesh(Mesh &m, TiXmlElement *c)
...
 
-  TiXmlElement *shape = g->FirstChildElement();
+  tinyxml2::XMLElement *shape = g->FirstChildElement();
   if (!shape)
   {
     return geom;
   }
 
-  std::string type_name = shape->ValueStr();
+  std::string type_name = shape->Value();
```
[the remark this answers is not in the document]
mjcarroll: 9c9b74a7

src/parser_urdf.cc
mjcarroll: 9c9b74a7

src/urdf/urdf_parser/joint.cpp
mjcarroll: 9c9b74a7

src/parser_urdf.cc
mjcarroll: It shouldn't be leaked because the corresponding `new` was also removed.  Instead, it's created via the `_sdfXmlOut->NewElement()` call on link 3224.

This is in line with the TinyXML2 documentation:

> An XMLDocument is a C++ object like any other, that can be on the stack, or new'd and deleted on the heap.
> However, any sub-node of the Document, XMLElement, XMLText, etc, can only be created by calling the appropriate XMLDocument::NewElement, NewText, etc. method. Although you have pointers to these objects, they are still owned by the Document. When the Document is deleted, so are all the nodes it contains.

src/parser_urdf.cc
mjcarroll: I'll just drop the return from here, it's unnecessary.

src/urdf/urdf_parser/joint.cpp
scpeters: >> this makes me question our test coverage

> Based on a quick look, I don't think that we have any test coverage of the internal URDF parser in this direction (this is exporting). Instead, we are only using the internal parser to read URDF into SDF.

yeah, this wasn't a critique of this PR, just an overall comment

src/parser_urdf.cc
scpeters: 👍

src/urdf/urdf_parser/joint.cpp
mjcarroll: It looks like there are `urdfdom` unit tests, but they aren't present here? https://github.com/ros2/urdfdom/blob/ros2/urdf_parser/test/urdf_unit_test.cpp

scpeters:

scpeters:
the build farm is backed up, but I expect it to pass, so I'll approve

let's wait to merge until the last job finishes

scpeters:
ok, it looks it just needs an update to the migration guide and changelog, and then we can squash and merge

@mjcarroll do you want me to update these?

scpeters:
> ok, it looks it just needs an update to the migration guide and changelog, and then we can squash and merge
> 
> @mjcarroll do you want me to update these?

updated in 7155942

scpeters:
updating dependency in release repo: https://github.com/ignition-release/sdformat10-release/pull/1

PR #270, opened 2020-05-07
Simplify data embedding

jwnimmer-tri:
This is an internal-only refactoring that changes the mechanics of how the `sdf/**` file data is embedded into the library.

For one, it solves the https://isocpp.org/wiki/faq/ctors#static-init-order fiasco for the embedded data.  The data is only loaded into memory upon first use, instead of as the library is being loaded.  This is the magic of function-local statics.

It also simplifies the ruby embedding code, making it more concise.  I hope this is easier to maintain, but it also helps me when sharing Drake with consumers that always build from source, but will not install ruby as a compile-time prerequisite (https://github.com/RobotLocomotion/drake/issues/13231).

I am happy to take feedback.  Is this the correct branch to PR into?  Should I be adding a changelog entry?  You are welcome to push fixes to my branch -- it looks like by convention I should not be rebasing nor squashing it?

### DIFF
--- a/src/Converter.cc
+++ b/src/Converter.cc
@@ -29,11 +29,17 @@
 #include "sdf/Types.hh"
 
 #include "Converter.hh"
-
-// This include file is generated at configure time.
-#include "sdf/EmbeddedSdf.hh"
+#include "EmbeddedSdf.hh"
 
 using namespace sdf;
+
+namespace {
+bool EndsWith(const std::string& _a, const std::string& _b)
+{
+  return (_a.size() >= _b.size()) &&
+      (_a.compare(_a.size() - _b.size(), _b.size(), _b) == 0);
+}
+}
 
 /////////////////////////////////////////////////
 bool Converter::Convert(TiXmlDocument *_doc, const std::string &_toVersion,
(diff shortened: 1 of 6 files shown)


src/Converter.cc
```diff
@@ -29,12 +29,18 @@
...
+#include "EmbeddedSdf.hh"
 
 using namespace sdf;
 
+namespace {
+bool EndsWith(const std::string& a, const std::string& b)
+{
+  return (a.size() >= b.size()) &&
+      (a.compare(a.size() - b.size(), b.size(), b) == 0);
+}
```
azeey, line 41: Style: function parameters should start with underscore 
```suggestion
bool EndsWith(const std::string &_a, const std::string &_b)
{
  return (_a.size() >= _b.size()) &&
          (_a.compare(_a.size() - _b.size(), _b.size(), _b) == 0);
}

```

src/Converter.cc
```diff
@@ -73,42 +79,47 @@ bool Converter::Convert(TiXmlDocument *_doc, const std::string &_toVersion,
...
+  std::string curVersion = origVersion;
+  while (curVersion != _toVersion)
   {
-    // Get the SDF to version.
-    toVer = fromIter->second.first;
+    // Find the (at most one) file named, e.g., ".../1_7.convert".
+    std::string snakeVersion = curVersion;
+    std::replace(snakeVersion.begin(), snakeVersion.end(), '.', '_');
+    const std::string suffix = "/" + snakeVersion + ".convert";
+    const char* convert_xml = nullptr;
```
azeey, line 94: nit: `convert_xml` -> `convertXml`

jwnimmer-tri:
FYI I've locally squashed and force-pushed as part of placating the DCO bot

> I wonder how hard it would be to generate the file using CMake.

Probably possible, but there's a more complicated use of ruby still in tools/, so we wouldn't be able to drop the ruby dependency yet.  That could be future work.  For now, this is enough to make my users happier.

src/EmbeddedSdf.hh
```diff
@@ -0,0 +1,40 @@
...
+  inline namespace SDF_VERSION_NAMESPACE {
+  //
+
+  /// \internal
+
+  /// A map where the keys are a source-relative pathnames within the "sdf"
+  /// directory such as "1.8/root.sdf", and the values are the contents of
+  /// that source file.
+  const std::map<std::string, std::string> &GetEmbeddedSdf();
+}
```
EricCousineau-TRI, line 38: nit Mismatched indentation?

scpeters: it looks like we usually close the internal namespace on the same line that it was created ([Converter.hh](https://github.com/osrf/sdformat/blob/master/src/Converter.hh#L30-L112)) but not always ([SDFImplPrivate.hh](https://github.com/osrf/sdformat/blob/master/src/SDFImplPrivate.hh#L29-L51)), so I'm not worried about it

PR #276, opened 2020-05-12
Don't install deprecated parser_urdf.hh header

scpeters:
The `URDF2SDF` class in `parser_urdf.hh` was deprecated in [bitbucket pr #658](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/658/page/1), and I believe the plan was to remove it from the public API in `libsdformat10`, so this PR removes it from the list of installed headers and moves it to the `src` folder.

I included two additional changes:

* 5e99d80: I removed the deprecation macros from the private header as well as the warning suppression macros.
* 04f84ec: fix a cmake warning about multiline strings (see also https://github.com/osrf/gazebo/issues/2664 and [bitbucket gazebo pr #3138](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/3138/page/1))

~~I did not remove the `SDFORMAT_VISIBLE` macro for `URDF2SDF` as it was starting to take a while to figure out how to fix it, so I just left that part out.~~ I removed the `SDFORMAT_VISIBLE` macro in b763472.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,9 @@
 
 ### libsdformat 10.0.0 (202X-XX-XX)
 
+1. Don't install deprecated parser_urdf.hh header file, fix cmake warning about newline file, fix cmake warning about newlines.
+    * [Pull request 276](https://github.com/osrf/sdformat/pull/276)
+
 1. Changed the default radius of a Cylinder from 1.0 to 0.5 meters.
     * [BitBucket pull request 643](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/643)
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -20,6 +20,15 @@
 
 ### libsdformat 9.X.X (202X-XX-XX)
 
+1. Change bitbucket links to GitHub.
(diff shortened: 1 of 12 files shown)


scpeters:
the code_check script apparently doesn't like the visibility macro still being there:

*  [Build Status](https://build.osrfoundation.org/job/sdformat-ci-pr_any-ubuntu_auto-amd64/995/) https://build.osrfoundation.org/job/sdformat-ci-pr_any-ubuntu_auto-amd64/995/cppcheckResult/

scpeters:
> For the symbol visibility issue, I think parser_urdf_TEST.cc needs to be compiled with parser_urdf.cc and SDFExtention.cc similar to what we're doing for Convter_TEST.cc. I think it will also need to link against liburdf*.

good call; I was stuck before, but I think it's working now as of b763472

scpeters:
there were code check complaints about unused private functions, so I made them public (within the private header) and used them in a unit test in 330e6a5

scpeters:

azeey:
Looks good!

scpeters:
I just squashed and will "rebase and merge" once CI clears

PR #295, opened 2020-06-02
Bump master branch to major version 11

scpeters:
We are planning to release `libsdformat10` with dome in September, but save the SDFormat 1.8 specification changes for December. Since the master branch already has some 1.8 specification files, bump this branch to major version 11 and create an `sdf10` branch that doesn't have the 1.8 specification changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ if(UNIX AND NOT WIN32)
  endif(CMAKE_UNAME)
 endif()
 
-project (sdformat10)
+project (sdformat11)
 
 # The protocol version has nothing to do with the package version set below.
 # It represents the current version of sdformat implement by the software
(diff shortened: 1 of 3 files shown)


scpeters:
> Can you update the example to look for libsdformat11? Otherwise, LGTM

0af0e998b7f3cd6b7aab8c3de9b0302092a305af

scpeters:
> LGTM.
> 
> For completeness, Ignition Dome coming up on September 2020 will use SDFormat 10, and Ignition-E coming up on March 2021 will use SDFormat 11.
> 
> This will also mean a bump of all downstream libraries for both Ignition collections. It would be good to bump Dome dependencies ASAP so we can stay in sync with SDF 10.

yes this is true but it still feels 😞

scpeters:
> This will also mean a bump of all downstream libraries for both Ignition collections. It would be good to bump Dome dependencies ASAP so we can stay in sync with SDF 10.

👷 👍

PR #299, opened 2020-06-17
[WIP] Add copy of sdf element in Actor copy constructor

luca-della-vedova:
The Actor copy constructor doesn't copy the sdf property and this causes a lot of issues down the line.

Specifically, when a user spawns an actor through the Create user command a [copy of an Actor object is created](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/systems/user_commands/UserCommands.cc#L524), then the CreateEntities function emits a [LoadPlugin event](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/SdfEntityCreator.cc#L315-L316) with the Element() pointer, which wasn't initialized, is a null pointer and causes a segmentation fault.

This bug fixes this specific segfault, reproducible by running an ign gazebo world and calling the ros_ign service create:
`ros2 run ros_ign_gazebo create -file ~/.ignition/fuel/fuel.ignitionrobotics.org/Mingfei/models/actor/1/model.sdf`
This results in a segmentation fault before the PR while the actor is correctly spawned with the PR.

I had a look at the CopyFrom function and noticed that it copies everything except [links, joints and the sdf element](https://github.com/osrf/sdformat/blob/master/src/Actor.cc#L107-L114). I'm wondering if this is intended (I'm guessing it is otherwise we would be using the default copy operator). I'm a bit worried if we don't copy these properties similar bugs might happen in future when users assume that all the properties are copied.
Would it make sense to get rid of the manual copy and fall back to the default copy of the private data? Similar to what is done in [Model](https://github.com/osrf/sdformat/blob/master/src/Model.cc#L98).

I'll fix the tests once we agree on what the copy behavior should be.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-syntax: glob
 
 build
 build_*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ set (CMAKE_INCLUDE_DIRECTORIES_PROJECT_BEFORE ON)
 # to choose the flag -std=gnu++14 instead of -std=c++14 when the C++14
 # features are requested. Explicitly turning this flag off will force cmake to
 # choose -std=c++14.
-# See https://bitbucket.org/ignitionrobotics/ign-cmake/issues/13 for more info.
+# See https://github.com/ignitionrobotics/ign-cmake/issues/13 for more info.
 set(CMAKE_CXX_EXTENSIONS off)
 
 # Include GNUInstallDirs to get canonical paths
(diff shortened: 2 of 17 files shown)


chapulina:
How about checking the SDF in these tests?

https://github.com/osrf/sdformat/blob/55029bbf533b26ddb2ea35acfdb4e1c7434a401e/src/Actor_TEST.cc#L64

> Would it make sense to get rid of the manual copy and fall back to the default copy of the private data? Similar to what is done in Model.

+1

PR #300, opened 2020-06-18
sdf 1.8: support specifying frames as joint child/parent 

### LINKED ISSUE #204: joint: should support specifying frames in `<child>` and `<parent>`
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/204)) by Eric Cousineau (Bitbucket: [eacousineau](https://bitbucket.org/%7B71c9fc05-02c8-45e1-adfa-13c8ac3399dc%7D/), GitHub: [EricCousineau-TRI](https://github.com/EricCousineau-TRI)).**

# Summary

At present, the `<child>` and `<parent>` elements of `<joint>` are constrained to be links:
http://sdformat.org/spec?ver=1.6&elem=joint#joint_parent

It would be excellent if the specification permitted using frames. Per #200, if a link creates an implicit frame, then this change will be backwards-compatible at a specification level.

# Motivation

When specifying a world in SDF, it would be nice to specify joints with more semantics than just a link and an offset; e.g. if I want to weld a mug to the top of a table, I can specify two frames (completely invariant of the model frames) on the mug and table, and then just weld those two frames together. For a robot with revolute joints, you can embed better semantics when assembling the robot.

I am willing to provide a draft change to the spec, and possibly to the API as well to accommodate this change while considering backwards compatibility.

# Describe alternatives you've considered

TTBOMK, the alternative is to use the present state, which is to manually compute the kinematics; while feasible, this is brittle due to numerics. Xacros / ruby templates shouldn't be necessary for conveniently joining things.

# Additional context

If this proposal is accepted, I would like to ensure that this feature is implemented in Drake here:
https://drake.mit.edu/doxygen_cxx/namespacedrake_1_1multibody_1_1parsing.html#a3e9bff0197c6c7f86f9ff6f8854def53
https://drake.mit.edu/pydrake/pydrake.multibody.multibody_tree.parsing.html#pydrake.multibody.multibody_tree.parsing.AddModelFromSdfFile
Downstream issue: https://github.com/RobotLocomotion/drake/issues/9910

\cc @nkoenig

scpeters:
Resolves #204.

The meat of this PR is in 53bdf38, which modifies the `sdf/1.8/joint.sdf` documentation and 
cf095f3, which changes how the graphs are constructed in FrameSemantics.cc. There's a bit of noise from deprecation of API's like `Joint::ChildLinkName` and tests, so start with those two commits.

---
This change is [Reviewable](https://reviewable.io/reviews/osrf/sdformat/300)

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -189,6 +189,10 @@ but with improved human-readability..
 
 ## SDF protocol 1.7 to 1.8
 
+### Modifications
+
+1. **joint.sdf** `child` and `parent` elements accept frame names instead of only link names
+
 ### Deprecations
 
 1. **joint.sdf** `initial_position` element in `<joint><axis>` and `<joint><axis2>` is deprecated
--- a/Migration.md
+++ b/Migration.md
@@ -26,6 +26,10 @@ but with improved human-readability..
    + ***Deprecation:*** void SetParentLinkName(const std::string &)
    + ***Replacement:*** void SetParentName(const std::string &)
 
+1. **sdf/parser.hh**
+   + ***Deprecation:*** bool checkJointParentChildLinkNames(const sdf::Root \*)
+   + ***Replacement:*** bool checkJointParentChildNames(const sdf::Root \*)
+
 ## SDFormat 8.x to 9.0
 
 ### Additions
--- a/Migration.md
+++ b/Migration.md
@@ -12,6 +12,20 @@ forward programmatically.
 This document aims to contain similar information to those files
 but with improved human-readability..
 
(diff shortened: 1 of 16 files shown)


EricCousineau-TRI:
Reviewable: https://reviewable.io/reviews/osrf/sdformat/300#-

scpeters:
I'm going to close this in favor of #304 because I think the deprecations are too much noise for a PR adding functionality. I've incorporated some of the feedback about pose offsets in the example files and adding a link to the migration guide in #304.

PR #301, opened 2020-06-19
Fix Actor.cc copy operators and restructure tests

luca-della-vedova:
This PR is a followup of #299, it changes all the copy operators to use the default copy operator of the private data object, fixing a few properties that were not copied in the manual copy (such as the actor properties trajectories, joints, links, sdf element).

I also took the chance to implement unit tests for every class (before only actors were being tested) and restructure the tests a bit to reduce code duplication.
The tests are now much more comprehensive. What can be set through setter functions is tested in the unit tests. The parts that are set on SDF parsing are tested in the integration test (the example sdf has also been updated to add joints and links, previously untested).

Open to feedback about what to do with the CopyFrom functions, right now they are broken since they don't copy all the properties, I wanted to deprecate them but is that OK to do in a minor release?

### DIFF
--- a/src/Actor_TEST.cc
+++ b/src/Actor_TEST.cc
@@ -61,7 +61,7 @@ bool TrajectoriesEqual(const sdf::Trajectory &_traj1,
     const sdf::Trajectory &_traj2)
 {
   constexpr double EPS = 1e-6;
-  bool waypoints_equal = true;
+  bool waypointsEqual = true;
   if (_traj1.WaypointCount() != _traj2.WaypointCount())
   {
     return false;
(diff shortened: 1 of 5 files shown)


src/Actor_TEST.cc
```diff
@@ -19,6 +19,127 @@
...
+  {
+    return false;
+  }
+  for (uint64_t traj_idx = 0; traj_idx < _actor1.TrajectoryCount(); ++traj_idx)
+  {
+    trajectories_equal &= TrajectoriesEqual(
+        *_actor1.TrajectoryByIndex(traj_idx),
+        *_actor2.TrajectoryByIndex(traj_idx));
+  }
+  return animations_equal && trajectories_equal &&
```
iche033, line 128: nitpick: camelCase for `animations_equal` and `trajectories_equal`

luca-della-vedova: [Done](https://github.com/osrf/sdformat/pull/301/commits/f00cf338e2131fdedfe35184c14f47e36bec93d9), a quick search for `_equal` showed that `waypoints_equal` had the same issue so fixed that as well.

scpeters:
I don't know why the GitHub action didn't run for this PR

luca-della-vedova:
> I don't know why the GitHub action didn't run for this PR

I'm not sure how to fix that, can someone retrigger the action?

scpeters:

j-rivero:
@osrf-jenkins run tests

scpeters:
I think #310 should enable the actions CI for this branch

Issue #260, opened 2020-04-29
Parsing of bad values (syntactically and semantically) should be more strict


azeey:
Example of bad syntax: #228 
Example of bad semantics: https://github.com/robotlocomotion/drake/pull/12863#pullrequestreview-375615059

This is actually a spec change; "bad" values should throw an error by design, not be ignored by design. Currently, a negative effort limit implies ignoring a joint limit...

For handling bad semantics, look into using `min` and `max` values in the spec descriptions

For non-scalar values like `//joint/axis/xyz`, libsdformat should enforce that the axis is a unit vector by either re-normalizing or requiring it to be normalized.

Also, spec should say that it will disregard non-unit vectors for axes.

We should do PR for libsdformat10 as an error.

If we have extra time, backport to libsdformat9, but as a warning. (With updates to Gazebo Classic + Ignition Gazebo.)

azeey:
Addressed by #244 and #259

EricCousineau-TRI:
Marking as done, but spun off separate issue about `//joint/axis/xyz`
https://github.com/osrf/sdformat/issues/307

PR #303, opened 2020-06-23
Enforce minimum/maximum values specified in SDFormat description files (Retry PR)

azeey:
Currently, some SDFormat description files contain min and max values, but the parser does not enforce those values in any way. This PR adds a validation step in the parser that checks if scalar values are within the allowed range. Note that this PR addresses only element values, not attributes.

Toward #260.

This is a retry of #259.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,9 @@
 
 ### libsdformat 10.0.0 (202X-XX-XX)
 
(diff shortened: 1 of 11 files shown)


scpeters:
I would add something to the Migration guide about this, and the changelog as well

PR #304, opened 2020-06-24
sdf 1.8: support specifying frames as joint child/parent

### LINKED ISSUE #204: joint: should support specifying frames in `<child>` and `<parent>`

scpeters:
Resolves #204, replacement for #300.

This changes the definition of the `//joint/child` and `//joint/parent` elements in SDFormat 1.8 so that they may specify a frame name instead of only a link name, per the [composition proposal](http://sdformat.org/tutorials?tut=composition_proposal&cat=pose_semantics_docs&#1-2-interface-elements). I've added some example models using this behavior in 85ab1c5. The parsing stages used to construct the `FrameAttachedToGraph` and `PoseRelativeToGraph` were changed a bit in de80f92 in order to support this feature, as noted in https://github.com/osrf/sdf_tutorials/commit/77b0db1f4e8e91b6cbdefe5f318f056001855712 (currently blocked by https://github.com/osrf/sdf_tutorials/pull/28).

Methods for resolving the name of a joint's parent and child links are added in 655833c:

* `Joint::ResolveChildLink`
* `Joint::ResolveParentLink`

I've redacted the deprecations from #301 from this PR to reduce the noise. We can rename API's in a follow-up PR.

---
This change is [Reviewable](https://reviewable.io/reviews/osrf/sdformat/304)

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -221,6 +221,10 @@ but with improved human-readability..
 
 ## SDF protocol 1.7 to 1.8
 
+### Modifications
+
+1. **joint.sdf** `child` and `parent` elements accept frame names instead of only link names
+
 ### Deprecations
 
 1. **joint.sdf** `initial_position` element in `<joint><axis>` and `<joint><axis2>` is deprecated
--- a/Migration.md
+++ b/Migration.md
@@ -35,7 +35,7 @@ but with improved human-readability..
 1. + Removed the `parser_urdf.hh` header file and its `URDF2SDF` class
    + [Pull request 276](https://github.com/osrf/sdformat/pull/276)
 
-1. + Removed the deprecated `Pose()`, `SetPose(), and `*PoseFrame()` API's in all DOM classes:
+1. + Removed the deprecated `Pose()`, `SetPose()`, and `*PoseFrame()` API's in all DOM classes:
    + const ignition::math::Pose3d &Pose()
    + void SetPose(const ignition::math::Pose3d &)
    + const std::string &PoseFrame()
--- a/Migration.md
+++ b/Migration.md
@@ -232,6 +232,7 @@ but with improved human-readability..
 ### Modifications
 
 1. **joint.sdf** `child` and `parent` elements accept frame names instead of only link names
+    * [Issue 204](https://github.com/osrf/sdformat/issues/204)
 
 ### Deprecations
 
--- a/Migration.md
+++ b/Migration.md
@@ -12,6 +12,14 @@ forward programmatically.
 This document aims to contain similar information to those files
 but with improved human-readability..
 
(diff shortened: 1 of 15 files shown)


scpeters:
ready for review from @EricCousineau-TRI

EricCousineau-TRI:
*[a discussion](https://reviewable.io/reviews/osrf/sdformat/304#-MBooD0vAQaJ395C7pwX:-MBorFUhDCT2YCspc9Rp:b-n139l8) (no related file):*
<details><summary><i>Previously, EricCousineau-TRI (Eric Cousineau) wrote…</i></summary><blockquote>

To do:
Make a mention of "stacking joints" via joint implicit frames, and how that will do parallel, not serial, joints.
</blockquote></details>

Will write up XML for `L1 -> J1 -> J2 -> L2`

---

PR #312, opened 2020-07-02
Normalize joint axis xyz vector when parsing from SDFormat

### LINKED ISSUE #307: //joint/axis/xyz: Should normalize, or fail if norm(xyz) < tol
At present, `JointAxis` asks for vectors to be normalized, but does not enforce / fix it:
https://github.com/osrf/sdformat/blob/55029bbf533b26ddb2ea35acfdb4e1c7434a401e/include/sdf/JointAxis.hh#L85-L87

Ideal:
* If `norm(xyz) < eps`, fail. Normalizing this is junk.
* Otherwise, `xyz /= norm(xyz)`.

azeey:
Closes #307. Note that normalization is done only when parsing from SDFormat, i.e, values set via `JointAxis::SetXyz` are not normalized.


scpeters:
ready for review from @EricCousineau-TRI

azeey:

azeey:
I've looked through the libsdformat code and we do already use SDF_ASSERT in a few places ([Element.cc:970](https://github.com/osrf/sdformat/blob/bd119e5b368b65b7052ae8a88cea99639f4a3c2d/src/Element.cc#L970), [Param.cc:71](https://github.com/osrf/sdformat/blob/bd119e5b368b65b7052ae8a88cea99639f4a3c2d/src/Param.cc#L71)), so I've added it in `JointAxis::SetXyz`. But I would note that all of the existing uses of `SDF_ASSERT` are in code that provides the `sdf::Element` API, so this the first use of exceptions in the DOM API.

EricCousineau-TRI:
Per f2f, if the return type is changed from `void` to `Error`, then the function should gain a `[[nodiscard]]` attribute so that warnings can be thrown.

include/sdf/JointAxis.hh
```diff
@@ -93,6 +94,7 @@ namespace sdf
 
     /// \brief Set the x,y,z components of the axis unit vector.
     /// \param[in] _xyz The x,y,z components of the axis unit vector.
+    /// \throws sdf::AssertionInternalError if the norm of the xyz vector is 0.
```
scpeters, line 97: I would prefer returning `Errors` from this function rather than throwing an exception

azeey: Changed to returning `sdf::Errors` with a `[[nodiscard]]` attribute in https://github.com/osrf/sdformat/pull/312/commits/6ccb301fda9659681f1af1eba97b04dcff70e3e2

azeey:

EricCousineau-TRI:
Per f2f, this is most likely broken due to some oddities with github actions (e.g. fetching opaquely from `pull/XYZ/merge` rather than `pull/XYZ/head`).

dirk-thomas:
@azeey This pull request seem to have introduced the following compiler warnings: https://build.osrfoundation.org/job/ignition_launch-ci-master-bionic-amd64/10/

[Discovered during osrf/buildfarmer#84]

scpeters:
> @azeey This pull request seem to have introduced the following compiler warnings: https://build.osrfoundation.org/job/ignition_launch-ci-master-bionic-amd64/10/
> 
> [Discovered during [osrf/buildfarmer#84](https://github.com/osrf/buildfarmer/issues/84)]

yes, this has caused some test failures, now that it has been included in the prerelease 10.0.0~pre210.0.0~pre2

the test failures will be fixed by https://github.com/ignitionrobotics/ign-gazebo/pull/288

Issue #313, opened 2020-07-06
Precision loss when adding a nested model


hauke76:
Hi,

I stumbled over the issue that adding nested models results in a floating point precision loss. When adding a nested model, the `_includeSDF` is [converted to a string](https://github.com/osrf/sdformat/blob/master/src/parser.cc#L1325) through a call to `Element::ToString`. This ultimately calls `Param::GetAsString()` which uses the [ParamStreamer](https://github.com/osrf/sdformat/blob/master/include/sdf/Param.hh#L78) to write data to a `std::ostream`. By injecting `std::setprecision(std::numeric_limits<T>::max_digits10)` in the output stream this precision loss can be prevented.

Kind regards,
Hauke

hauke76:
I created a pull request for the master branch here: https://github.com/osrf/sdformat/pull/314
I have a [similar fix](https://github.com/osrf/sdformat/compare/sdf6...hauke76:fix_precision_loss_v6) available for v6.x and it would be great if you could consider that as a merge request as well.

EricCousineau-TRI:
Given merge of #314, closing this issue. Thanks!

PR #314, opened 2020-07-06
Fixes precision loss when adding nested models.

hauke76:
The precision loss is resolved by explicitly calling std::setprecision
for float and double parameters when these are streamed/requested as
strings.


scpeters:
addresses #313

scpeters:
I think we would need a test before we could merge this. Fortunately, there are some existing tests for the text output after including nested models (see [test/integration/nested_model.cc](https://github.com/osrf/sdformat/blob/sdf9/test/integration/nested_model.cc#L573-L616)) that could be modified to increase precision. For example, https://github.com/scpeters/sdformat/commit/dce1c4d21490d36c6ecf8bd068ac3487ece1729b increases the precision of a pose value in a nested model and expects the same precision in the output. Is that the type of thing this is supposed to fix? I tried that test with your branch, and it didn't seem to fix it. With which data types have you tested your branch? I think Vector3 and Quaternion values might have precision limited by the ign-math class:

* https://github.com/ignitionrobotics/ign-math/blob/master/include/ignition/math/Quaternion.hh#L1208-L1219
* https://github.com/ignitionrobotics/ign-math/blob/master/include/ignition/math/Vector3.hh#L737-L747

hauke76:
The proposed fix only addresses double and float types stored in the parameter variant. I ran into this when using camera intrinsics such as fx, fy, etc. I already though about the vector and matrix types. We need special branches to fix that too as the default streaming operators you pointed to only serialize with [6 decimals](https://github.com/ignitionrobotics/ign-math/blob/master/include/ignition/math/Quaternion.hh#L1216-L1217) precision.

I will extend this CL to also fix those types and add a test.

hauke76:
I have modified the existing test to show that the correct code path is taken and to show that the logic is correct. Since I have not found an SDF property with 32bit float precision, I only tested for double precision. I used the sphere's [radius property](https://github.com/osrf/sdformat/blob/master/sdf/1.8/sphere_shape.sdf#L3) which is a double.

All floating point ignition math types are affected as well. I could easily provide full specialization for them as and re-implement the serialization while taking care to exactly copy the math libraries' behavior but I am wondering if the better fix were to remove the `precision` function invocation in the ignition math library. It were then possible to simply write `os << std::setprecision(std::numeric_limits<T>::max_digits10) << v.val` for the ignition types. Right now, that won't work.

EricCousineau-TRI:
For the `std::setprecision()`, would it be possible to make sure you reverse any global effects? e.g. using some RAII pattern?

EricCousineau-TRI:
Relates https://github.com/osrf/sdf_tutorials/pull/37. For example:
https://github.com/osrf/sdf_tutorials/blob/921244bca7068050d5d2c74b9dec58bf0f46ddb3/specify_pose/proposal.md#motivation

jennuine:
> For the `std::setprecision()`, would it be possible to make sure you reverse any global effects?

There are no global effects because the `ostream` used in the `operator<<` functions is a locally created stream object

azeey (approved):

EricCousineau-TRI (approved):
Sweet, thank you for checking!

Issue #293, opened 2020-05-28
Should allow "drilling down" into nested model frames


EricCousineau-TRI:
Towards #278

Should happen after #283

Some options:
- Just "hoist" duplicates of the frames (and frame graphs) with `{model_name}::{frame_name}`
- Actually try to parse the name and actually navigate down

scpeters:
Some User Story ideas for this feature:

* Loading a model that contains `::` syntax for referencing frames inside a nested model in the following places:
    * `//joint/parent`
    * `//joint/child`
    * `//frame/@attached_to`
    * `//pose/@relative_to`
* Resolving a `SemanticPose` to a nested frame (i.e. `link->SemanticPose.Resolve(pose, "nestedModel::nestedLink");`)
* Resolving the attached-to body for a frame attached to a nested link (i.e. it returns `"nestedModel::nestedLink"`)
* Resolving the attached-to body for a frame attached to a nested model to the canonical link of that nested model
* Resolving a joint parent/child to a nested link (i.e. it returns `"nestedModel::nestedLink"`)
* Support passing scoped names to the following functions:
    * Model::ModelByName
    * Model::LinkByName
    * Model::JointByName
    * Model::FrameByName
* Getting a `SemanticPose` object corresponding to a scoped frame name (i.e. `Errors Model::SemanticPoseByName(std::unique_ptr<SemanticPose> &_semanticPose, const std::string _scopedFrameName)`). We could actually implement an unscoped version of this in `libsdformat9` that just checks the Links, Joints, and Frames in the current model for matches with the passed name and returns the SemanticPose of the named object.

scpeters:
> Support passing scoped names to the following functions:
> 
>     * Model::ModelByName
> 
>     * Model::LinkByName
> 
>     * Model::JointByName
> 
>     * Model::FrameByName

I've prototyped a variation on this in https://github.com/scpeters/sdformat/commit/9eec062d048abbe5f70b3d12c69d07b4cbd0f7eb by adding `Model::*ByNestedName` and `Model::*NestedNameExists` methods that accept a `::` delimited name while leaving the existing APIs unchanged.

EricCousineau-TRI:
For supergraph + subgraph assembly, scoping, and queries, options for recursion:
- Copy + Graft
- Scoped Alias + Propagate
Pros/cons are in maintaining invariants and ease of local checks (e.g. encapsulation at each level).

[...]

Issue #283, opened 2020-05-15
Nested models should work via DOM in `libsdformat9`


EricCousineau-TRI:
At present, models that are *directly* nested in a `*.sdf` file are unable to be parsed:
https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/610/page/1

They should be supported, e.g. via `Model::NestedModels()`, or something akin to that.

Towards: https://github.com/osrf/sdformat/issues/278

scpeters:
I've started work on this in the following branch:

* https://github.com/scpeters/sdformat/tree/nested_model_dom

scpeters:
this implies a change in the parsing stages, so I'd like to make update those in the documentation as well (see #280)

scpeters:
I've implemented the `Model::Model*` nested model accessors in a draft PR (https://github.com/scpeters/sdformat/pull/1). It loads nested model DOM objects but doesn't add support for frame semantics like the following:

~~~
<model name="outside_model">
  <link name="middle_link"/>
  <model name="middle_model">
    <link name="inner_link"/>
  </model>
  <frame name="middle_frame" attached_to="middle_model"/>
</model>    
~~~

scpeters:
I was planning to submit a PR about this after #304 lands, but after discussing it with @azeey, it seems like we could target this to `libsdformat9` / SDFormat 1.7. We will discuss with @EricCousineau-TRI

scpeters:
since `libsdformat9` has not been included in Ubuntu 20.04, I think we have a chance to add this behavior to libsdformat 9.3 in time for 20.10

EricCousineau-TRI:
Marking as closed. Farmed out canonical links to #343 and #342

PR #316, opened 2020-07-17
[sdf9] Support nested models in DOM and frame semantics

scpeters:
Nested model elements (`//model/model`) are currently supported in the SDFormat 1.7 spec, but are not supported by the DOM API or frame semantics operations in libsdformat 9.2. Per [Amendment 1 of the SDFormat 1.7 proposal](http://sdformat.org/tutorials?tut=pose_frame_semantics_proposal#amendment-1-directly-nested-models), this PR adds support for nested models in the DOM API and frame semantics (fixing #283) through three steps:

* adding `Model::Model*` methods for accessing nested models via the DOM API (047ec96)
* loading nested models in `Model::Load` (b57fea2, step 3 of [model parsing stages](http://sdformat.org/tutorials?tut=pose_frame_semantics_proposal#1-model))
* supporting nested models (`//model/model`) in frame semantics as opaque frames to match how models in the world scope (`//world/model`) are treated (85e0b4f, steps 6-9 of [model parsing stages](http://sdformat.org/tutorials?tut=pose_frame_semantics_proposal#1-model))

The first two steps are straightforward, while the 3rd step adds a bit of new behavior. I think it is reasonable to add this new behavior to `libsdformat9` because it is compatible with the SDFormat 1.7 spec (which supports nested models) and makes the treatment of models more consistent with regard to frame semantics. Currently in libsdformat 9.2.0, a `//world/model` supports frame semantics: they have frames that can be reference by name with `//pose/@relative_to` and `//world/frame/@attached_to` values can resolve to a `//world/model`. This PR extends that same behavior to nested `//model/model` elements; they would have their own frames in the frame and pose graphs, and a `//model/frame` is permitted to resolve to a model.

This does not add support for referencing elements within a model via the `::` syntax in the frame semantics or DOM APIs. That will be added in SDFormat 1.8 (see #293).

A companion pull request to `sdf_tutorials` will be submitted that documents the changes proposed in this pull request. This pull request bumps the minor version to 9.3. I believe we can make this behavior change because `libsdformat9` has not been included as an official package in Ubuntu 20.04, so we don't have to worry about lots of people being stuck on an old version. Version 9.2 is currently in packages.ros.org, but I think we could get that version updated. (cc @j-rivero)

---
This change is [Reviewable](https://reviewable.io/reviews/osrf/sdformat/316)


EricCousineau-TRI:
Reviewable link: https://reviewable.io/reviews/osrf/sdformat/316#-

EricCousineau-TRI:
Oh... dern haha, we missed that entirely. I think we should make it work, b/c it makes sense.
Will file a PR to `sdf_tutorials` to try to capture this.

EricCousineau-TRI:
Here's a (kinda messy) proposed change: https://github.com/osrf/sdf_tutorials/pull/36

EricCousineau-TRI:
Per VC (again), the nested canonical link proposal may be implemented as part of this PR, or a preceding PR.

PR #341, opened 2020-08-20
Support implicit nested canonical links

scpeters:
This is a follow-up to #283 to support some models without direct child links if they contain a nested model with nested links like the following:

~~~
<model name="top">
  <model name="nested">
    <link name="link"/>
  </model>
</model>
~~~

Nested canonical links can be used implicitly, but not explicitly since the `::` syntax for referencing across model boundaries will not be supported until SDFormat 1.8.

The `FrameAttachedToGraph` does need to store the relative name of a nested canonical link using `::` syntax, so a private helper function `Model::CanonicalLinkAndRelativeName` is added that computes this relative name. This helper function is called by `Model::CanonicalLink` to avoid code duplication.

---
This change is [Reviewable](https://reviewable.io/reviews/osrf/sdformat/341)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,7 @@
 
 1. Support nested models in DOM and frame semantics.
     * [Pull request 316](https://github.com/osrf/sdformat/pull/316)
+    + [Pull request 341](https://github.com/osrf/sdformat/pull/341)
 
 1. Find python3 in cmake, fix cmake warning.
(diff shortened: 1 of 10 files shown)


EricCousineau-TRI:
Should resolve #342

iche033:
I tested this PR with ignitionrobotics/ign-physics#66 and ignitionrobotics/ign-gazebo#258. A model with a nested link and no direct child links now successfully loads in ign-gazebo. Changes look good to me.

PR #347, opened 2020-09-01
Add collection labeler action

chapulina:
Tested here: https://github.com/chapulina/sdformat/pull/1

This action will automatically label pull requests that target branches used by Gazebo collections according to the target branch. For sdformat, this currently means:

* `sdf8`: blueprint
* `sdf9`: citadel
* `sdf10`: dome


scpeters (approved):

chapulina:
Humm it's looking like this will need to be backported. Sorry for the wrong info above.

PR #328, opened 2020-08-03
Find python3 in cmake, fix warning

scpeters:
There are some cmake warnings now reported by jenkins since we upgraded one of the jenkins plugins:

* [Build Status](https://build.osrfoundation.org/job/sdformat-ci-sdformat9-bionic-amd64/31/) https://build.osrfoundation.org/job/sdformat-ci-sdformat9-bionic-amd64/31/

~~~
CMake Warning at CMakeLists.txt:226 (message):
  -- 	Python psutil package not found.  Memory leak tests will be skipped
~~~

The issue is that we are installing the python3 version of `psutil`, so we need to find the `python3` interpreter in cmake.

PR #349, opened 2020-09-02
Store material file path information

nkoenig:
Storing the file path of material is required to fix https://github.com/ignitionrobotics/ign-gazebo/issues/286.

This follows the same pattern as `Mesh::FilePath`.


nkoenig:
See also: https://github.com/ignitionrobotics/ign-gazebo/pull/328

scpeters (approved):

scpeters:
there are just a few cmake warnings that would be fixed by back-porting #328, which I will do in a separate PR

Issue #231, opened 2020-04-10
noise.sdf: add min="0.0" to the precision and dynamic_bias_correlation_time parameters


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/231)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

----------------------------------------
This is a follow-up task from [pr #329](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/329), which proposed to "Add random walk params to noise for sdf 1.6". Most of the changes in this PR have already been merged in [PR #525](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/525) except for the differences noted in the following comment:

* https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/329/add-random-walk-params-to-noise-for-sdf-16/diff#comment-134273497

Two of the differences are still worth applying, but I'm putting them in an issue to be resolved after the migration to GitHub.

* noise.sdf: add min="0.0" to the precision parameter
* noise.sdf: add min="0.0" to the dynamic_bias_correlation_time parameter

PR #357, opened 2020-09-04
Return positive infinity instead of -1 for boundless positive values

scpeters:
Targeted at `sdf10`, this is a resubmission of #249, which itself was a revival of [bitbucket PR 458](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/pull-requests/458) per #231. I've added tests in 148d782. cc @mxgrey 

## Original description:

SDFormat has a convention of using -1 to represent positive infinity. If users of the DOM API are not careful and cognizant of this convention, it can result in confusing bugs that are hard to track down. My dartsim plugin was exhibiting subtle non-physical behavior after loading structures from SDF, because it had effort and velocity limits of +1 (minimum) and -1 (maximum).

This PR is a suggested tweak to the API behavior so that the API returns positive infinity instead of -1 to represent boundlessly positive values. Hopefully this might save other SDF users from long, confusing debugging sessions.

However, this does deviate from the exact convention of the SDF document specification, so I’d understand if this PR is declined.

### DIFF
--- a/include/sdf/JointAxis.hh
+++ b/include/sdf/JointAxis.hh
@@ -197,13 +197,13 @@ namespace sdf
     public: double Effort() const;
 
     /// \brief Set the value for enforcing the maximum joint effort applied.
-    /// Limit is not enforced if value is negative.
+    /// Limit is not enforced if value is infinity.
     /// \param[in] _effort Effort limit.
     /// \sa double Effort() const
     public: void SetEffort(double _effort);
 
     /// \brief Get the value for enforcing the maximum joint velocity. The
-    /// default value is -1.
+    /// default value is infinity.
     /// \return The value for enforcing the maximum joint velocity.
     /// \sa void SetVelocity(const double _velocity) const
     public: double MaxVelocity() const;
--- a/include/sdf/JointAxis.hh
+++ b/include/sdf/JointAxis.hh
@@ -190,27 +190,31 @@ namespace sdf
     /// \sa double Upper() const
     public: void SetUpper(const double _upper) const;
 
-    /// \brief Get the value for enforcing the maximum joint effort applied.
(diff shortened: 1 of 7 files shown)


scpeters:
@mxgrey if you want to amend with a `--signoff` your old commit migrated from bitbucket, it will make the DCO bot happy

scpeters:

scpeters:

include/sdf/JointAxis.hh
```diff
@@ -197,13 +197,13 @@ namespace sdf
...
     /// \brief Set the value for enforcing the maximum joint effort applied.
-    /// Limit is not enforced if value is negative.
+    /// Limit is not enforced if value is infinity.
     /// \param[in] _effort Effort limit.
     /// \sa double Effort() const
     public: void SetEffort(double _effort);
 
     /// \brief Get the value for enforcing the maximum joint velocity. The
-    /// default value is -1.
+    /// default value is infinity.
```
chapulina, line 206: Maybe it's worth mentioning in the docs that these values are absolute, so they apply to both positive and negative directions?

chapulina (approved):
LGTM, I just have a minor comment about documentation.

include/sdf/JointAxis.hh
scpeters: sure, I'll add it

scpeters: 0227719

scpeters:
I think this doesn't require any downstream changes in ion-physics, since it is already using a similar conversion:

* https://github.com/ignitionrobotics/ign-physics/blob/ignition-physics2_2.2.0/dartsim/src/SDFFeatures.cc#L126-L129

PR #358, opened 2020-09-04
CMake fixes

joxoby:
When trying to build from Conan, not including `CMakePackageConfigHelpers` ends in an error saying that it can't find `configure_package_config_file`. I'm not exactly sure why it's actually working without it.

Also, use modern CMake target for `ignition-math`.


mjcarroll (approved):

scpeters:

scpeters (approved):

scpeters:
yeah, I don't know why it was working without this fix either. Thanks for the contribution!

PR #359, opened 2020-09-04
Fix whitespace preservation behavior with TinyXML2

### LINKED ISSUE #322: Whitespace preservation inconsistency
While backporting some tests from Dome to Blueprint (https://github.com/ignitionrobotics/ign-sensors/pull/33/), I noticed that there's an inconsistency with how `sdf8` and `sdf10` treat multiple whitespaces inside elements (not inside arguments). The following test fails on `sdf8` but passes on `sdf10`.

I believe this difference comes from the default behaviour of `tinyxml` versus `tinyxml2`, according to `tinyxml` documentation on [whitespace preservation](https://leethomason.github.io/tinyxml2/):  "_By default, TinyXML-2 preserves white space in a (hopefully) sane way that is almost compliant with the spec. (TinyXML-1 used a completely different model, much more similar to 'collapse'_".

It would be nice to be consistent across SDF versions. I can see arguments for each behaviour though. This is not very important for our use cases right now, but I thought we should keep track of it.

### Test

```.diff
diff --git a/src/SDF_TEST.cc b/src/SDF_TEST.cc
index 27f194ad..cca71d2e 100644
--- a/src/SDF_TEST.cc
+++ b/src/SDF_TEST.cc
@@ -696,6 +696,37 @@ TEST(SDF, WriteURIPath)
 }
 #endif  // _WIN32
 
+TEST(SDF, readString)
+{
+  std::string testWorld = R"sdf(
+    <?xml version="1.0" ?>
+    <sdf version="1.5">
+      <world name="hello  world(123)">
+        <frame>
+          <name>hello  world(456)</name>
+        </frame>
+      </world>
+    </sdf>)sdf";
+
+  sdf::SDF sdfParsed;
+  sdf::initFile("root.sdf", sdfParsed.Root());
+  auto rootOriginal = sdfParsed.Root();
+  EXPECT_TRUE(sdf::readString(testWorld, rootOriginal));
+
+  EXPECT_TRUE(sdfParsed.Root()->HasElement("world"));
+  auto worldElem = sdfParsed.Root()->GetElement("world");
+  ASSERT_NE(nullptr, worldElem);
+
+  EXPECT_EQ("hello  world(123)", worldElem->Get<std::string>("name"));
+
+  EXPECT_TRUE(worldElem->HasElement("world"));
+  auto frameElem = worldElem->GetElement("frame");
+  ASSERT_NE(nullptr, frameElem);
+
+  EXPECT_EQ("hello  world(456)", frameElem->Get<std::string>("name"));
+}
+
```

### Failure on `sdf8`

```
/home/chapulina/blueprint_ws/src/sdformat/src/SDF_TEST.cc:723: Failure
Value of: worldElem->HasElement("world")
  Actual: false
Expected: true
/home/chapulina/blueprint_ws/src/sdformat/src/SDF_TEST.cc:727: Failure
  "hello  world(456)"
  frameElem->Get<std::string>("name")
    Which is: "hello world(456)"
```

mjcarroll:
This addresses a change in how whitespace is preserved/collapsed between
tinyxml and tinyxml2. The change primarily impacts SDF files that have
newlines, which is frequently done for aesthetics with include uris.

In this case, we use an alternate constructor for TinyXML2 that will
collapse whitespace by default.

Note that there is a performance impact of this behavior, which causes
the parsing to run essentially twice.

More information on the behavior may be found here: https://leethomason.github.io/tinyxml2/

Closes #322

Signed-off-by: Michael Carroll <michael@openrobotics.org>


src/parser.cc
```diff
@@ -78,11 +78,27 @@ bool readStringInternal(
     const bool _convert,
     Errors &_errors);
 
+//////////////////////////////////////////////////
+/// \brief Internal helper for creating XMLDocuments
```
chapulina, line 82: cppcheck warnings:

```
./src/parser.cc:82:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
./src/parser.cc:83:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
./src/parser.cc:89:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
./src/parser.cc:540:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
```

scpeters:
> I'm going to test a narrower approach that calls `sdf::trim` on the contents of `<uri>` tags

narrower approach in #365

scpeters (approved):
I think this is the right approach. I'll close #365

scpeters:

azeey (commented):
Do we need to make the same changes in the URDF parser (`src/parser_urdf.cc`)?

mjcarroll:
> Do we need to make the same changes in the URDF parser (src/parser_urdf.cc)?

I don't believe so, because my understanding was that URDF only uses attributes in tags, versus the actual content of tags:

```<pose rpy="0 0 0" xyz="0 0 0" />```
vs
```<pose>0 0 0 0 0 0</pose>```

Is that correct?

**Comment by azeey** [comment]
> > Do we need to make the same changes in the URDF parser (src/parser_urdf.cc)?
> 
> I don't believe so, because my understanding was that URDF only uses attributes in tags, versus the actual content of tags:
> 
> `<pose rpy="0 0 0" xyz="0 0 0" />`
> vs
> `<pose>0 0 0 0 0 0</pose>`
> 
> Is that correct?

I think you're right about that, but if the URDF contains Gazebo extensions (eg. https://answers.gazebosim.org//question/7082/using-sdf-tags-in-urdf-gazebo-extension-directly/), it could have values in tags.

**Comment by mjcarroll** [comment]
> URDF contains Gazebo extensions

Failed to account for that.  I will update and add a test for it.

**Comment by mjcarroll** [comment]
I added a test case in 2721714, it actually didn't require any adjustment for me in the actual URDF parser.

I think this is because we trim the values that we parse from URDF in the process of changing to SDF. 

In the case of this test, the `tinyxml2::XMLDocument doc` (parsed URDF) is:

```
<robot name="test">
    <link name="link">
        <inertial>
            <mass value="0.1"/>
            <origin rpy="1.570796326794895 0 0" xyz="0.123456789123456 0 0.0"/>
            <inertia ixx="0.01" ixy="0" ixz="0" iyy="0.01" iyz="0" izz="0.01"/>
        </inertial>
        <visual>
            <geometry>
                <sphere radius="1.0"/>
            </geometry>
        </visual>
        <collision>
            <geometry>
                <sphere radius="1.0"/>
            </geometry>
        </collision>
    </link>
    <gazebo reference="link">
        <material>
      Gazebo/Orange
    </material>
        <mu1>
      100
    </mu1>
        <mu2>

      1000

    </mu2>
    </gazebo>
</robot>
```

And then `sdfXml` (the converted SDF) is

```
<sdf version="1.7">
    <model name="test">
        <link name="link">
            <inertial>
                <pose>0.123456789123456 0 0 1.570796326794895 0 0</pose>
                <mass>0.1</mass>
                <inertia>
                    <ixx>0.01</ixx>
                    <ixy>0</ixy>
                    <ixz>0</ixz>
                    <iyy>0.01</iyy>
                    <iyz>0</iyz>
                    <izz>0.01</izz>
                </inertia>
            </inertial>
            <collision name="link_collision">
                <pose>0 0 0 0 0 0</pose>
                <geometry>
                    <sphere>
                        <radius>1</radius>
                    </sphere>
                </geometry>
                <surface>
                    <contact>
                        <ode/>
                    </contact>
                    <friction>
                        <ode>
                            <mu>100</mu>
                            <mu2>1000</mu2>
                        </ode>
                    </friction>
                </surface>
            </collision>
            <visual name="link_visual">
                <pose>0 0 0 0 0 0</pose>
                <geometry>
                    <sphere>
                        <radius>1</radius>
                    </sphere>
                </geometry>
                <material>
                    <script>
                        <name>Gazebo/Orange</name>
                        <uri>file://media/materials/scripts/gazebo.material</uri>
                    </script>
                </material>
            </visual>
            <gravity>true</gravity>
            <velocity_decay/>
        </link>
    </model>
</sdf>
```

azeey (approved):
Looks good! And thanks for adding the URDF Test.

PR #360, opened 2020-09-04
Add tinyxml2 as the config name for find_package

joxoby:
This will look for a file named `tinyxml2Config.cmake` instead of `TinyXML2Config.cmake`. This new name matches the Conan package name. If there's a use case in which we need a `TinyXML2Config.cmake` file, we can add it to the name list.


scpeters:

scpeters (approved):

PR #365, opened 2020-09-04
Trim whitespace from uri element text

### LINKED ISSUE #322: Whitespace preservation inconsistency

scpeters:
This copies the whitespace test from #359
but takes a narrower approach by only trimming
whitespace from `<uri>` element text.

Closes #322.

Signed-off-by: Steve Peters <scpeters@openrobotics.org>

### DIFF
--- a/src/Material.cc
+++ b/src/Material.cc
@@ -135,6 +135,7 @@ Errors Material::Load(sdf::ElementPtr _sdf)
     std::pair<std::string, bool> uriPair = elem->Get<std::string>("uri", "");
     if (uriPair.first == "__default__")
       uriPair.first = "";
+    uriPair.first = sdf::trim(uriPair.first);
 
     if (!uriPair.second || uriPair.first.empty())
     {
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -915,7 +915,8 @@ bool readXml(tinyxml2::XMLElement *_xml, ElementPtr _sdf, Errors &_errors)
 
         if (elemXml->FirstChildElement("uri"))
         {
-          std::string uri = elemXml->FirstChildElement("uri")->GetText();
+          std::string uri =
+              sdf::trim(elemXml->FirstChildElement("uri")->GetText());
           modelPath = sdf::findFile(uri, true, true);
 
           // Test the model path
--- a/test/integration/CMakeLists.txt
+++ b/test/integration/CMakeLists.txt
@@ -40,6 +40,7 @@ set(tests
   urdf_gazebo_extensions.cc
   urdf_joint_parameters.cc
   visual_dom.cc
+  whitespace.cc
   world_dom.cc
 )
 
--- a/test/integration/whitespace.cc
+++ b/test/integration/whitespace.cc
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2019 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include <iostream>
+#include <string>
+#include <gtest/gtest.h>
+
+#include "sdf/Actor.hh"
+#include "sdf/Collision.hh"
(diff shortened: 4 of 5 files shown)


chapulina (commented):
> only trimming whitespace from <uri> element text.

While working on an `ign-sensors` PR I needed the whitespace fix within a `<topic>` field: https://github.com/ignitionrobotics/ign-sensors/pull/33#discussion_r462408029

I think it's not safe to assume that only URIs need the fix. I wouldn't be surprised to see some of these around, for example:

```
<name>
    my_robot
</name>
```

scpeters:
> > only trimming whitespace from  element text.
> 
> While working on an `ign-sensors` PR I needed the whitespace fix within a `<topic>` field: [ignitionrobotics/ign-sensors#33 (comment)](https://github.com/ignitionrobotics/ign-sensors/pull/33#discussion_r462408029)
> 
> I think it's not safe to assume that only URIs need the fix. I wouldn't be surprised to see some of these around, for example:
> 
> ```
> <name>
>     my_robot
> </name>
> ```

I hear you, but it still feels a bit too heavy-handed to modify every xml file like this

scpeters:
clicked the wrong button

chapulina:
> I hear you, but it still feels a bit too heavy-handed to modify every xml file like this

I hear you too, the other PR should affect performance. My main concern at the moment is to keep the behavior backward-compatible.

Maybe there's a less invasive way of doing it. Extending your approach to all string values, maybe? I haven't checked how this affects elements like pose... Does this work, for example?

```
<pose>
1
2
3
0.1
0.2
0.3
</pose>
```

scpeters:
> > I hear you, but it still feels a bit too heavy-handed to modify every xml file like this
> 
> I hear you too, the other PR should affect performance. My main concern at the moment is to keep the behavior backward-compatible.
> 
> Maybe there's a less invasive way of doing it. Extending your approach to all string values, maybe? I haven't checked how this affects elements like pose... Does this work, for example?
> 
> ```
> <pose>
> 1
> 2
> 3
> 0.1
> 0.2
> 0.3
> </pose>
> ```

I wrote that but then thought better of it but clicked "close with comment" by accident. I've since deleted it. I'm still thinking it through.

scpeters:
closing in favor of #359

PR #369, opened 2020-09-08
Define PATH_MAX for Debian Hurd system

j-rivero:
The Hurd operativy system does not fully follow the POSIX standard and does not define the PATH_MAX variable, see bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892231. I just following how[ other projects](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=824005) patched this by defining if it does not exists.

### DIFF
--- a/src/Filesystem.cc
+++ b/src/Filesystem.cc
@@ -55,6 +55,11 @@
 
 #include "sdf/Filesystem.hh"
 
+/* PATH_MAX is undefined on GNU Hurd */
+#ifndef PATH_MAX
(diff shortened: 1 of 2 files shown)


scpeters (approved):

scpeters:
do you need this in a 9.3.1?

j-rivero:
> do you need this in a 9.3.1?

It can wait, I've included this changes in Debian latest version. thanks

PR #383, opened 2020-10-09
Fix supported shader types (normal_map_X_space)

chapulina:
Going over models on the Gazebo model database, I don't see a single model that uses:

* [normal_map_objectspace](https://github.com/osrf/gazebo_models/search?q=normal_map_objectspace&type=code)
* [normal_map_tangentspace](https://github.com/osrf/gazebo_models/search?q=normal_map_tangentspace&type=code)

Instead, there are models using:

* [normal_map_object_space](https://github.com/osrf/gazebo_models/search?q=normal_map_object_space&type=code)
* [normal_map_tangent_space](https://github.com/osrf/gazebo_models/search?q=normal_map_tangent_space&type=code)

I suspect that the documentation has been wrong all along, and when the `sdf::Material` class was created, the error was propagated.

The fix here changes the documentation to the "correct" values, but keeps the "old" values supported to avoid breaking existing code.

### DIFF
--- a/sdf/1.2/visual.sdf
+++ b/sdf/1.2/visual.sdf
@@ -41,7 +41,7 @@
       <description></description>
 
       <attribute name="type" type="string" default="pixel" required="1">
-        <description>vertex, pixel, normal_map_objectspace, normal_map_tangentspace</description>
+        <description>vertex, pixel, normal_map_object_space, normal_map_tangent_space</description>
       </attribute>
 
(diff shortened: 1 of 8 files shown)


iche033 (approved):
looks good, [gazebo agrees](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/rendering/Visual.cc#L775)

j-rivero:
@osrf-jenkins run tests please (just checking that new setting in not broken)

PR #388, opened 2020-10-13
Add Heightmap class

### LINKED ISSUE #382: Geometry C++ object lacks API's for Heightmaps
Using SDFormat 1.9 and libsdformat9, SDFormat XML `<geometry>` has [a `<heightmap>` tag](http://sdformat.org/spec?ver=1.7&elem=geometry#geometry_heightmap), but [the SDFormat C++ API `Geometry` class does not seem to support it](http://osrf-distributions.s3.amazonaws.com/sdformat/api/9.0.0/classsdf_1_1v9_1_1Geometry.html).

chapulina:
Resolves #382.

Added a C++ API for heightmap geometries that matches the XML spec.

I don't plan on changing the PR, but I'd like to see it working with `ign-gazebo` before merging, so I'll keep it as a draft while we work on the other parts.

### DIFF
--- a/include/sdf/CMakeLists.txt
+++ b/include/sdf/CMakeLists.txt
@@ -18,6 +18,7 @@ set (headers
   Frame.hh
   Geometry.hh
   Gui.hh
+  Heightmap.hh
   Imu.hh
   Joint.hh
   JointAxis.hh
--- a/include/sdf/Geometry.hh
+++ b/include/sdf/Geometry.hh
@@ -31,6 +31,7 @@ namespace sdf
   class GeometryPrivate;
   class Box;
   class Cylinder;
+  class Heightmap;
   class Mesh;
   class Plane;
   class Sphere;
@@ -56,6 +57,9 @@ namespace sdf
 
     /// \brief A mesh geometry.
     MESH = 5,
+
+    /// \brief A heightmap geometry.
+    HEIGHTMAP = 6,
   };
 
   /// \brief Geometry provides access to a shape, such as a Box. Use the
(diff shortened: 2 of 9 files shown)


scpeters:
looks ok to me; I'll look into how to add it to ign-physics; it will probably be similar to how meshes are handled, using `ign-common`

chapulina:
> I'll look into how to add it to ign-physics; it will probably be similar to how meshes are handled, using ign-common

In case it helps, [this](https://github.com/ignitionrobotics/ign-rendering/compare/main...chapulina/5/heightmap) is how I'm using `common::HeightmapData` in `ign-rendering`.

chapulina:
I'm working on the `ign-gazebo` PR and I have the need for some API to add textures and blends to a heightmap object. We don't do that much in the SDF DOM API, the only place where I saw `Add` functions is in `sdf::Actor`. Given that people have been discussing the API design (https://github.com/osrf/sdformat/issues/292), are there new guidelines for this use case?

If it's still undecided, I guess I could leave this as it is and use the `Element` API to construct an object instead...

chapulina:
This PR is working with https://github.com/ignitionrobotics/ign-gazebo/pull/487. I still have some details to fix on that one, but I think the API on this one is settled, so I'm opening it for review, @scpeters.

On a related note, because `ign-rendering` doesn't directly depend on SDF, and I needed a class there to hold heightmap data, I ended up making an almost exact copy of `sdf::Heightmap` for `ignition::rendering::HeightmapDescriptor` on https://github.com/ignitionrobotics/ign-rendering/pull/180. It's worth mentioning that similar data is also present in [ignition::msgs::HeightmapGeom](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs6/proto/ignition/msgs/heightmapgeom.proto). I'll ticket an issue about all this duplication with some ideas.

scpeters (approved):

PR #392, opened 2020-10-16
Move list of debian dependencies to packages.apt

chapulina:
This will help keeping the list of dependencies in a single place, which is easier to maintain. Being close to the source code also makes it simple to update when a PR introduces a new dependency.

Our Jenkins CI can start referring to this list instead of [dependencies_archive.sh](https://github.com/ignition-tooling/release-tools/blob/c9a13eecfc578d1157721eb3b003a92ae75c7185/jenkins-scripts/lib/dependencies_archive.sh#L55-L113) (see https://github.com/ignition-tooling/release-tools/pull/317) and other tools, like the [API doc uploader](https://github.com/ignitionrobotics/docs/blob/1e4a23088105e47aa89837203daf8f5ee951ce9a/tools/scripts/build_ign.sh#L18-L19) can also make use of it.

The pattern we've been using is:

* `packages.apt` for dependencies common to all platforms
* `packages-<distro>.apt` for dependencies specific to Ubuntu <distro>

### DIFF
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -1,0 +1,8 @@
+libignition-cmake2-dev
+libignition-math6-dev
+libignition-tools-dev
+libtinyxml-dev
+liburdfdom-dev
+libxml2-utils
+python-psutil
+ruby-dev
--- a/.github/workflows/linux-ubuntu-bionic.yml
+++ b/.github/workflows/linux-ubuntu-bionic.yml
@@ -16,7 +16,7 @@ jobs:
         sudo sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable $(lsb_release -cs) main" > /etc/apt/sources.list.d/gazebo-stable.list';
         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys D2486D2DD83DB69272AFE98867170598AF249743;
         sudo apt-get update;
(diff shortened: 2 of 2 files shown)


scpeters:
apparently including the string `packages.apt` in the branch name broke our jenkins script

~~~
sort -u ./sdformat/.github/ci/packages.apt ./sdformat/.git/refs/remotes/origin/chapulina/9/packages.apt ./sdformat/.git/logs/refs/remotes/origin/chapulina/9/packages.apt
~~~

chapulina:
> apparently including the string packages.apt in the branch name broke our jenkins script

Yeah I noticed it here :upside_down_face: That's why I did `find .github` instead.

I need to think of a better workaround, because the actions CI relies on finding dependencies of several packages in a workspace, on different folders.

scpeters:
> > apparently including the string packages.apt in the branch name broke our jenkins script
> 
> Yeah I noticed it here 🙃 That's why I did `find .github` instead.
> 
> I need to think of a better workaround, because the actions CI relies on finding dependencies of several packages in a workspace, on different folders.

fix for release-tools: https://github.com/ignition-tooling/release-tools/pull/323

scpeters:

scpeters:

chapulina:
CI is clear :raised_hands:  Can we get this in?

scpeters (approved):

PR #393, opened 2020-10-19
Add force torque sensor

nlamprian:
Addresses https://github.com/ignitionrobotics/ign-sensors/issues/25

Signed-off-by: Nick Lamprianidis <nlamprian@gmail.com>


src/ForceTorque_TEST.cc
```diff
@@ -0,0 +1,79 @@
...
+  sdf::Errors errors = ft.Load(sdf);
+  EXPECT_FALSE(errors.empty());
+  EXPECT_TRUE(errors[0].Message().find("is not a <force_torque>")
+      != std::string::npos) << errors[0].Message();
+
+  EXPECT_NE(nullptr, ft.Element());
+  EXPECT_EQ(sdf.get(), ft.Element().get());
+
+  // The ForceTorque::Load function is tested more thouroughly in the
+  // link_dom.cc integration test.
```
scpeters, line 78: wow, that test is actually wrong because the `force_torque` sensor is only supposed to be attached to a `joint`, not a `link`. we can merge this as is, but at some point we will need to correct `test/sdf/sensors.sdf`

here's an example from osrf/gazebo:

* https://github.com/osrf/gazebo/blob/6fd426b3949c4ca73fa126cde68f5cc4a59522eb/test/worlds/force_torque_frame_test.world#L88-L163

scpeters:
This PR is straight-forward; it exposes the existing data fields from the specification. It does raise some questions for me about the fact that `force_torque` sensors should only be attached to joints, not links, but this is not clearly stated in the specification; it is not reflected in our example files, and we don't check for that anywhere

I'll open an issue to discuss this further.

scpeters:
> This PR is straight-forward; it exposes the existing data fields from the specification. It does raise some questions for me about the fact that `force_torque` sensors should only be attached to joints, not links, but this is not clearly stated in the specification; it is not reflected in our example files, and we don't check for that anywhere
> 
> I'll open an issue to discuss this further.

#421

scpeters (approved):

osrf-triage:

osrf-triage:

PR #398, opened 2020-10-29
Move construction of PoseRelativeToGraph and FrameAttachedToGraph to sdf::Root

azeey:
This builds on top of #381 and simplifies the creation of PoseRelativeTo and FrameAttachedTo graphs by moving them to one place instead of being in `sdf::World` and `sdf::Model`. It also simplifies the logic of determining if a model is a standalone model that needs its own graph or a nested model that adds to an existing graph. 

This PR also changes the pointer held by ScopedGraph from a `std::weak_ptr` to a `std::shared_ptr`. This means that a DOM object, such as `sdf::Model`, can have its poses resolved even if the containing `sdf::Root` or `sdf::World` has gone out of scope.

A couple more notes:

* When copying any DOM object with the exception of `sdf::Root`, there is no need to call `SetPoseRelativeToGraph` or `SetFrameAttachedToGraph`on child elements because the copy constructors of those child elements copy the `ScopedGraph` from the corresponding child element from the copied-from object. Since a new graph is never created, the child elements should be pointing to the graph of the copied-from object when their copy constructor is finished.

* When copying from sdf::Root, we have to decide whether we want a deep copy of the graph or not. If we do a shallow copy, the constructor that simply copies the `dataPtr` is sufficient. Otherwise, we have to allocate new memory, make a copy of the graph and call `SetPoseRelativeToGraph` and `SetFrameAttachedToGraph` on all child elements of `sdf::Root`.


EricCousineau-TRI:
Per f2f, will retarget by PR'ing against Addisu's fork's branch.

azeey:
Closing in favor of https://github.com/azeey/sdformat/pull/1

PR #402, opened 2020-10-30
Remove custom element warning/error

jennuine:
Removed emitted error for custom elements and added test.

### DIFF
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -864,7 +864,7 @@ bool readXml(tinyxml2::XMLElement *_xml, ElementPtr _sdf, Errors &_errors)
   {
     // Avoid printing a warning message for missing attributes if a namespaced
     // attribute is found
-    if (std::strchr(attribute->Name(), ':') != NULL)
+    if (std::strchr(attribute->Name(), ':') != nullptr)
     {
       _sdf->AddAttribute(attribute->Name(), "string", "", 1, "");
       _sdf->GetAttribute(attribute->Name())->SetFromString(
@@ -1115,7 +1115,8 @@ bool readXml(tinyxml2::XMLElement *_xml, ElementPtr _sdf, Errors &_errors)
         }
       }
 
-      if (descCounter == _sdf->GetElementDescriptionCount())
+      if (descCounter == _sdf->GetElementDescriptionCount()
+            && std::strchr(elemXml->Value(), ':') == nullptr)
       {
         sdfdbg << "XML Element[" << elemXml->Value()
                << "], child of element[" << _xml->Value()
--- a/src/parser_TEST.cc
+++ b/src/parser_TEST.cc
@@ -20,6 +20,9 @@
 #include "sdf/Element.hh"
 #include "test_config.h"
 
+#include <fstream>
+#include <stdlib.h>
+
 /////////////////////////////////////////////////
 TEST(Parser, initStringTrim)
 {
@@ -57,6 +60,30 @@ sdf::SDFPtr InitSDF()
   sdf::SDFPtr sdf(new sdf::SDF());
   sdf::init(sdf);
   return sdf;
+}
+
+/////////////////////////////////////////////////
+/// Checks emitted warnings for custom/unknown elements in log file
+TEST(Parser, CustomUnknownElements)
+{
+  std::string pathBase = PROJECT_SOURCE_PATH;
+  pathBase += "/test/sdf";
+  const std::string path = pathBase +"/custom_and_unknown_elements.sdf";
+
+  sdf::SDFPtr sdf = InitSDF();
+  EXPECT_TRUE(sdf::readFile(path, sdf));
+
+  std::string line;
+  std::string pathHome = getenv("HOME");
+  std::ifstream file(pathHome + "/.sdformat/sdformat.log");
+
+  if (file.is_open())
+  {
+    std::getline(file, line);
+    file.close();
+  }
+
+  EXPECT_NE(line.find("XML Element[test_unknown]"), std::string::npos);
 }
 
 /////////////////////////////////////////////////
--- a/src/parser_TEST.cc
+++ b/src/parser_TEST.cc
@@ -75,8 +75,16 @@ TEST(Parser, CustomUnknownElements)
   sdf::SDFPtr sdf = InitSDF();
   EXPECT_TRUE(sdf::readFile(path, sdf));
 
+#ifndef _WIN32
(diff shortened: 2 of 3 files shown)


src/parser_TEST.cc
```diff
@@ -59,6 +62,30 @@ sdf::SDFPtr InitSDF()
...
+{
+  std::string pathBase = PROJECT_SOURCE_PATH;
+  pathBase += "/test/sdf";
+  const std::string path = pathBase +"/custom_and_unknown_elements.sdf";
+
+  sdf::SDFPtr sdf = InitSDF();
+  EXPECT_TRUE(sdf::readFile(path, sdf));
+
+  std::string line;
+  std::string pathHome = getenv("HOME");
```
azeey, line 77: Thanks for adding this test. Is there a way to make it write to a temporary file in the build directory instead of the user's home for testing purposes? Maybe by temporarily setting the `HOME` environment variable? Because otherwise, if the users `sdformat.log` already contains the expected "XML Element[test_unknown]" entry from previous runs, it becomes unclear whether the test is functioning properly.

src/parser_TEST.cc
```diff
@@ -59,6 +62,30 @@ sdf::SDFPtr InitSDF()
...
+  std::ifstream file(pathHome + "/.sdformat/sdformat.log");
+
+  if (file.is_open())
+  {
+    std::getline(file, line);
+    file.close();
+  }
+
+  EXPECT_NE(line.find("XML Element[test_unknown]"), std::string::npos);
```
azeey, line 86: Can you also check that `test:custom` is not in the file?

azeey (changes requested):

src/parser_TEST.cc
jennuine: Does it not get overwritten after a new execution? When I was testing it that seemed to be the case. And I put the test function towards the top so that other tests wouldn't output new items in the log file

src/parser.cc
```diff
@@ -1115,7 +1115,8 @@ bool readXml(tinyxml2::XMLElement *_xml, ElementPtr _sdf, Errors &_errors)
         }
       }
 
-      if (descCounter == _sdf->GetElementDescriptionCount())
+      if (descCounter == _sdf->GetElementDescriptionCount()
+            && std::strchr(elemXml->Value(), ':') == nullptr)
       {
         sdfdbg << "XML Element[" << elemXml->Value()
```
scpeters, line 1121: we should consider promoting this to a visible console message in libsdformat11

src/parser_TEST.cc
azeey: hmm, it does seem to be getting overwritten. That seems like a bug and we probably should not rely on it. I would still recommend writing to a test file so we don't write to anything outside of the build directory. The idea of hermetically sealing our libraries so that they don't write to files outside of the build directory is something we should strive to do as it makes our tests less flaky. It also makes them more portable to other build systems (eg. Bazel) . If manipulating environment variables without affecting other tests is a concern, we could put this in its own `.cc` file and state that no additional tests should be added to the file.

jennuine: Sounds good. So in the test, before `RUN_ALL_TESTS()` is called, https://github.com/osrf/sdformat/blob/95500e0e75f75a1e5ab69e4b264138cbe78f477a/src/parser_TEST.cc#L670 should I set the `HOME` variable to the build directory, then after `RUN_ALL_TESTS()` set it back to what it was originally? It shouldn't affect the other tests in `parser_TEST`

azeey: Yeah, and I don't think you need to reset it either since the test is done at that point. Setting the environment variable here shouldn't affect the environment variable of the shell.

jennuine: Done 95500e0e75f75a1e5ab69e4b264138cbe78f477a

src/parser_TEST.cc
jennuine: 95500e0e75f75a1e5ab69e4b264138cbe78f477a

src/parser.cc
azeey: https://github.com/osrf/sdformat/issues/409

src/parser_TEST.cc
```diff
@@ -15,9 +15,13 @@
  *
  */
 
+#include <sstream>
+#include <fstream>
+#include <stdlib.h>
```
azeey, line 20: nit: stdlib.h -> cstdlib

src/parser_TEST.cc
```diff
@@ -59,6 +63,31 @@ sdf::SDFPtr InitSDF()
...
+TEST(Parser, CustomUnknownElements)
+{
+  std::string pathBase = PROJECT_SOURCE_PATH;
+  pathBase += "/test/sdf";
+  const std::string path = pathBase +"/custom_and_unknown_elements.sdf";
+
+  sdf::SDFPtr sdf = InitSDF();
+  EXPECT_TRUE(sdf::readFile(path, sdf));
+
+  std::string pathLog = PROJECT_BINARY_DIR;
```
azeey, line 77: I wonder if using `sdf::filesystem::append` here would fix the windows CI failure.

jennuine: Unfortunately it did not fix the Windows CI failure :confused: I'm wondering if it has something to do with `setenv` not actually being set. When using `setenv` in Windows, it should be using this right? https://github.com/osrf/sdformat/blob/00937472e6c3f70b0efee04341f48b640d882eae/test/test_config.h.in#L33-L36

src/parser_TEST.cc
```diff
@@ -59,6 +64,31 @@ sdf::SDFPtr InitSDF()
...
+{
+  std::string pathBase = PROJECT_SOURCE_PATH;
+  pathBase += "/test/sdf";
+  const std::string path = pathBase +"/custom_and_unknown_elements.sdf";
+
+  sdf::SDFPtr sdf = InitSDF();
+  EXPECT_TRUE(sdf::readFile(path, sdf));
+
+  std::string pathLog =
+    sdf::filesystem::append(PROJECT_BINARY_DIR, "/.sdformat/sdformat.log");
```
azeey, line 79: How about?
```suggestion
    sdf::filesystem::append(PROJECT_BINARY_DIR, ".sdformat", "sdformat.log");
```

jennuine: Still failing :confused:

azeey (approved):
LGTM! Thanks for taking the time to debug the windows failure.

PR #406, opened 2020-10-31
Use internal urdf library.

mahiuchun:
This make sdformat buildable regardless of whether liburdfdom-dev is installed or not.

urdf/ folder is replaced with sdf10 branch version as it had incomplete tinyxml2 port.

Unused //ign_tools dependencies removed as it causes some red herring build errors.

### DIFF
--- a/.github/packages.apt
+++ b/.github/packages.apt
@@ -7,4 +7,3 @@ libxml2-utils
 ruby-dev
 python-psutil
 cppcheck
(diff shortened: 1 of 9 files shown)


mjcarroll:
@mahiuchun take a look at #408 which will supercede this as well as land on the upstream `sdf10` branch.

mahiuchun:
Ack, thanks.

PR #410, opened 2020-11-04
Add <double_sided> to material spec

iche033:
Add new SDF parameter for configuring a material's double sided property. If set to true, the mesh will be rendered as double sided,  i.e. every polygon of a mesh will be visible from both the front and back sides. The default value is false.


chapulina (approved):

luca-della-vedova (approved):
Looks good, thanks!

Issue #202, opened 2018-09-14
bad_any_cast after Element::GetAny with gz11 on homebrew


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/202)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

----------------------------------------
There are [two failing tests](https://build.osrfoundation.org/job/sdformat-ci-pr_any-homebrew-amd64-mojave/1/testReport/) for the `gz11` branch on macOS mojave.

* https://build.osrfoundation.org/job/sdformat-ci-pr_any-homebrew-amd64-mojave/1/testReport/(root)/SDF/UpdateElement/
* https://build.osrfoundation.org/job/sdformat-ci-pr_any-homebrew-amd64-mojave/1/testReport/(root)/SDF/GetAny/

I started by investigating the `GetAny` test failure from `UNIT_SDF_TEST`, which failed at [line 418](https://github.com/osrf/sdformat/blob/3826e3878b15598634df332d0c93f7186a506f49/src/SDF_TEST.cc#L418), which occurs when a `std::bad_any_cast` is thrown after calling `std::any_cast<ignition::math::Pose3d>(anyValue)`. I added some print statements to confirm that `anyValue` should have the right type:

~~~
diff -r 3826e3878b15598634df332d0c93f7186a506f49 src/SDF_TEST.cc
--- a/src/SDF_TEST.cc	Fri Sep 14 00:49:58 2018 +0000
+++ b/src/SDF_TEST.cc	Thu Sep 13 17:59:35 2018 -0700
@@ -410,6 +410,13 @@
     std::any anyValue = poseElem->GetAny();
     try
     {
+      using Pose = ignition::math::Pose3d;
+      std::cerr << typeid(Pose).name()
+                << ", " << typeid(Pose).hash_code()
+                << std::endl;
+      std::cerr << anyValue.type().name()
+                << ", " << anyValue.type().hash_code()
+                << std::endl;
       EXPECT_EQ(std::any_cast<ignition::math::Pose3d>(anyValue),
           ignition::math::Pose3d(0, 1, 2, 0, 0, 0));
     }
~~~

which gives the following output, showing that the types have matching names, but different `hash_code`s.

~~~
[...]

PR #414, opened 2020-11-07
Enable tests for #202 on macOS with Xcode 12+, add catalina workflow

scpeters:
I believe the clang bug causing #202 has been fixed in Xcode 12+, which is available on macOS 10.15 Catalina. This re-enables those disabled tests for Xcode 12+ and adds a GitHub workflow for macOS Catalina to confirm the fix.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,45 @@
+name: macOS latest
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    env:
+      PACKAGE: sdformat8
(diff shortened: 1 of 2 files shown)


.github/workflows/macos.yml
```diff
@@ -0,0 +1,45 @@
+name: macOS latest
```
chapulina, line 1: Sweet, we should use this template for the other libraries too! https://github.com/ignition-tooling/action-ignition-ci/issues/25

chapulina (approved):

PR #422, opened 2020-11-21
Test included model folders missing model.config

scpeters:
Currently there is a confusing error message if a file is loaded that finds a folder matching the name of a model to
be included but the folder does not have a model.config file. This improves the first error message and stops further loading
to prevent additional confusing messages. To illustrate this, I've added a model folder `box_missing_config` to `test/integration/model` containing a `model.sdf` but no `model.config` and a test that tries to include a model matching that folder name. Without the change to parser.cc, the test prints the following:

~~~
[ RUN      ] IncludesTest.IncludeModelMissingConfig
Error [parser.cc:749] Could not find model.config or manifest.xml for the model
Error [parser.cc:749] Could not find model.config or manifest.xml for the model
Error [parser.cc:397] File [] doesn't exist.
Error:   Could not find the 'robot' element in the xml file
         at line 80 in /build/urdfdom-YMMa9X/urdfdom-1.0.0/urdf_parser/src/model.cpp
Error [parser_urdf.cc:3193] Unable to call parseURDF on robot model
Error [parser.cc:488] parse as old deprecated model file failed.
Error Code 1 Msg: Unable to read file[]
Error Code 8 Msg: Error reading element <sdf>
/home/scpeters/clone/sdformat/test/integration/includes.cc:337: Failure
Value of: sdf::readString(stream.str(), sdfParsed)
  Actual: false
Expected: true
[  FAILED  ] IncludesTest.IncludeModelMissingConfig (51 ms)
~~~

Note that `box_missing_config` is not listed anywhere, which makes it hard to figure out what is the cause of the problem. It also causes `readString` to fail. After modifying parser.cc, the parser doesn't fail and the test shows the following console output (the second message comes from printing out the errors returned by `Root::Load`):

~~~
[ RUN      ] IncludesTest.IncludeModelMissingConfig
Error [parser.cc:749] Could not find model.config or manifest.xml in [/home/scpeters/clone/sdformat/test/integration/model/box_missing_config]
Error Code 12 Msg: Unable to resolve uri[box_missing_config] to model path [/home/scpeters/clone/sdformat/test/integration/model/box_missing_config] since it does not contain a model.config file.
[       OK ] IncludesTest.IncludeModelMissingConfig (34 ms)
~~~

EDIT: it doesn't actually cause `sdf::readString` to fail, though it does return / print an error depending on which overload is called.

### DIFF
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -746,7 +746,8 @@ std::string getModelFilePath(const std::string &_modelDirPath)
     if (!sdf::filesystem::exists(configFilePath))
     {
       // We didn't find manifest.xml either, output an error and get out.
-      sdferr << "Could not find model.config or manifest.xml for the model\n";
+      sdferr << "Could not find model.config or manifest.xml in ["
+             << _modelDirPath << "]\n";
       return std::string();
     }
     else
@@ -932,6 +933,15 @@ bool readXml(TiXmlElement *_xml, ElementPtr _sdf, Errors &_errors)
 
           // Get the config.xml filename
           filename = getModelFilePath(modelPath);
+
+          if (filename.empty())
+          {
+            _errors.push_back({ErrorCode::URI_LOOKUP,
+                "Unable to resolve uri[" + uri + "] to model path [" +
+                modelPath + "] since it does not contain a model.config " +
+                "file."});
+            continue;
+          }
         }
         else
         {
--- a/test/integration/includes.cc
+++ b/test/integration/includes.cc
@@ -336,7 +336,10 @@ TEST(IncludesTest, IncludeModelMissingConfig)
   sdf::init(sdfParsed);
   sdf::Errors errors;
   ASSERT_TRUE(sdf::readString(stream.str(), sdfParsed, errors));
+
+  ASSERT_GE(1u, errors.size());
   EXPECT_EQ(1u, errors.size());
+  std::cout << errors[0] << std::endl;
   EXPECT_EQ(errors[0].Code(), sdf::ErrorCode::URI_LOOKUP);
   EXPECT_NE(std::string::npos, errors[0].Message().find(
       "Unable to resolve uri[box_missing_config] to model path")) << errors[0];
--- a/test/integration/includes.cc
+++ b/test/integration/includes.cc
@@ -334,10 +334,18 @@ TEST(IncludesTest, IncludeModelMissingConfig)
 
   sdf::SDFPtr sdfParsed(new sdf::SDF());
   sdf::init(sdfParsed);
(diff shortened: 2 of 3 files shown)


src/parser.cc
```diff
@@ -932,6 +933,15 @@ bool readXml(TiXmlElement *_xml, ElementPtr _sdf, Errors &_errors)
 
           // Get the config.xml filename
           filename = getModelFilePath(modelPath);
+
+          if (filename.empty())
+          {
+            _errors.push_back({ErrorCode::URI_LOOKUP,
```
chapulina, line 939: This error code seems like a bit of a stretch in this situation, because there isn't really a URI that the user provided and couldn't be found.  But the message is descriptive enough.

test/integration/includes.cc
```diff
@@ -318,3 +318,29 @@ TEST(IncludesTest, Includes_15_convert)
...
+
+  sdf::SDFPtr sdfParsed(new sdf::SDF());
+  sdf::init(sdfParsed);
+  ASSERT_TRUE(sdf::readString(stream.str(), sdfParsed));
+
+  sdf::Root root;
+  sdf::Errors errors = root.Load(sdfParsed);
+  for (auto e : errors)
+    std::cout << e.Message() << std::endl;
+  EXPECT_TRUE(errors.empty());
```
chapulina, line 343: How about checking the error code and the exact number of errors?

src/parser.cc
scpeters: do you think `URI_INVALID` is better? I wasn't sure what to put

chapulina: Both the `URI` errors sound to me like there's a mistake within an SDF file. But I don't feel too strongly about this. Do you think adding a new error code would be too much? Something like `MISSING_CONFIG`?

test/integration/includes.cc
scpeters: the error was actually coming from `sdf::readString`, because I was using an overload of `sdf::readString` that printed `sdf::Errors` instead of returning them. I've switched APIs and added expectations for the error type and message in 5bf57af2a6b00d3de29fba9bea4858ea54177e92.

src/parser.cc
scpeters: `URI_LOOKUP` is used earlier in this same block of code if a model matching the specified `uri` can't be found, so I think it is consistent to use it here as well

* https://github.com/osrf/sdformat/blob/sdformat9_9.3.0/src/parser.cc#L912

chapulina: Ok, it makes sense to me then. The URI specifies the path that must contain both `.sdf` and `.config` files.

chapulina (approved):

azeey (approved):

PR #426, opened 2020-12-01
Make sdf::Model::CanonicalLinkAndRelativeName public

azeey:
This member function is the only way to obtain the relative name of the canonical link when the link is implicitly nested, i.e, the `//model/@canonical_link` attribute is empty and the canonical link resolves to a link nested in a child model. While `sdf::Model::CanonicalLink` can be used to resolve the link, the returned `sdf::Link*` can only provide the local (non-relative) name of the link.

When constructing models, a downstream application may first recursively construct internal representations of all the links of a model before creating an internal representation of a model. When doing so, the application would need to identify one of the internal links as the canonical link and associate it with the internal representation of the model. In order to use `sdf::Model::CanonicalLink` for this task, it becomes necessary to track the mapping between `sdf::Link*` and the internal links of the application. Whereas if the relative name of the canonical link is available via `sdf::Model::CanonicalLinkAndRelativeName`, the application can query its internal list of links by the the relative name. It's reasonable to assume that the mapping from name to internal link is maintained by any application that consumes libsdformat so using `sdf::Model::CanonicalLinkAndRelativeName` avoids extra overhead.

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -20,6 +20,9 @@ but with improved human-readability..
     + Errors ResolveChildLink(std::string&) const
     + Errors ResolveParentLink(std::string&) const
 
+1. **sdf/Model.hh**:
+    + std::pair<const Link *, std::string> CanonicalLinkAndRelativeName() const;
+
 ### Modifications
 
 1. **sdf/Model.hh**: the following methods now accept nested names relative to
--- a/include/sdf/Model.hh
+++ b/include/sdf/Model.hh
@@ -314,6 +314,13 @@ namespace sdf
     /// \param[in] _name Name of the placement frame.
     public: void SetPlacementFrameName(const std::string &_name);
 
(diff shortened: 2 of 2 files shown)


scpeters:
sounds reasonable to me; I initially made it private to avoid exposing `::` syntax in `sdf9`, but I think it's fine to make it public in 11. aside from adding it to the migration guide; this look good to me

azeey:
> sounds reasonable to me; I initially made it private to avoid exposing `::` syntax in `sdf9`, but I think it's fine to make it public in 11. aside from adding it to the migration guide; this look good to me

I have updated the migration guide in https://github.com/osrf/sdformat/pull/426/commits/031ffbb91b53a23d18d8e349f3b5d06b594c5d5a. I put it in the "Additions" section since it's being added to the public API. Let me know if I should put it in "Modifications" instead.

scpeters (approved):

PR #431, opened 2020-12-07
Add contributing guide and code of conduct

chapulina:
Copied the code of conduct that we have on some Ignition libraries. I don't think it will change much, so we've been copying it instead of linking to a central place.

For the contribution guide, I linked to the central location like all Ignition libraries do. Since SDFormat isn't technically an "Ignition" library, this may be a bit odd though. Not sure what others think.


azeey (approved):
LGTM. I think it's fine to say we use the Ignition contributing guide.

PR #434, opened 2020-12-11
Add ellipsoid shape to sdf

brawner:
For the second part of https://github.com/osrf/sdformat/issues/376 and following up #389, this PR adds the ellipsoid simple shape. It is defined by a 3-component vector of doubles for its radii.

This wasn't branched off from #389, but it will need to be rebased on top of it to address a couple of related changes.

Depends directly on https://github.com/ignitionrobotics/ign-math/pull/182 

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/include/sdf/Ellipsoid.hh
+++ b/include/sdf/Ellipsoid.hh
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2020 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+#ifndef SDF_ELLIPSOID_HH_
+#define SDF_ELLIPSOID_HH_
+
+#include <ignition/math/Ellipsoid.hh>
+#include <sdf/Error.hh>
+#include <sdf/Element.hh>
+#include <sdf/sdf_config.h>
+
+namespace sdf
+{
+  // Inline bracket to help doxygen filtering.
+  inline namespace SDF_VERSION_NAMESPACE {
+  //
+
+  // Forward declare private data class.
+  class EllipsoidPrivate;
+
+  /// \brief Ellipsoid represents a ellipsoid shape, and is usually accessed
+  /// through a Geometry.
+  class SDFORMAT_VISIBLE Ellipsoid
+  {
+    /// \brief Constructor
+    public: Ellipsoid();
+
+    /// \brief Copy constructor
+    /// \param[in] _ellipsoid Ellipsoid to copy.
+    public: Ellipsoid(const Ellipsoid &_ellipsoid);
+
+    /// \brief Move constructor
+    /// \param[in] _ellipsoid Ellipsoid to move.
+    public: Ellipsoid(Ellipsoid &&_ellipsoid) noexcept;
+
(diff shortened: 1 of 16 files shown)


include/sdf/Geometry.hh
```diff
@@ -56,6 +57,9 @@ namespace sdf
 
     /// \brief A mesh geometry.
     MESH = 5,
+
+    /// \brief An ellipsoid geometry
+    ELLIPSOID = 6,
```
scpeters, line 62: I think this will conflict with the `HEIGHTMAP` value proposed in #388 (targeting `libsdformat9`); that's why I used `7` for `CAPSULE` in #389

brawner: Ahh, that makes sense. Setting it to 8 in latest commit.

brawner:
Rebasing onto master and resolving conflicts

scpeters:
the following is needed for Ubuntu CI: https://github.com/ignition-tooling/gzdev/pull/26

scpeters:

scpeters:
we should require version 6.8 of ignition-math6 in cmake/SearchForStuff.cmake

I guess we also aren't using gzdev in the GitHub workflow for this package, so it's not picking up the prereleases

azeey (approved):
LGTM once CI is green

include/sdf/Ellipsoid.hh
```diff
@@ -0,0 +1,97 @@
...
+
+    /// \brief Load the ellipsoid geometry based on a element pointer.
+    /// This is *not* the usual entry point. Typical usage of the SDF DOM is
+    /// through the Root object.
+    /// \param[in] _sdf The SDF Element pointer
+    /// \return Errors, which is a vector of Error objects. Each Error includes
+    /// an error code and message. An empty vector indicates no error.
+    public: Errors Load(ElementPtr _sdf);
+
+    /// \brief Get the ellipsoid's radius in meters.
```
chapulina, line 70: Nit:

```suggestion
    /// \brief Get the ellipsoid's radii in meters.
```

Also below

include/sdf/Geometry.hh
```diff
@@ -141,6 +145,17 @@ namespace sdf
     /// \param[in] _cylinder The cylinder shape.
     public: void SetCylinderShape(const Cylinder &_cylinder);
 
+    /// \brief Get the ellipsoid geometry, or nullptr if the contained
+    /// geometry is not an ellipsoid.
+    /// \return Pointer to the visual's ellipsoid geometry, or nullptr if the
```
chapulina, line 150: I know that this is probably repeating the pattern above, but I'll just point out that the geometry may be attached to a collision, not necessarily a visual.

src/Ellipsoid.cc
```diff
@@ -0,0 +1,148 @@
...
+*/
+#include <sstream>
+#include "sdf/Ellipsoid.hh"
+
+using namespace sdf;
+
+// Private data class
+class sdf::EllipsoidPrivate
+{
+  // An ellipsoid with a three radiuses of 1 meter
```
chapulina, line 25:
```suggestion
  /// \brief An ellipsoid with all three radii of 1 meter
```

chapulina (approved):
LGTM, I just have minor comments about documentation.

Because this is targeting `master`, I think it can get in using `ign-math`'s pre-release, just like how other libraries are using some nightlies. We just need to make sure we make the stable `ign-math 6.8` release before libSDFormat 11.0.0 is released.

include/sdf/Ellipsoid.hh
brawner: Fixed

include/sdf/Geometry.hh
brawner: Adjusted wording in this comment

src/Ellipsoid.cc
brawner: Fixed

brawner (commented):
Rebased onto master and addressed pr feedback. The last build succeeded, so I'm pretty confident about this one.

brawner:
This added the prereleases in the github action workflow, which should be removed before a stable release is created.

PR #435, opened 2020-12-11
Decrease far clip lower bound

nkoenig:
A far clip plane for a camera can be less than 10m. I'm setting the minimum value to match the value for the depth_camera.


chapulina (approved):

nkoenig:
The test failures look unrelated to this change.

PR #438, opened 2020-12-15
Re-disable tests for #202

scpeters:
This reverts part of "Enable tests for #202 on macOS with Xcode 12+, add catalina workflow (#414)", since those tests have started failing again in the macOS CI action. It keeps the macOS workflow.

* https://github.com/osrf/sdformat/actions?query=workflow%3A%22macOS+latest%22

I don't know what has changed or why, but I've re-opened #202 and am disabling the tests again here.


scpeters:
merging without review to fix failing tests and unblock other PR's

Issue #373, opened 2020-09-10
Stop using global variables


scpeters:
There are several places where global variables are used in `libsdformat`, which makes it difficult to control behavior when there are multiple paths that parse SDFormat files. As @jwnimmer-tri noted in https://github.com/osrf/sdformat/issues/334#issuecomment-683985626:

> parsing should always be done against a parsing context or configuration, so that using-code can pass down customizations to any given request to parse, without having process-wide globals. We could add more argument to `LoadSdfString` but that gets awkward when there are too many. A parsing object would help consolidate (and reuse) the configuration.

> In the long term, the best outcome is that libsdformat posts its messages (calls a function) on some logger interface or infrastructure, so that downstream code can easily capture / prefix / redirect / threshold the messages. That might even be better formulated as a "parsing error" interface, rather than unstructured text logging. A parsing library really has no business just remarking plain text into a console -- if the input text is malformed, then libsdformat should be reporting those errors in a structured way (with input line and column numbers, severity, configuration of strict vs permissive parsing, etc.) so that application code can inform the user in a way consistent with the application's behavior.

### Places where global variables are used:

* https://github.com/osrf/sdformat/blob/sdformat10_10.0.0-pre2/src/Console.cc#L31-L42
* https://github.com/osrf/sdformat/blob/sdformat10_10.0.0-pre2/src/SDF.cc#L43-L47
* https://github.com/osrf/sdformat/blob/sdformat10_10.0.0-pre2/src/parser_urdf.cc#L53-L63

EricCousineau-TRI:
@azeey Do you think we should close this out, or keep it open to simplify the number of overloads in the future (via a tick-tock)?

azeey:
I vote for closing this out and creating a new issue for addressing the number of overloads.

EricCousineau-TRI:
Closed via #439

jwnimmer-tri:
Here are some globals still found on the head of master.  Maybe I checked the wrong branch?

https://github.com/osrf/sdformat/blob/f59a342469bd9c71bd2dc5b325d777164375b06a/src/SDF.cc#L40
https://github.com/osrf/sdformat/blob/f59a342469bd9c71bd2dc5b325d777164375b06a/src/parser_urdf.cc#L52-L63
https://github.com/osrf/sdformat/blob/f59a342469bd9c71bd2dc5b325d777164375b06a/src/Console.cc#L43

jwnimmer-tri:
[...]

PR #439, opened 2020-12-15
Add ParserConfig class to encapsulate file path settings

azeey:
Toward #373 

This adds a new `ParserConfig` object that contains configuration options for the libsdformat parser. This replaces the old globals `g_uriPathMap` and `g_findFileCB`. To do this in a backward compatible way, the `ParserConfig` class provides a singleton object that calls to `sdf::setSdfFindCallback` and `sdf::addURIPath` use. 

New overloads of the main SDFormat file and string loading APIs allow users to pass in a `ParserConfig` object. If omitted, these APIs will default to the singleton `ParserConfig` object.


include/sdf/ParserConfig.hh
```diff
@@ -0,0 +1,130 @@
...
+///
+///   // Use the new config when loading a new SDFormat file
+///   sdf::Root root;
+///   root.Load("path/to/file.sdf", config);
+/// \endcode
+class SDFORMAT_VISIBLE ParserConfig
+{
+  /// type alias for the map from URI scheme to search directories
+  public: using SchemeToPathMap =
+          std::map<std::string, std::list<std::string> >;
```
brawner, line 71: Out of curiosity, why a std::list instead of std::vector?

brawner (commented):
I don't feel confident enough in the code base to approve/request changes yet, so I'll leave that to other reviewers. But I did have a few comments.

My main question is whether the `DefaultConfig()` should be a mutable singleton. Typically a default cannot be changed by client code, so I would expect the API to return a `static const ParserConfig &`. What were the tradeoffs compared to passing around a reference to a config object in the parsing code, with `sdf::Root::Load()` and related functions taking a default argument?

If the mutable singleton is the best approach here, what would you think about changing it's name to `GlobalConfig()`?

src/ParserConfig.cc
```diff
@@ -0,0 +1,117 @@
...
+  return this->dataPtr->uriPathMap;
+}
+
+/////////////////////////////////////////////////
+void ParserConfig::AddURIPath(const std::string &_uri, const std::string &_path)
+{
+  // Split _path on colons.
+  // Add each part of the colon separated path to the global URI map.
+#ifdef _WIN32
+  const char *multiplePathSeparator = ";";
```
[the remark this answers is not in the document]
azeey: Done in https://github.com/osrf/sdformat/pull/439/commits/f9e3ef294641cf7f963b688789b45684582c4b02

include/sdf/ParserConfig.hh
azeey: This was copied form https://github.com/osrf/sdformat/blob/master/src/SDF.cc#L40 but I also think `std::vector` makes more sense. Changed in https://github.com/osrf/sdformat/pull/439/commits/f9e3ef294641cf7f963b688789b45684582c4b02

azeey (commented):
> If the mutable singleton is the best approach here, what would you think about changing it's name to GlobalConfig()?

I thought a mutable singleton was needed to maintain backward compatibility. The functions [`sdf::setFindFileCallback`](https://github.com/osrf/sdformat/blob/master/src/SDF.cc#L51) and [`sdf::addURIPath`](https://github.com/osrf/sdformat/blob/master/src/SDF.cc#L177) used to modify the globals `g_findFileCB` and `g_uriPathMap` respectively. Now they modify the singleton. But I do agree that `GlobalConfig` is a much better name. Renamed in https://github.com/osrf/sdformat/pull/439/commits/deda523ac72d463391ccac2b0ee49b6c7e8bef2e

azeey:

azeey:
Not sure why the Jenkins CI status is not updating even though they are finished:
  * Ubuntu [Build Status](https://build.osrfoundation.org/job/sdformat-ci-pr_any-ubuntu_auto-amd64/1369/)
  * Homebrew [Build Status](https://build.osrfoundation.org/job/sdformat-ci-pr_any-homebrew-amd64/1635/)
  * Windows [Build Status](https://build.osrfoundation.org/job/sdformat-ci-pr_any-windows7-amd64/1749/)

EricCousineau-TRI:
May need rebase or merge (again)?
> image

include/sdf/parser.hh
```diff
@@ -40,22 +40,60 @@ namespace sdf
   //
   class Root;
 
-  /// \brief Init based on the installed sdf_format.xml file
```
EricCousineau-TRI, line 43: BTW This is a large amount of combinatorics... But I don't have a better solution.

include/sdf/parser.hh
```diff
@@ -144,13 +238,27 @@ namespace sdf
   /// string it is converted to SDF first. All strings are converted to the
   /// latest SDF version
   /// \param[in] _xmlString XML string to be parsed.
-  /// \param[in] _sdf Pointer to an SDF object.
+  /// \param[out] _sdf Pointer to an sdf Element object.
```
EricCousineau-TRI, line 241: nit Case "sdf" is inconsistent with other usages "SDF"

(Here and elsewhere)

include/sdf/ParserConfig.hh
```diff
@@ -0,0 +1,130 @@
...
+#include "sdf/system_util.hh"
+
+namespace sdf
+{
+inline namespace SDF_VERSION_NAMESPACE
+{
+// Forward declare private data class.
+class ParserConfigPrivate;
+
+/// This class contains configuration options for the libsdformat parser.
```
EricCousineau-TRI, line 37: In addition to below comment (how path resolution works), it's unclear to me what the precedence is for these different attributes (e.g. does callback override search paths, or the other way around).

Is there documentation that this can forward-reference, or should it be written here?

include/sdf/SDFImpl.hh
```diff
@@ -64,6 +65,21 @@ namespace sdf
                        bool _searchLocalPath = true,
                        bool _useCallback = false);
 
+  /// \brief Find the absolute path of a file.
```
EricCousineau-TRI, line 68: This new function, and the old flavor, don't fully specify how the searching actually takes place. (I understand it's implementation specific, but still.)

From `_useCallback`, I infer that "normal mechanism" means `_searchLocalPath`, but I don't know it for sure.
Can you make this text more concise, and tell what the priorities are?

include/sdf/ParserConfig.hh
```diff
@@ -0,0 +1,130 @@
...
+  /// \brief Set the callback to use when libsdformat can't find a file.
+  /// The callback should return a complete path to the requested file, or
+  /// and empty string if the file was not found in the callback.
+  /// \param[in] _cb The callback function.
+  public: void SetFindCallback(
+              std::function<std::string(const std::string &)> _cb);
+
+  /// \brief Get the URI scheme to search directories map
+  /// \return Immutable reference to the URI scheme to search directories map
+  public: const SchemeToPathMap &URIPathMap() const;
```
EricCousineau-TRI, line 116: nit It smells weird to be able to look at all paths, add a new path, but not remove a path. (assymmetry)

Is it hard to add that here?

Or perhaps the right workflow is to just reassign? e.g. just clear it out.

So if I wanted to remove or reorder a path, I would just do something like `Parser::GLobalConfig() = my_new_config`?

test/integration/parser_config.cc
```diff
@@ -0,0 +1,240 @@
+/*
```
EricCousineau-TRI, line 1: Are there any tests which explicitly show shadowing / prioritization?
e.g. both a search path and a callback could resolve the same file (possibly to different places), but only one of them wins out?

EricCousineau-TRI (commented):
First pass done, PTAL!

include/sdf/ParserConfig.hh
```diff
@@ -0,0 +1,130 @@
...
+  public: static ParserConfig &GlobalConfig();
+
+  /// \brief Get the find file callback function
+  /// \return Immutable reference to the find file callback function
+  public: const std::function<std::string(const std::string &)> &
+          FindFileCallback() const;
+
+  /// \brief Set the callback to use when libsdformat can't find a file.
+  /// The callback should return a complete path to the requested file, or
+  /// and empty string if the file was not found in the callback.
```
scpeters, line 109: "and empty string" -> "an empty string"

include/sdf/ParserConfig.hh
```diff
@@ -0,0 +1,130 @@
...
+  /// \return A mutable reference to the singleton ParserConfig object
+  public: static ParserConfig &GlobalConfig();
+
+  /// \brief Get the find file callback function
+  /// \return Immutable reference to the find file callback function
+  public: const std::function<std::string(const std::string &)> &
+          FindFileCallback() const;
+
+  /// \brief Set the callback to use when libsdformat can't find a file.
+  /// The callback should return a complete path to the requested file, or
```
scpeters, line 108: can you be more explicit about what the input argument to the callback function should be? I believe it can be a file name with no path qualifiers at all, but can it be a relative path? I doubt it would ever be an absolute path based on "when libsdformat can't find a file", but it would be helpful to be explicit about what are the expected inputs for that callback function

include/sdf/parser.hh
```diff
@@ -40,22 +40,60 @@ namespace sdf
...
   bool initString(const std::string &_xmlString, SDFPtr _sdf);
 
+  /// \brief Initialize the SDF interface using a string
+  /// \param[in] _xmlString XML string to be parsed.
+  /// \param[in] _config Custom parser configuration
+  /// \param[out] _sdf Pointer to an SDF object.
+  /// \return True if successful.
+  SDFORMAT_VISIBLE
+  bool initString(
+      const std::string &_xmlString, SDFPtr _sdf, const ParserConfig &_config);
```
scpeters, line 95: nit: the `initFile` overloads put the `ParserConfig` in the middle and the output last, while this overload puts the output in the middle

scpeters (commented):
there is a `SDFPtr readFile(const std::string &)` at parser.hh:105 that didn't get overloaded

also, #474 was just merged, so you can use ImplPtr in ParserConfig

azeey:
> there is a SDFPtr readFile(const std::string &) at parser.hh:105 that didn't get overloaded

I was trying to have one less overload by only overloading the signature that also contains an `sdf::Error`. I did the same for `readString`. 

Here's a list of all the functions in `parser.hh` that work with `ParserConfig`. I think it helps to see them without the documentation:
```c++
bool init(SDFPtr _sdf);

bool initFile(const std::string &_filename, SDFPtr _sdf);
bool initFile(const std::string &_filename, const ParserConfig &_config, SDFPtr _sdf);

bool initFile(const std::string &_filename, ElementPtr _sdf);
bool initFile(const std::string &_filename, const ParserConfig &_config, ElementPtr _sdf);

bool initString(const std::string &_xmlString, SDFPtr _sdf);
bool initString(const std::string &_xmlString, const ParserConfig &_config, SDFPtr _sdf);

sdf::SDFPtr readFile(const std::string &_filename);
sdf::SDFPtr readFile(const std::string &_filename, Errors &_errors);
sdf::SDFPtr readFile(const std::string &_filename, const ParserConfig &_config, Errors &_errors);

bool readFile(const std::string &_filename, SDFPtr _sdf, Errors &_errors);
bool readFile(const std::string &_filename, const ParserConfig &_config, SDFPtr _sdf, Errors &_errors);

bool readFileWithoutConversion(const std::string &_filename, SDFPtr _sdf, Errors &_errors);
bool readFileWithoutConversion(const std::string &_filename, const ParserConfig &_config, SDFPtr _sdf, Errors &_errors);

bool readFile(const std::string &_filename, SDFPtr _sdf);

bool readString(const std::string &_xmlString, SDFPtr _sdf);
bool readString(const std::string &_xmlString, SDFPtr _sdf, Errors &_errors);
bool readString(const std::string &_xmlString, const ParserConfig &_config, SDFPtr _sdf, Errors &_errors);

bool readString(const std::string &_xmlString, ElementPtr _sdf);
bool readString(const std::string &_xmlString, ElementPtr _sdf, Errors &_errors);
bool readString(const std::string &_xmlString, const ParserConfig &_config, ElementPtr _sdf, Errors &_errors);

bool readStringWithoutConversion( const std::string &_xmlString, SDFPtr _sdf, Errors &_errors);
bool readStringWithoutConversion(const std::string &_xmlString, const ParserConfig &_config, SDFPtr _sdf, Errors &_errors);

bool convertFile(const std::string &_filename, const std::string &_version, SDFPtr _sdf);
bool convertFile(const std::string &_filename, const std::string &_version, const ParserConfig &_config, SDFPtr _sdf);

bool convertString(const std::string &_sdfString, const std::string &_version, SDFPtr _sdf);
bool convertString(const std::string &_sdfString, const std::string &_version, const ParserConfig &_config, SDFPtr _sdf);
```

It's interesting that we have `sdf::SDFPtr readFile(const std::string &_filename);` and `bool readString(const std::string &_xmlString, SDFPtr _sdf);`. I'm guessing the latter was added along with `bool readString(const std::string &_xmlString, ElementPtr _sdf);`. Should we deprecate `sdf::SDFPtr readFile(const std::string &_filename);`?

include/sdf/ParserConfig.hh
azeey: I don't think there's any limitation as to what the input argument would be. It really depends on the users callback function. `sdf::findFile` calls the function if all other ways of searching for the file fail. This means it can even be called with absolute paths if the file pointed to by the path doesn't exist on the filesystem.

include/sdf/ParserConfig.hh
azeey: I was trying to make this PR replicate existing functionality without adding any new features other than being able to specify a non-global config. It would be easy to add a function to remove a URI scheme, but adding a path to an existing URI scheme or reordering paths seem best suited for a future PR, IMO.

include/sdf/parser.hh
azeey: e42bbda

include/sdf/ParserConfig.hh
azeey: I added some documentation in SDFImpl.hh (https://github.com/osrf/sdformat/pull/439/commits/6f53a9f113b1d2b8e4749ee5a55f76913b9f162b) and referenced it here in https://github.com/osrf/sdformat/pull/439/commits/1d321e01b8a1b5befb6097c4eac7c512b0c769f4

include/sdf/SDFImpl.hh
azeey: I don't know about more concise, but I listed the search order in https://github.com/osrf/sdformat/commit/6f53a9f113b1d2b8e4749ee5a55f76913b9f162b. How does that look?

test/integration/parser_config.cc
azeey: Added test in https://github.com/osrf/sdformat/pull/439/commits/8977cd1111165a0b2f13c902e96f960f96d1cae9

include/sdf/ParserConfig.hh
```diff
@@ -0,0 +1,130 @@
...
+  /// \brief Copy assignment operator.
+  /// \param[in] _config ParserConfig to copy.
+  /// \return Reference to this.
+  public: ParserConfig &operator=(const ParserConfig &_config);
+
+  /// \brief Destructor
+  public: ~ParserConfig();
+
+  /// Mutable access to a singleton ParserConfig that serves as the global
+  /// ParserConfig object for all parsing operations
```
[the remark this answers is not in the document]
azeey: Sure. https://github.com/osrf/sdformat/pull/439/commits/fde3b1bf99776c5eecd921200324eea391cf806b

include/sdf/ParserConfig.hh
azeey: https://github.com/osrf/sdformat/pull/439/commits/fde3b1bf99776c5eecd921200324eea391cf806b

include/sdf/parser.hh
azeey: Changed the order in https://github.com/osrf/sdformat/pull/439/commits/fde3b1bf99776c5eecd921200324eea391cf806b.

include/sdf/parser.hh
azeey: Yeah :(. I think there are some redundancies that we can deprecate, but we still end up with a lot of functions.

include/sdf/ParserConfig.hh
scpeters: I think that's still useful to explain, that it's up to the callback to interpret the input arguments. Please correct me if I'm wrong, but perhaps we could give an example that if a file has text like `custom://model_name` in a `//include/uri` tag, that text should eventually be passed to this callback function if a file with that name can't be found?

scpeters:
> > there is a SDFPtr readFile(const std::string &) at parser.hh:105 that didn't get overloaded
> 
> I was trying to have one less overload by only overloading the signature that also contains an `sdf::Error`. I did the same for `readString`.

that sounds fine to me

> Should we deprecate `sdf::SDFPtr readFile(const std::string &_filename);`?

I would be ok with that. It's not the safest API since it doesn't have any way to report errors

scpeters:
I've reported the windows compiler warning to the ign-utils repository https://github.com/ignitionrobotics/ign-utils/issues/9

include/sdf/ParserConfig.hh
azeey: I added an explanation in https://github.com/osrf/sdformat/pull/439/commits/8a9e4e82e99d5a3760b02a5be56dbae2a02f3008

azeey:

include/sdf/ParserConfig.hh
scpeters: :+1:

test/integration/parser_config.cc
```diff
@@ -0,0 +1,240 @@
...
+
+  const auto &config = sdf::ParserConfig::GlobalConfig();
+  auto it = config.URIPathMap().find("file://");
+  ASSERT_NE(config.URIPathMap().end(), it);
+  ASSERT_EQ(1u, it->second.size());
+  EXPECT_EQ(it->second.front(), testDir);
+
+  ASSERT_TRUE(sdf::ParserConfig::GlobalConfig().FindFileCallback());
+  EXPECT_EQ("test/dir2",
+      sdf::ParserConfig::GlobalConfig().FindFileCallback()("empty"));
```
scpeters, line 53: how do you feel about adding some expectations about `sdf::findFile`?

~~~
diff --git a/test/integration/parser_config.cc b/test/integration/parser_config.cc
index d6bd2fb3..3f9780ac 100644
--- a/test/integration/parser_config.cc
+++ b/test/integration/parser_config.cc
@@ -51,6 +51,9 @@ TEST(ParserConfig, GlobalConfig)
   ASSERT_TRUE(sdf::ParserConfig::GlobalConfig().FindFileCallback());
   EXPECT_EQ("test/dir2",
       sdf::ParserConfig::GlobalConfig().FindFileCallback()("empty"));
+  // sdf::findFile requires explicitly enabling callbacks
+  EXPECT_EQ("test/dir2", sdf::findFile("empty", false, true));
+  EXPECT_EQ("test/dir2", sdf::findFile("empty", true, true));
 }
 
~~~

test/integration/parser_config.cc
```diff
@@ -0,0 +1,240 @@
...
+        return "test/dir2";
+      });
+
+  auto it = config.URIPathMap().find("file://");
+  ASSERT_NE(config.URIPathMap().end(), it);
+  ASSERT_EQ(1u, it->second.size());
+  EXPECT_EQ(it->second.front(), testDir);
+
+  ASSERT_TRUE(config.FindFileCallback());
+  EXPECT_EQ("test/dir2", config.FindFileCallback()("empty"));
```
scpeters, line 81: how do you feel about calling `sdf::findFile` here as well?

~~~
diff --git a/test/integration/parser_config.cc b/test/integration/parser_config.cc
@@ -79,6 +82,8 @@ TEST(ParserConfig, NonGlobalConfig)
 
   ASSERT_TRUE(config.FindFileCallback());
   EXPECT_EQ("test/dir2", config.FindFileCallback()("empty"));
+  EXPECT_EQ("test/dir2", sdf::findFile("empty", false, true, config));
+  EXPECT_EQ("test/dir2", sdf::findFile("empty", true, true, config));
 
   EXPECT_TRUE(sdf::ParserConfig::GlobalConfig().URIPathMap().empty());
   EXPECT_FALSE(sdf::ParserConfig::GlobalConfig().FindFileCallback());
~~~

scpeters (approved):
this is looking good to me; I just have two small suggestions for extra test expectations

include/sdf/SDFImpl.hh
EricCousineau-TRI: Ooh, that's excellent! Thank you!

include/sdf/ParserConfig.hh
EricCousineau-TRI: Aye, makes sense to me.

include/sdf/ParserConfig.hh
EricCousineau-TRI: Yay! Thanks!

EricCousineau-TRI (approved):

azeey:

PR #444, opened 2020-12-15
Set sdf::Root to contain only one model/actor/light

brawner:
This is a followup PR to #433. This changes sdf::Root to have only one Actor/Light/Model.

This change will introduce a hard break to the root element.
Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/src/Root.cc
+++ b/src/Root.cc
@@ -52,15 +52,15 @@ class sdf::RootPrivate
 
   /// \brief The models specified under the root SDF element
   /// Deprecated: to be removed in libsdformat12
-  public: std::vector<sdf::Model> models SDF_DEPRECATED(11.0);
+  public: std::vector<sdf::Model> models;
 
   /// \brief The lights specified under the root SDF element
   /// Deprecated: to be removed in libsdformat12
-  public: std::vector<sdf::Light> lights SDF_DEPRECATED(11.0);
+  public: std::vector<sdf::Light> lights;
 
   /// \brief The actors specified under the root SDF element
   /// Deprecated: to be removed in libsdformat12
-  public: std::vector<sdf::Actor> actors SDF_DEPRECATED(11.0);
+  public: std::vector<sdf::Actor> actors;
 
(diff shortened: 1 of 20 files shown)


azeey (commented):
We usually do a tick-tock cycle on public APIs so I think we need to first deprecate the member functions in `Root` in libsdformat11 and then remove them in libsdformat12.

brawner:
Makes sense, I'll plan on updating this PR

brawner:
Deprecated the original methods and adjusted their logic for the addition of the `unique_ptr`. I also rebased onto master, so this should be easier to review now.

brawner:
In this last commit, I've switched model/light/actor to being stored as part of a `std::variant` so that the implementation better matches the sdf spec. It adds more lines than the unique_ptr variation, but better enforces the spec.

I'm happy to switch back though.

src/Root.cc
```diff
@@ -41,20 +42,32 @@ class sdf::RootPrivate
...
+  /// the first model/light/actor if they exist. When the vectors below are
+  /// removed this should be changed from a variant of pointers to a variant of
+  /// objects
+  public: std::variant<std::monostate, sdf::Model*, sdf::Light*, sdf::Actor*>
+              modelLightOrActor;
+
   /// \brief The models specified under the root SDF element
-  public: std::vector<Model> models;
+  /// Deprecated: to be removed in libsdformat12
+  public: std::vector<sdf::Model> models SDF_DEPRECATED(11.0);
```
brawner, line 55: I'm not sure if deprecating these member variables is useful here, as compiling this file creates a bunch of warnings.

azeey: Yeah, I would only deprecate the public API. Stuff in `RootPrivate` is all private.

azeey (approved):
LGTM! Thanks for iterating. Could you add the deprecations to `Migration.md` before merging?

brawner:
Updated Migration.md

brawner:
Had to update some older commits to fix DCO.

PR #447, opened 2020-12-22
Demonstration of using utils::ImplPtr

mjcarroll:
A demonstration of using the new ignition-utils::ImplPtr: https://github.com/ignitionrobotics/ign-utils/pull/1

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/cmake/SearchForStuff.cmake
+++ b/cmake/SearchForStuff.cmake
@@ -111,3 +111,16 @@ else()
   set(IGN_MATH_VER ${ignition-math6_VERSION_MAJOR})
   message(STATUS "Looking for ignition-math${IGN_MATH_VER}-config.cmake - found")
 endif()
+
+########################################
+# Find ignition utils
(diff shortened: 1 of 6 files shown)


scpeters:
I don't think we should add a dependency to `libsdformat10`, since it's already been released

scpeters:
and I would check with @azeey to see what he thinks

mjcarroll:
> I don't think we should add a dependency to libsdformat10, since it's already been released

I agree, this is a quick draft to show the utility of the ImplPtr.  It should have been targeted to main.

azeey:
Oh this is super nice!! Thanks @mjcarroll

scpeters:
> Oh this is super nice!! Thanks @mjcarroll

yes it is way more elegant

mjcarroll:
Superceded by #472 (and friends)

Issue #430, opened 2020-12-03
Should implement interface API


EricCousineau-TRI:
Towards #278

Should be co-developed with checkbox 2 in https://github.com/RobotLocomotion/drake/issues/14295

PR #448, opened 2020-12-23
Allow files paths for include URIs

azeey:
`//include/uri` used to only take directories that contain a model.config file and the sdformat file. However, allowing `//include/uri` to take file names can be supported without loss of functionality and will allow users to specify non-sdformat files to include via the [Interface API](http://sdformat.org/tutorials?tut=composition_proposal&cat=pose_semantics_docs&#1-5-minimal-libsdformat-interface-types-for-non-sdformat-models), when it's implemented.

Toward #430


test/integration/includes.cc
```diff
@@ -209,6 +209,19 @@ TEST(IncludesTest, Includes)
...
+  EXPECT_EQ("test_model_with_file", model2->Name());
+  EXPECT_FALSE(model2->Static());
+  EXPECT_EQ(1u, model2->LinkCount());
+  ASSERT_FALSE(nullptr == model2->LinkByIndex(0));
+  ASSERT_FALSE(nullptr == model2->LinkByName("link"));
+  EXPECT_EQ(model2->LinkByName("link")->Name(), model2->LinkByIndex(0)->Name());
+  EXPECT_TRUE(nullptr == model2->LinkByIndex(1));
+  EXPECT_TRUE(model2->LinkNameExists("link"));
+  EXPECT_FALSE(model2->LinkNameExists("coconut"));
+  EXPECT_EQ("1.6", model2->Element()->OriginalVersion());
```
brawner, line 224: Does the version number of the includes.sdf need to be bumped to 1.8?

test/integration/includes.cc
```diff
@@ -209,6 +209,19 @@ TEST(IncludesTest, Includes)
   EXPECT_EQ("", model1->PoseRelativeTo());
   ASSERT_NE(nullptr, model1->Element());
   EXPECT_TRUE(model1->Element()->HasElement("plugin"));
+
+  const sdf::Model *model2 = world->ModelByIndex(2);
+  ASSERT_NE(nullptr, model2);
+  EXPECT_EQ("test_model_with_file", model2->Name());
+  EXPECT_FALSE(model2->Static());
+  EXPECT_EQ(1u, model2->LinkCount());
+  ASSERT_FALSE(nullptr == model2->LinkByIndex(0));
```
brawner, line 218: `ASSERT_NE`? Here and lines 219 and 221

brawner (approved):
Looks good to me with some minor nitpicks.

test/integration/includes.cc
azeey: Fixed in 5b80879

test/integration/includes.cc
azeey: Bumped to 1.8 in 5b80879.

EricCousineau-TRI (approved):
Sweet, looks great!

scpeters (approved):

PR #449, opened 2020-12-23
⬆️  8.9.1 🏁

chapulina:
Preparing for a last major version 8 release before Ignition Blueprint EOLs in 8 days.

Comparison to 8.9.0:

https://github.com/osrf/sdformat/compare/sdformat8_8.9.0...sdf8

There are no PRs open against `sdf8`.


azeey:
I think we need to backport #438 to sdf8 to fix the build failures on macOS actions.

chapulina:
> I think we need to backport #438 to sdf8 to fix the build failures on macOS actions.

Here you go https://github.com/osrf/sdformat/pull/450

scpeters (approved):

chapulina:
I think the Homebrew failure can be ignored. It's just saying that `sdformat8` is deprecated:

```
Warning: Some installed formulae are deprecated or disabled.
You should find replacements for the following formulae:
  sdformat8
```
Which is weird, because it should only be [deprecated in 3 days](https://github.com/osrf/homebrew-simulation/blob/b205f350f1e5de986281ef535a5907f3bc41d4c4/Formula/sdformat8.rb#L16).

scpeters:
> I think the Homebrew failure can be ignored. It's just saying that `sdformat8` is deprecated:
> 
> ```
> Warning: Some installed formulae are deprecated or disabled.
> You should find replacements for the following formulae:
>   sdformat8
> ```
> 
> Which is weird, because it should only be [deprecated in 3 days](https://github.com/osrf/homebrew-simulation/blob/b205f350f1e5de986281ef535a5907f3bc41d4c4/Formula/sdformat8.rb#L16).

I believe you've already seen this, but for broader visibility, the `disable!` line in the homebrew formulae was automatically deprecating it, so I removed that in https://github.com/osrf/homebrew-simulation/pull/1256

PR #452, opened 2020-12-29
Use action-ignition-ci

scpeters:
Revives #266.

Signed-off-by: Steve Peters <scpeters@openrobotics.org>


chapulina (approved):
LGTM. I see the code checker has some errors that aren't causing the build to fail:

```
  [src/ScopedGraph.hh:85]: (error) syntax error
  [src/ScopedGraph.hh:85]: (error) syntax error
  [src/Utils.cc:79]: (style) The function 'isValidFrameReference' is never used.
  [src/Types.cc:83]: (style) The function 'operator<<' is never used.
```

I think that's probably something to be fixed in `code_check.sh`.

We could probably also add Focal CI, that's just using the `focal` branch of the action.

chapulina:
I'm curious about the long-term plan, are we going to keep the action defined here?

https://github.com/osrf/sdformat/blob/master/.github/workflows/linux-ubuntu-bionic.yml

scpeters:
> I'm curious about the long-term plan, are we going to keep the action defined here?
> 
> https://github.com/osrf/sdformat/blob/master/.github/workflows/linux-ubuntu-bionic.yml

I think we can delete that workflow. I'll get back to it for sure next week when I return from vacation; otherwise someone else can take this over

scpeters:
> We could probably also add Focal CI, that's just using the `focal` branch of the action.

:+1:

PR #453, opened 2020-12-30
Add GPS sensor to sdf9

Dotrar:
Here is a basic GPS implementation built on SDF10

this is related to the following issue: https://github.com/ignitionrobotics/ign-sensors/issues/23 as the GPS sensor will need an SDFormat to go with it.

I copied / got inspired by the code in the imu and altimeter sensors, but I only made a single noise component for both position and velocity.

Hope this suits. I couldn't find a way to squash all of my commits into the 1 result. But if this is accepted then it can just be squashed at merge.


chapulina (changes requested):
Thank you for the contribution! This is going in a good direction, but the API needs to be updated to match the XML spec.

Also, please also add a test for the GPS to `test/integration/link_dom.cc`, similar to the IMU tests there. This should cover the code path from XML to C++.

https://github.com/osrf/sdformat/blob/f6e029bc88d2ca6c2586fc88dddc0c21f0798f03/test/integration/link_dom.cc#L388-L445

include/sdf/Gps.hh
```diff
@@ -0,0 +1,132 @@
...
+  /// SDF XML by the "gps" type.
+  ///
+  /// # Example SDF XML using gps type:
+  ///
+  /// ~~~{.xml}
+  /// <sensor name="gps_sensor" type="gps">
+  ///     <pose>1 2 3 0 0 0</pose>
+  ///     <topic>/gps</topic>
+  ///     <gps>
+  ///       <position_sense>
```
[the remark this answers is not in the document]
Dotrar: cheers, I've made the gps implementation follow spec now.

Dotrar:
I seem to have broken the DCO by using the github auto-edit feature, but each of the commits should be signed off.
my git log:
```
commit 26296f04edde7443e24c171c8c02e0f04db70c72 (HEAD -> gps_sensor_sdf10, origin/gps_sensor_sdf10)
Author: Dre Westcook <dre.west@outlook.com>
Date:   Thu Dec 31 10:14:12 2020 +1100

    fixed lints
    
    Signed-off-by: Dre Westcook <dre.west@outlook.com>

commit b5501938eb38e6d392a5dccc17908074d2e55d61
Author: Dre Westcook <dre.west@outlook.com>
Date:   Thu Dec 31 09:23:52 2020 +1100

    changed gps to follow spec
    
    Signed-off-by: Dre Westcook <dre.west@outlook.com>

commit 6599658c77bb818715079d069c16a51346e432af
Author: Dre West <sir.west.dre@gmail.com>
Date:   Thu Dec 31 08:38:53 2020 +1100

    Update include/sdf/Gps.hh
    
    Co-authored-by: Louise Poubel <louise@openrobotics.org>
    Signed-off-by: Dre Westcook <dre.west@outlook.com>

commit 0e54367ab6f874373bac239d3553068f2cac7781
Author: Alejandro Hernández Cordero <alejandro@openrobotics.org>
Date:   Tue Dec 29 00:22:16 2020 +0100

    Added render_order to material (#446)
    
commit 8623a03459644429ebcb45c5df8357073dbd5e08
Author: Dre Westcook <dre.west@outlook.com>
Date:   Tue Dec 29 10:54:46 2020 +1100

    Add GPS into SDFormat.
    
    Signed-off-by: Dre Westcook <dre.west@outlook.com>

commit 7dbf12136f01c70470a41255618dd08937e7bfad (tag: sdformat10_10.1.0)
Author: Nate Koenig <nkoenig@users.noreply.github.com>
Date:   Tue Dec 15 13:59:31 2020 -0800

    Prepare for 10.1.0 (#442)
    
    Co-authored-by: Nate Koenig <nate@openrobotics.org>
```

mjcarroll:
How would you feel about changing this to be something like `NavSat`, as it is more generic.  GPS is the American variant of a Satellite Navigation system, and chips that use multiple variants (Glonass, Galileo, Compass) are becoming widely available

chapulina:
> How would you feel about changing this to be something like NavSat

We could add NavSat and keep supporting GPS for a while as legacy, like we're doing for the ray sensors:

https://github.com/osrf/sdformat/blob/cd2ad01d709e754bfdd63b76eb5ffb714cf679a3/sdf/1.7/sensor.sdf#L33

src/Gps.cc
```diff
@@ -0,0 +1,230 @@
+/*
+ * Copyright 2019 Open Source Robotics Foundation
```
chapulina, line 2:
```suggestion
 * Copyright 2021 Open Source Robotics Foundation
```

src/Gps.cc
```diff
@@ -0,0 +1,230 @@
...
+  // Load gps sensor properties
+  if (_sdf->HasElement("position_sensing"))
+  {
+    sdf::ElementPtr elem = _sdf->GetElement("position_sensing");
+    if (elem->HasElement("horizontal"))
+    {
+      sdf::ElementPtr horiz = elem->GetElement("horizontal");
+      if (horiz->HasElement("noise"))
+      {
+      this->dataPtr->horizontalPositionNoise.Load(horiz->GetElement("noise"));
```
chapulina, line 120:
```suggestion
        this->dataPtr->horizontalPositionNoise.Load(horiz->GetElement("noise"));
```

src/Gps.cc
```diff
@@ -0,0 +1,230 @@
...
+  return * this;
+}
+
+//////////////////////////////////////////////////
+/// \brief Load the gps based on an element pointer. This is *not*
+/// the usual entry point. Typical usage of the SDF DOM is through the Root
+/// object.
+/// \param[in] _sdf The SDF Element pointer
+/// \return Errors, which is a vector of Error objects. Each Error includes
+/// an error code and message. An empty vector indicates no error.
```
chapulina, line 85: No need to repeat the documentation here, just on the header is fine.

chapulina (commented):
I just have minor comments, this is looking good to me.

What do you think of the NavSat idea, @Dotrar ? You'd need to change the new class to `sdf::NavSat`, as well as the enum... As a reference, see how lidar / ray sensors are implemented.

Dotrar:
> I just have minor comments, this is looking good to me.
> 
> What do you think of the NavSat idea, @Dotrar ? You'd need to change the new class to `sdf::NavSat`, as well as the enum... As a reference, see how lidar / ray sensors are implemented.

Yeah no worries; I'll see if i can track down all references and change it to NavSat

should it be as easy as a s/Gps/NavSat/g ? also I'll put in those other changes soon.

chapulina:
> should it be as easy as a s/Gps/NavSat/g ?

Thanks! You may also need some `s/GPS/NAVSAT/g` and `s/gps/navSat/g`. And `git mv Gps NavSat`.

chapulina:
Hi @Dotrar , are you planning to follow up on the NavSat idea? Thanks!

Dotrar:
> Hi @Dotrar , are you planning to follow up on the NavSat idea? Thanks!

yes, Sorry I'm being a little slow here, all sorts of stuff going on; I'll make sure I get this done soon.

Dotrar:
Hey @chapulina ; 

I've changed GPS to SATNAV, but I'm not sure if I went *too far* with changing stuff like the enum in Sensor.hh 

https://github.com/osrf/sdformat/blob/master/include/sdf/Sensor.hh#L72 

do you want to see if I've got this right? I did try to copy the ray/lidar example but I don't know if that includes in the enum, as there isn't a `RAY` in the enum.

also I think I've buggered up the commit log a little when I committed without signing then went back trying to sign previous commits; do you mind if I nuke this PR and write it all into a new one? or is this recoverable/not a huge drama?

include/sdf/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ set (headers
   Filesystem.hh
   Frame.hh
   Geometry.hh
+  SatNav.hh
```
chapulina, line 20: Please move it down for alphabetical order

include/sdf/Sensor.hh
```diff
@@ -65,8 +66,8 @@ namespace sdf
     /// \brief A force-torque sensor.
     FORCE_TORQUE = 5,
 
-    /// \brief A GPS sensor.
-    GPS = 6,
+    /// \brief A SatNav sensor, such as GPS.
+    SATNAV = 6,
```
chapulina, line 70: We should keep GPS here for a while for backward compatibility. Could you add a new enum at the end with the number 21? Keeping GPS and moving this to the end should make sure we don't break ABI.

src/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ set (sources
   FrameSemantics.cc
   Filesystem.cc
   Geometry.cc
+  SatNav.cc
```
chapulina, line 30: Same thing here about alphabetizing

src/SatNav.cc
```diff
@@ -0,0 +1,226 @@
...
+    {
+      sdf::ElementPtr vert = elem->GetElement("vertical");
+      if (vert->HasElement("noise"))
+      {
+        this->dataPtr->verticalVelocityNoise.Load(vert->GetElement("noise"));
+      }
+    }
+  }
+
+
```
chapulina, line 147: Nit: can we leave only 1 blank line? :grimacing:

src/Sensor.cc
```diff
@@ -41,7 +42,7 @@ const std::vector<std::string> sensorTypeStrs =
   "contact",
   "depth_camera",
   "force_torque",
-  "gps",
+  "satnav",
```
chapulina, line 45: Let's keep `gps` for backwards compatibility, the same way as in the enum

chapulina (commented):
> I've changed GPS to SATNAV

Would it be possible to use `NAVSAT` everywhere instead of `SATNAV`?

include/sdf/Sensor.hh
scpeters: agreed; here's a link to the ABI checker report:

* [Build Status](https://build.osrfoundation.org/job/sdformat-abichecker-any_to_any-ubuntu_auto-amd64/1093/) https://build.osrfoundation.org/job/sdformat-abichecker-any_to_any-ubuntu_auto-amd64/1093/API_5fABI_20report/

src/Sensor.cc
adlarkin: https://github.com/osrf/sdformat/pull/453/commits/b128c9cc8c5aad02e8db480bd25626ae8769c545

src/SatNav.cc
adlarkin: https://github.com/osrf/sdformat/pull/453/commits/b128c9cc8c5aad02e8db480bd25626ae8769c545

src/CMakeLists.txt
adlarkin: https://github.com/osrf/sdformat/pull/453/commits/b128c9cc8c5aad02e8db480bd25626ae8769c545

include/sdf/Sensor.hh
adlarkin: Should be fixed now in https://github.com/osrf/sdformat/pull/453/commits/b128c9cc8c5aad02e8db480bd25626ae8769c545. Let's see what the ABI checker says :crossed_fingers:

include/sdf/CMakeLists.txt
adlarkin: https://github.com/osrf/sdformat/pull/453/commits/b128c9cc8c5aad02e8db480bd25626ae8769c545

adlarkin:
@chapulina and @scpeters: I have addressed the latest review comments, and believe that this is ready for final review!

> Would it be possible to use `NAVSAT` everywhere instead of `SATNAV`?

Done in https://github.com/osrf/sdformat/pull/453/commits/b128c9cc8c5aad02e8db480bd25626ae8769c545. Some files had to be re-named, so apologies if it's a bit of a pain to review these changes :grimacing: 

> I think we should add a `navsat.sdf` file too, which is a copy of `gps.sdf`. Like we have both `ray.sdf` and `lidar.sdf`.

I added this in https://github.com/osrf/sdformat/pull/453/commits/b128c9cc8c5aad02e8db480bd25626ae8769c545. I also updated `sensors.sdf` to note that `gps` is now deprecated in favor of `navsat`. If there are any other documentation files that need to be updated, let me know!

---

Regarding DCO: once final reviews are done and this PR has been approved, I can try to fix the Git history so that DCO will pass.

azeey:
Can I ask why Navsat is preferred over Satnav? The latter seems more generic to me.

adlarkin:
> Can I ask why Navsat is preferred over Satnav? The latter seems more generic to me.

I think the original motivation came from https://github.com/osrf/sdformat/pull/453#issuecomment-755327583. Maybe @mjcarroll or @chapulina can give their thoughts?

src/Sensor_TEST.cc
```diff
@@ -247,7 +247,7 @@ TEST(DOMSensor, Type)
     sdf::SensorType::CONTACT,
     sdf::SensorType::DEPTH_CAMERA,
     sdf::SensorType::FORCE_TORQUE,
-    sdf::SensorType::GPS,
+    sdf::SensorType::NAVSAT,
```
adlarkin, line 250: Should we leave `sdf::SensorType::GPS` in this test since we support both options?

chapulina:
> Can I ask why Navsat is preferred over Satnav? The latter seems more generic to me.

Good question, [Satellite Navigation](https://en.wikipedia.org/wiki/Satellite_navigation) refers to the navigation system, while to me "Navigation Satellite" seems to refer to the specific device, i.e. sensor, being used.

But what motivated my original request was the equivalent ROS message [sensor_msgs/NavSatFix](https://github.com/ros2/common_interfaces/blob/foxy/sensor_msgs/msg/NavSatFix.msg). ROS messages are widely used and usually receive a lot of scrutiny (like the recent [API review](https://discourse.ros.org/t/foxy-message-api-review/13105)) so I think it's usually a good idea to follow their patterns unless we have a good reason not to.

That said, I don't feel too strongly about this after giving it more thought. If there's enough interest in keeping SatNav I'm ok with that.

include/sdf/Sensor.hh
adlarkin: ABI checker is green, so I'm marking this as resolved

azeey:
Ah, I didn't know about `sensor_msgs/NavSatFix`. It makes sense to use `NavSat` then since there's a precedent.

mjcarroll:
> That said, I don't feel too strongly about this after giving it more thought. If there's enough interest in keeping SatNav I'm ok with that.

If it helps at all, NAVSAT was the precursor to GPS https://en.wikipedia.org/wiki/Transit_(satellite)

src/Sensor_TEST.cc
azeey: Since the enum set by `sdf::Sensor::Load` will be `NAVSAT` for either `gps` or `navsat` strings, I don't think we can test this.

azeey (approved):
The code looks good. I think it can go into `sdf9` though.

chapulina (approved):
Rebased to `sdf9`, LGTM!

PR #454, opened 2021-01-06
Add laser_retro in Visual

doisyg:
First PR needed for laser_retro support as discussed here: https://github.com/ignitionrobotics/ign-rendering/issues/181


include/sdf/Collision.hh
```diff
@@ -186,6 +186,24 @@ namespace sdf
...
+    /// \return True if the lidar reflective intensity was set was set.
+    public: bool HasLaserRetro() const;
+
+    /// \brief Get the flidar reflective intensity.
+    /// \return The lidar reflective intensity.
+    public: double LaserRetro() const;
+
+    /// \brief Set the lidar reflective intensity.
+    /// \param[in] _laserRetro The lidar reflective intensity.
+    public: void SetLaserRetro(double _laserRetro);
```
chapulina, line 206: Mind moving the new public functions up with the existing ones?

include/sdf/Collision.hh
```diff
@@ -186,6 +186,24 @@ namespace sdf
 
     /// \brief Private data pointer.
     private: CollisionPrivate *dataPtr = nullptr;
+
+    /// \brief Set whether the lidar reflective intensity
+    /// has been specified.
+    /// \param[in] _laserRetro True if the lidar reflective intensity
+    /// has been set in the sdf.
+    public: void SetHasLaserRetro(bool _laserRetro);
```
chapulina, line 194: We haven't been consistently adding `Has` functions for optional elements. I only see this in `Camera.hh`. I'm not sure if that's the direction we want to go in, or if we want to just leave the default value in case it's unset.

Thoughts, @azeey ?

src/Visual.cc
```diff
@@ -193,6 +201,13 @@ Errors Visual::Load(ElementPtr _sdf)
   Errors geomErr = this->dataPtr->geom.Load(_sdf->GetElement("geometry"));
   errors.insert(errors.end(), geomErr.begin(), geomErr.end());
 
+  // Load the lidar reflective intensity if it is given
+  if (_sdf->HasElement("laser_retro"))
+  {
+    this->SetHasLaserRetro(true);
```
chapulina, line 207: This is being set inside `SetLasetRetro` too, do we need it in both places?

chapulina (commented):
Thanks for the PR. Mind adding some tests for the collision too? And also signing the commits for DCO?

src/Visual.cc
doisyg: You're right

doisyg:
Should be signed now. Removed all Collision modifications (the .hh changes were residual) as I want to implement laser_retro only in Visual for now. More simple as there is no equivalent to https://github.com/osrf/sdformat/blob/32d96a6b3587d518695253fb4d4c70b8404525b0/src/Visual.cc#L75-L83 in Collision yet

chapulina:
The diff looks weird, as if this PR were adding the code for #388, even though that's already merged into `sdf9` :thinking: Git does these weird things sometimes. I wonder if a rebase removing commit 2ae6fc758890ce2708aaaf54157f28d0b080cd49 would fix it :thinking:

doisyg:
> The diff looks weird, as if this PR were adding the code for #388, even though that's already merged into `sdf9` thinking Git does these weird things sometimes. I wonder if a rebase removing commit [2ae6fc7](https://github.com/osrf/sdformat/commit/2ae6fc758890ce2708aaaf54157f28d0b080cd49) would fix it thinking

I merged the last sdf9 branch into this PR branch, this seems to have fix the diff for now. I ll do a rebase or reset if that's not enough

chapulina (approved):
Looking good. I still have the request to move the public functions above the private ones before merging. Thanks!

PR #455, opened 2021-01-06
Return to previous build directory after testing examples

azeey:
I believe coverage generation is failing right now because the example test changes the directory but doesn't return to the original build directory  (eg. https://github.com/osrf/sdformat/pull/444/checks?check_run_id=1647029579#step:4:4010)

### DIFF
--- a/.github/ci/after_make_test.sh
+++ b/.github/ci/after_make_test.sh
@@ -6,9 +6,11 @@ set -x
 make install
 
 # Compile examples
+curdir=$PWD
 cd ../examples
 mkdir build;
 cd build;
 cmake ..;
 make;
 ./simple ../simple.sdf;
+cd $curdir


chapulina (approved):

azeey:
Actions are failing because of an issue with gzdev. https://github.com/ignition-tooling/gzdev/pull/28

.github/ci/after_make_test.sh
```diff
@@ -6,9 +6,11 @@ set -x
...
 
 # Compile examples
+curdir=$PWD
 cd ../examples
 mkdir build;
 cd build;
 cmake ..;
 make;
 ./simple ../simple.sdf;
+cd $curdir
```
scpeters, line 16: nit: alternatively we could `pushd` then `popd`

azeey: Yeah, I considered that, but thought this was simpler because there would have been two `pushd` and `popd`s.

scpeters: I think if you use `pushd` first and `popd` last, then you can use `cd` in the middle

azeey: Ah, you're right. But I tried it locally and it looks like `pushd` and `popd` are not available in `dash`, the default `sh` binary in Ubuntu. Seeing in https://github.com/ignition-tooling/action-ignition-ci/blob/master/entrypoint.sh#L168 that this script is sourced from a shell script run by `sh` and we don't install `bash` in that docker image, I'm guessing `sh` would still be symlinked to `dash` and not `bash`.

scpeters: ok, `pushd` is too complicated; let's ignore my nit

scpeters (approved):

PR #456, opened 2021-01-07
Disable ign tests on Windows (sdf10)

j-rivero:
Following @scpeters suggestion https://github.com/ignition-tooling/release-tools/pull/365#issuecomment-748378863, the PR disable the run of ign tests on Windows.

### DIFF
--- a/cmake/SearchForStuff.cmake
+++ b/cmake/SearchForStuff.cmake
@@ -101,6 +101,13 @@ macro (check_gcc_visibility)
 endmacro()
 
 ########################################
+# Find ignition cmake2
+# Only for using the testing macros, not really
+# being use to configure the whole project
+find_package(ignition-cmake2 2.3 REQUIRED)
+set(IGN_CMAKE_VER ${ignition-cmake2_VERSION_MAJOR})
+
+########################################
 # Find ignition math
 # Set a variable for generating ProjectConfig.cmake
 find_package(ignition-math6 QUIET)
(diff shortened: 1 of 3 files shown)


cmake/SearchForStuff.cmake
```diff
@@ -111,3 +111,10 @@ else()
   set(IGN_MATH_VER ${ignition-math6_VERSION_MAJOR})
   message(STATUS "Looking for ignition-math${IGN_MATH_VER}-config.cmake - found")
 endif()
+
+########################################
+# Find ignition cmake2
+# Only for using the testing macros, not really
+# being use to configure the whole project
+find_package(ignition-cmake2 2.3 REQUIRED)
```
chapulina, line 119: How about checking `if (IGNITION-TOOLS_BINARY_DIRS)`?

scpeters, line 119: does this need to be called before finding `ignition-math6`? it's a dependency of `ignition-math6`, so I'm not sure if an old version would be improperly cached

scpeters: I don't think that is useful because we often install `ign-tools` in windows workspaces. I think we want the macros available in all cases, so I don't think we should conditionally find this

src/CMakeLists.txt
```diff
@@ -134,6 +134,14 @@ if (BUILD_SDF_TEST)
 
   sdf_build_tests(${gtest_sources})
 
+  if (IGNITION-TOOLS_BINARY_DIRS)
```
scpeters, line 137: `if (TARGET UNIT_ign_TEST)`

scpeters (approved):

PR #457, opened 2021-01-09
Add split/join for separating the link name from it's fully qualified name

brawner:
This PR adds small functionality for splitting and joining a name based on the '::' separator. It's necessary for https://github.com/ignitionrobotics/ign-gazebo/pull/542

### DIFF
--- a/include/sdf/Types.hh
+++ b/include/sdf/Types.hh
@@ -60,6 +60,8 @@ namespace sdf
   inline namespace SDF_VERSION_NAMESPACE {
   //
 
+  const std::string kSdfScopeDelimiter = "::";
+
   /// \brief Split a string using the delimiter in splitter.
   /// \param[in] str       The string to split.
   /// \param[in] splitter  The delimiter to use.
@@ -228,6 +230,14 @@ namespace sdf
   /// \param[in] _in String to convert to lowercase
   /// \return Lowercase equilvalent of _in.
   std::string SDFORMAT_VISIBLE lowercase(const std::string &_in);
+
+  SDFORMAT_VISIBLE
+  std::pair<std::string, std::string> SplitName(
+      const std::string &_absoluteName);
+
+  SDFORMAT_VISIBLE
+  std::string JoinName(
+      const std::string &_scopeName, const std::string &_localName);
   }
 }
 #endif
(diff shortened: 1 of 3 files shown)


include/sdf/Types.hh
```diff
@@ -228,6 +231,14 @@ namespace sdf
   /// \param[in] _in String to convert to lowercase
   /// \return Lowercase equilvalent of _in.
   std::string SDFORMAT_VISIBLE lowercase(const std::string &_in);
+
+  SDFORMAT_VISIBLE
```
azeey, line 235: Can you add docs here?

src/Types.cc
```diff
@@ -88,5 +88,59 @@ std::ostream &operator<<(std::ostream &_out, const sdf::Errors &_errs)
...
+  return {"", _absoluteName};
+}
+
+static bool EndsWithDelimiter(const std::string &_s)
+{
+  if (_s.size() < kSdfScopeDelimiter.size())
+    return false;
+
+  const size_t findStartPosition = _s.size() - kSdfScopeDelimiter.size();
+  return _s.substr(findStartPosition) == kSdfScopeDelimiter;
```
azeey, line 115: nit: Probably more efficient to do `_s.find(kSdfScopeDelimiter, findStartPosition) == findStartPosition` as it avoids constructing a new string.

brawner, line 115: This should actually just use std::string::compare(). Same in line 123

azeey (approved):
LGTM!

src/Types.cc
brawner: Done

include/sdf/Types.hh
brawner: Added

src/Types.cc
brawner: Totally, switched to std::string::compare

src/Types.cc
```diff
@@ -88,5 +88,60 @@ std::ostream &operator<<(std::ostream &_out, const sdf::Errors &_errs)
...
+
+static bool StartsWithDelimiter(const std::string &_s)
+{
+  if (_s.size() < kSdfScopeDelimiter.size())
+    return false;
+
+  return _s.compare(0, kSdfScopeDelimiter.size(), kSdfScopeDelimiter) == 0;
+}
+
+// Join an scope name prefix with a local name using the scope delimeter
```
scpeters, line 127: nit: `an` -> `a`

scpeters (approved):
looks great! just one nit that I'll fix right now

src/Types.cc
scpeters: 2db1d1c

brawner:
Thanks Steve for fixing the nit!

PR #459, opened 2021-01-12
Reduce default heightmap sampling from 2 to 1 

chapulina:
There's no good reason to subsample by default, see discussion in https://github.com/ignitionrobotics/ign-rendering/pull/180#discussion_r551661502.

This also needs to be changed in `Heightmap.hh` once that's merged forward, so I'll keep this PR as a draft until:

- [x] #458 is merged
- [x] then a 10 ➡️ 11 PR is merged
- [x] this PR is updated to also change the default sampling on the C++ code


chapulina:
This PR is ready for review

iche033:
changes look good. Just need to update the heightmap test expectations to fix the [failing tests](https://build.osrfoundation.org/job/sdformat-ci-pr_any-ubuntu_auto-amd64/1452/testReport/)

chapulina:
@scpeters , I'd like to make sure you agree with this change from the physics point of view

iche033 (approved):

scpeters (approved):

PR #432, opened 2020-12-09
10 ➡️ master

azeey:
I'd appreciate some scrutiny on this PR since this is my first time doing a cherry-pick forward port on sdformat. My process was:
* Create a branch from master called `10_master_reference`
* Merge `sdf10` into `10_master_reference` and fix any conflicts
* Use `git cherry -v master > cherry_list.txt` and manually go through the generated text and check if those commits have been added to `master`. If so, remove them from the list.
* Create another branch on master called `10_to_master`
* `cat cherry_list.txt | cut -d' ' -f2 | xargs git cherry-pick` and fix any conflicts
* Compare `10_to_master` and `10_master_reference` and verify that the differences left over do not need to be ported.

PR #462, opened 2021-01-13
10 ➡️  master

azeey:
Merge-forward from sdf10 to master. This replaces #432


azeey:
https://github.com/osrf/sdformat/pull/434 updated `.github/workflows/linux-ubuntu-bionic.yml` to use prereleases, but that workflow file was deleted in https://github.com/osrf/sdformat/pull/452. Do we have a way to use prereleases in github actions with our new workflow file?

azeey:
> Do we have a way to use prereleases in github actions with our new workflow file?

Looks like it uses gzdev which has been updated to use prereleases in https://github.com/ignition-tooling/gzdev/pull/26. I just needed to remove `.github/workflows/linux-ubuntu-bionic.yml`.

chapulina (approved):
Thanks for opening the PR, @azeey ! I went over all files and don't see anything out of place. CI seems happy too.

j-rivero (approved):
Compiles fine for me on Windows and fixed problems with lack of sky support on master.

PR #465, opened 2021-01-15
Changelog: remove duplicate entries from 10.2.0

scpeters:
Some duplicate entries from 9.3.0 were added to the 10.2.0 changeling, so this removes them. Also fix reference to `libsdformat` in some headings.


chapulina (approved):
Thanks for the cleanup! :pray:

Issue #467, opened 2021-01-16
Add Ignition Utils as dependency from libSDFormat 11


chapulina:
The new Ignition package has minimal dependencies and provides utility functions that could be useful to SDFormat, like the [ImplPtr](https://github.com/ignitionrobotics/ign-utils/blob/main/include/ignition/utils/ImplPtr.hh):

https://github.com/ignitionrobotics/ign-utils/

The dependency needs to be added on CI for all 3 platforms before it can be used here.

scpeters:
done in https://github.com/osrf/sdformat/pull/474

PR #472, opened 2021-01-20
Change all std::unique_ptr members of *Private classes to std::optional

azeey:
This avoids having to create copy constructors for those `*Private` classes and makes way for using `ignition::utils::ImplPtr`

Toward #467

### DIFF
--- a/src/Geometry.cc
+++ b/src/Geometry.cc
@@ -14,6 +14,9 @@
  * limitations under the License.
  *
 */
+
+#include <optional>
+
 #include "sdf/Geometry.hh"
 #include "sdf/Box.hh"
 #include "sdf/Capsule.hh"
@@ -24,6 +27,8 @@
 #include "sdf/Plane.hh"
 #include "sdf/Sphere.hh"
 
+#include "Utils.hh"
+
 using namespace sdf;
 
 // Private data class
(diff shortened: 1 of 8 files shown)


src/Geometry.cc
```diff
@@ -75,57 +80,8 @@ Geometry::~Geometry()
 
 //////////////////////////////////////////////////
 Geometry::Geometry(const Geometry &_geometry)
-  : dataPtr(new GeometryPrivate)
+  : dataPtr(new GeometryPrivate(*_geometry.dataPtr))
```
brawner, line 83: Out of curiosity, why aren't all these private pointers `std::unique_ptr`? It would let you have have implicit destruction, which can be more efficient.

src/Geometry.cc
```diff
@@ -248,97 +204,97 @@ void Geometry::SetType(const GeometryType _type)
 /////////////////////////////////////////////////
 const Box *Geometry::BoxShape() const
```
brawner, line 205: Could these getters now return a `const std::optional<T> &` or would that be too breaking of a change? It would probably save double checking if the optional type is valid. (once here in optionalToPointer and again at the calling site). It would just be pushed to the caller to check that the optional is valid.

src/Utils.hh
```diff
@@ -181,6 +182,14 @@ namespace sdf
 
     return errors;
   }
+
+  template <typename T>
+  T *optionalToPointer(std::optional<T> &_shape)
```
brawner, line 187: Should there also be a const overload? `const T *optionalToPointer(const std::optional<T> &_shape)`

src/Geometry.cc
azeey: We are soon changing them all to `ignition::utils::ImplPtr` as started in #447. However, to answer your question, I think old code used raw pointers and was never updated. There's also a pesky warning when you use `std::unique_ptr` in an exported class in Windows.

src/Utils.hh
azeey: I actually started with the const version, but there's one class:https://github.com/osrf/sdformat/blob/3ec5f5ae1e2cc3ad59c6f17f5003a6d6737bde1d/src/World.cc#L402 that needs a non-const pointer (not sure if that was intentional). I'll add the overload.

src/Geometry.cc
mjcarroll: Even in this case, the caller has to check that the returned pointer is valid, so I don't think it's too much of a stretch

src/Utils.hh
mjcarroll: Agree, also not to get super picky, but it's not necessarily a "shape" in this call, so maybe the arg could be renamed.

src/Geometry.cc
azeey: I would love to do that, but I think that would break existing code without a tick-tock.

src/Utils.hh
azeey: Added the overload and changed `shape` to `opt` in https://github.com/osrf/sdformat/pull/472/commits/23721255163308221625f95ec830c6508adc684a

src/Geometry.cc
brawner: If that's a desirable change, then it seems like it makes sense to start the tick-tock cycle. But how would that look if you're just changing the return type? Would you need to create new method name?

mjcarroll:

mjcarroll (approved):

PR #474, opened 2021-01-22
Update classes to use ignition::utils::ImplPtr or ignition::utils::UniqueImplPtr 

azeey:
This builds on top of the commits in #447 and updates almost all the classes in libsdformat to use either `ignition::utils::ImplPtr` or `ignition::utils::UniqueImplPtr`. While making this update, a number of member functions had to be updated for const correctness.

Needs: https://github.com/ignitionrobotics/ign-math/pull/187


brawner (approved):
Seems good to me. I don't see anything that sticks out, but it is a bit long to review so it's hard review every single addition.

Like, I'm not sure why sometimes you choose the unique variant or the other.

That is cool though that this addition caught the issue of const-correctness in some cases!

azeey:
> Like, I'm not sure why sometimes you choose the unique variant or the other.

I used the unique variant for `sdf::filesystem::DirIter` and `sdf::Root` because those classes were explicitly non-copiable.

chapulina:
This is needed, otherwise the nightlies will be broken: https://github.com/ignition-release/sdformat11-release/pull/2

scpeters (commented):
wow, great work! it took a while to review but only because there were so many files. it was easy enough to click through them quickly because it's mostly deleted code

I think the only thing it's missing is an entry in the Migration guide mentioning the new dependency on ignition-utils1

scpeters:
> I think the only thing it's missing is an entry in the Migration guide mentioning the new dependency on ignition-utils1

41becc4de1530f785836b7d52d05227a75882a9b

scpeters (approved):

chapulina:
I didn't expect the const correctness to be too intrusive, but it looks like it broke `ign-gazebo`. Something to keep in mind, we may want to revisit this before the stable release and add deprecated non-const overloads to tick-tock the change.

PR #477, opened 2021-01-28
Escape ampersand

### LINKED ISSUE #375: Ampersands not showing in generated SDF collide bitmask doc
The documentation source says
```Collision happens if ((category1 & collision2) | (category2 & collision1)) is not zero.```

Unfortunately the ampersands get lost in generating the html, and the website shows this, which is quite confusing

2020-09-18-003449_347x82_scrot

claireyywang:
closes #375

### DIFF
--- a/sdf/1.6/surface.sdf
+++ b/sdf/1.6/surface.sdf
@@ -145,7 +145,7 @@
     </element>
 
     <element name="category_bitmask" type="unsigned int" default="65535" required="0">
-      <description><![CDATA[Bitmask for category of collision filtering. Collision happens if ((category1 & collision2) | (category2 & collision1)) is not zero. If not specified, the category_bitmask should be interpreted as being the same as collide_bitmask.]]></description>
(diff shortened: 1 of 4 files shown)


EricCousineau-TRI (approved):

azeey (commented):
Technically, since the `&` is inside `CDATA`, it is legal xml. I believe the problem is with the code that processes all of these and generates the `full_*.sdf` files. That code seems to strip the `CDATA` and leave the ampersands unescaped.

What if we wrap all `<description>` tags with `CDATA` in
https://github.com/osrf/sdformat/blob/master/src/Element.cc#L256?

For context, the tool that generates the `full_*.sdf` files is `ign sdf -d` invoked from cmake at https://github.com/osrf/sdformat/blob/c4a029c862329a3be9459e0844fcdde2539b2709/sdf/CMakeLists.txt#L37

That in turn calls `sdf::SDF::PrintDescription` https://github.com/osrf/sdformat/blob/c4a029c862329a3be9459e0844fcdde2539b2709/src/ign.cc#L127

claireyywang:
closing since https://github.com/osrf/sdformat/pull/483 should fix the issue

PR #482, opened 2021-01-30
Fix temperature gradient default value in Atmosphere DOM

iche033:
The default value in the Atmosphere SDF DOM was off by a factor of 10, see [default value in atmosphere.sdf spec](https://github.com/osrf/sdformat/blob/master/sdf/1.7/atmosphere.sdf#L17). I believe the one in the spec is correct according to [wikipedia](https://en.wikipedia.org/wiki/Atmospheric_temperature): `Temperatures in the atmosphere decrease with height at an average rate of 6.5°C/km.` 

This could be a behavior change so I could retarget to `master` if needed

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -16,7 +16,7 @@ but with improved human-readability..
 
 ### Modifications
 
-1. Fixed Atmosphere DOM class's temperature default value. Changed it from from -0.065 to -0.0065.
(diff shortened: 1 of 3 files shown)


scpeters:
I fixed the unit test in 8aa6e13

scpeters (approved):
I would characterize this as a bug, since it deviates from the spec. We can note it in the migration guide

iche033:
> We can note it in the migration guide

yep done.

PR #487, opened 2021-02-05
Add L16 pixel format to Camera pixel format conversion function

iche033:
So we can convert `<camera><image><format>` string to `PixelFormatType` when parsing <camera> sdf.

### DIFF
--- a/sdf/1.7/camera.sdf
+++ b/sdf/1.7/camera.sdf
@@ -18,7 +18,7 @@
       <description>Height in pixels </description>
     </element>
     <element name="format" type="string" default="R8G8B8" required="0">
-      <description>(L8|R8G8B8|B8G8R8|BAYER_RGGB8|BAYER_BGGR8|BAYER_GBRG8|BAYER_GRBG8)</description>
+      <description>(L8|L16|FLOAT16|FLOAT32|R8G8B8|B8G8R8|BAYER_RGGB8|BAYER_BGGR8|BAYER_GBRG8|BAYER_GRBG8)</description>
     </element>
   </element> <!-- End Image -->
 
--- a/sdf/1.7/camera.sdf
+++ b/sdf/1.7/camera.sdf
@@ -18,7 +18,7 @@
       <description>Height in pixels </description>
     </element>
     <element name="format" type="string" default="R8G8B8" required="0">
-      <description>(L8|L16|FLOAT16|FLOAT32|R8G8B8|B8G8R8|BAYER_RGGB8|BAYER_BGGR8|BAYER_GBRG8|BAYER_GRBG8)</description>
+      <description>(L8|L16|R_FLOAT16|R_FLOAT32|R8G8B8|B8G8R8|BAYER_RGGB8|BAYER_BGGR8|BAYER_GBRG8|BAYER_GRBG8)</description>
     </element>
   </element> <!-- End Image -->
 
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

adlarkin:
Do we need to update the SDFormat specification to include this addition? Looking at the [website](http://sdformat.org/spec?ver=1.7&elem=sensor#image_format), it shows that the options for `<camera><image><format>` are `L8|R8G8B8|B8G8R8|BAYER_RGGB8|BAYER_BGGR8|BAYER_GBRG8|BAYER_GRBG8` (I don't see `L16` as an option).

scpeters:
> Do we need to update the SDFormat specification to include this addition? Looking at the [website](http://sdformat.org/spec?ver=1.7&elem=sensor#image_format), it shows that the options for `<camera><image><format>` are `L8|R8G8B8|B8G8R8|BAYER_RGGB8|BAYER_BGGR8|BAYER_GBRG8|BAYER_GRBG8` (I don't see `L16` as an option).

that documentation comes from: https://github.com/osrf/sdformat/blob/sdformat10_10.2.0/sdf/1.7/camera.sdf#L21

scpeters:
here's the full list in the enum for reference: https://github.com/osrf/sdformat/blob/sdformat10_10.2.0/include/sdf/Camera.hh#L39-L60

should we add `FLOAT16` and `FLOAT32`?

adlarkin:
> should we add FLOAT16 and FLOAT32?

I think these would be good to add, unless anyone else has a reason against it (I'm not too familiar with the `sdformat` repository, so perhaps there are reasons for not adding these that I am not aware of).

chapulina (approved):

adlarkin:
> should we add `FLOAT16` and `FLOAT32`?

@scpeters I went ahead and added `L8`, `FLOAT16` and `FLOAT32` as camera formats to the documentation: https://github.com/osrf/sdformat/pull/487/commits/067cf73758b373f109c6ff435c884f34b5109a14

nkoenig (approved):

scpeters:
> > should we add `FLOAT16` and `FLOAT32`?
> 
> @scpeters I went ahead and added `L8`, `FLOAT16` and `FLOAT32` as camera formats to the documentation: [067cf73](https://github.com/osrf/sdformat/commit/067cf73758b373f109c6ff435c884f34b5109a14)

thanks for making the change for `L8`, but now that I look more closely, I don't think we should document `FLOAT16` or `FLOAT32` unless we add support for them to the conversion functions. it seems that `R_FLOAT16` and `R_FLOAT32` are more precise and used in ign-msgs, so I retract my suggestion that we support `FLOAT16` and `FLOAT32`. I'll go ahead and revert those from your change; thanks for your patience

scpeters (approved):

chapulina (approved):

peci1:
@adlarkin The online docs at http://sdformat.org/spec?ver=1.7&elem=sensor#image_format still aren't updated. Is some manual intervention needed to update them?

chapulina:
> Is some manual intervention needed to update them?

Yup, I think @azeey has been triggering those.

azeey:
> @adlarkin The online docs at http://sdformat.org/spec?ver=1.7&elem=sensor#image_format still aren't updated. Is some manual intervention needed to update them?

The online docs have been updated.

PR #497, opened 2021-02-18
Parse rpyOffset as radians

iche033:
During fixed joint reduction, the `xyzOffset` and `rpyOffset` plugin parameters are modified. It was found that the `rpyOffset` is actually parsed as degrees instead of radians.

This created a problem when you have chained fixed joints - in the first joint reduction, the code writes out the new `rpyOffset` vector in radians, but in the second fixed joint reduction, the code parses it back as degrees.

Looking at [plugins in gazebo_ros_pkgs](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_imu.cpp#L116), this parameter should be in radians

Signed-off-by: Ian Chen <ichen@osrfoundation.org>

PR #500, opened 2021-02-23
Fix xyz and rpy offsets in fixed joint reduction

iche033:
Continuation of the fix from pull request #497. This is another report and contribution from external user.  

The `xyzOffset` and `rpyOffset` computation seems to be incorrect for a model plugin during fixed joint reduction.

This diagram describes the urdf configuration that is being tested:

fixed_joint_offset_tf

The model has three links connected together by two fixed joints with [0 1 0] pos and [0 0 pi/4] rotation joint origin as shown in the diagram (see `test/integration/fixed_joint_reduction_plugin_frame_extension.urdf`). The model plugin has these relevant params:

```
    <plugin name='test_plugin' filename='libtest_plugin.so'>
      ...
      <bodyName>link2</bodyName>
      <xyzOffset>0 0 0</xyzOffset>
      <rpyOffset>0 0 0</rpyOffset>
    </plugin>
```

After fixed joint reduction, `link1` and `link2` are merged into `base_link` and hence the `xyzOffset` and `rpyOffset` values are also modified to be in the `base_link` frame

 Before the changes: the urdf parser produces these offsets:
* `bodyName`: "base_link"
* `xyzOffset`:  -0.707108, **-1.70711**, 0
* `rpyOffset`: 0, 0, -**1.5708**

Looking at the offsets and also the diagram above, these values do not look correct.

 After the changes: the urdf parser produces these offsets:
* `bodyName`: "base_link"
* `xyzOffset`:  -0.707108, **1.70711**, 0
* `rpyOffset`: 0, 0, **1.5708**

These values match what's shown in the diagram.

Note the code has been around for many years so I appreciate a closer look at this. Looking at gazebo_ros_pkgs, the one plugin that could be affected is the [gazebo_ros_p3d](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/melodic-devel/gazebo_plugins/src/gazebo_ros_p3d.cpp) plugin, which is a model plugin and makes use of these offsets. The issue only happens if the user has this plugin and references a frame that's being reduced due to fixed joints. Here's a related [issue](https://github.com/ros-simulation/gazebo_ros_pkgs/issues/843) in gazebo_ros_pkgs. The comments suggest that people have worked around it by disabling fixed joint lumping or using continuous joints with zero limits.


azeey:
The fact that it generates

xyzOffset: -0.707108, **-1.70711**, 0
rpyOffset: 0, 0, -1.5708

before the updates in this PR confuse me. If it was doing the inverse (i.e, base_link relative to link2), wouldn't it be
xyzOffset: **-1.70711**, **-0.707108**, 0
rpyOffset: 0, 0, -1.5708


Aside from that, the changes makes sense to me, but like you said, this is problematic for gazebo_ros_p3d. Particularly, [gazebo_ros_p3d.cpp#L295](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/b0e3d71938e4cd4c80a2ec3ff7187aac49a9f62a/gazebo_plugins/src/gazebo_ros_p3d.cpp#L295) has
```
        pose.Rot() = this->offset_.Rot()*pose.Rot();
```
If we merge this PR, we need to update that line to

```
        pose.Rot() = pose.Rot() * this->offset_.Rot();
```

iche033:
The `inverseTransformToParentFrame` function could be doing the wrong? 

If I set the `bodyName` to `link1` (instead of link2`), I get an `xyzOffset` of `0 -1 0 0 0 -0.7854` but I would expect it to produce something like `-0.707108 -0.707108 0` (`base_link` relative to `link1`)

azeey:
I worked through the math in `inverseTransformToParentFrame` and it doesn't quite come out to the inverse of `TransformToParentFrame`.

The code currently does:
R_BL^-1 p_LQ - p_BL

but for it to be the inverse of `TransformToParentFrame`, I think it should be:
R_BL^-1 p_LQ - R_BL^-1 p_BL

where:
R_BL : `_parentToLinkTransform.rotation`
p_LQ : `_transformInLinkFrame.Pos()`
p_BL:  `_parentToLinkTransform.position`

Also, I noticed that with the changes in this PR, `inverseTransformToParentFrame` is not used anymore.

iche033:
thanks for checking the math. I've removed the `inverseTransformToParentFrame` function since that it's incorrect and no longer used. de3e268.

iche033:
here is the gazebo_ros_pkgs PR for the updating the rotation offset calculation: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1241

azeey (approved):

chapulina:
What's the status of this PR? It's been approved for a while.

scpeters:
fix for Ubuntu CI in https://github.com/ignitionrobotics/sdformat/pull/626

scpeters (approved):

scpeters:
I wonder if we should try to inject a `<ignition:fixed_offsets>true</ignition:fixed_offsets>` tag into the plugin block to simplify the logic in gazebo_plugins required to handle this?

iche033:
> inject a <ignition:corrected_offsets>true</ignition:corrected_offsets> tag into the plugin block

alright I implemented this suggestion in 536b8472a0edb363ee02280ec875db61bb680c91. One question, do we need the `ignition` namespace?

scpeters:
> > inject a ignition:corrected_offsetstrue</ignition:corrected_offsets> tag into the plugin block
> 
> alright I implemented this suggestion in [536b847](https://github.com/ignitionrobotics/sdformat/commit/536b8472a0edb363ee02280ec875db61bb680c91). One question, do we need the `ignition` namespace?

well, since it's in a plugin none of the tags are part of the spec, so a namespace wouldn't be required for that reason, but since we are injecting the element, I feel better about using a namespace so we don't conflict with any preexisting user tags. I think it's unlikely, so we could probably be fine without the namespace, but I have a slight preference for it. If others feel more strongly about not using `ignition:`, I'm happy to go along with removing the namespace

iche033:
ok sounds good! I'll keep the ignition namespace

scpeters:
I just noticed that it seems to be inserting two `<ignition::corrected_offsets>1</ignition::corrected_offsets>` tags

src/parser_urdf.cc
```diff
@@ -3473,14 +3444,21 @@ void ReduceSDFExtensionPluginFrameReplace(
           // remove xyzOffset and rpyOffset
           (*_blobIt)->RemoveChild(rpyKey);
         }
+        TiXmlNode* correctedOffsetKey =
+            (*_blobIt)->FirstChild("ignition::corrected_offsets");
+        if (!correctedOffsetKey)
```
scpeters, line 3449: I think the logic is backwards here. Should the `!` be removed?

scpeters: removed in 2eba874

scpeters:
> I just noticed that it seems to be inserting two `<ignition::corrected_offsets>1</ignition::corrected_offsets>` tags

should be fixed by 2eba874

iche033:
> should be fixed by 2eba874

thanks!

osrf-triage:

osrf-triage:

PR #509, opened 2021-03-11
Store include info in Element

### LINKED ISSUE #401: Should store an included model's source file name
This can be useful when a world is saved. Relates to a TODO in https://github.com/osrf/sdformat/blob/144f82cf8b994abd2a566a7bbe6ae873a9d92403/src/parser.cc#L1075-L1082

azeey:
# 🎉 New feature

Closes #401 

## Summary
This stores the contents of `<include>` inside the Element loaded from the included file.
For example, given the following SDFormat:
```xml
<sdf version = '1.8'>
  <world name='default'>
    <include>
      <uri>model_uri</uri>
      <pose>1 2 3 0 0 0</pose>
    </include>
  </world>
</sdf>
```

The `ElementPtr` associated with the model loaded from 'model_uri' will
have its `includeElement` set to

```xml
    <include>
      <uri>model_uri</uri>
      <pose>1 2 3 0 0 0</pose>
    </include>
```
This can be used to retrieve additional information available under the <include> tag after the entity has been loaded (eg. `<name>`, `<pose>`, `<plugin>`, etc.). An example use case for this is when saving a loaded world back to SDFormat in such a way that the `<include>` tags are left intact.

This also deprecates the functions `Element::SetInclude` and `Element::GetInclude` since they were never used. Based on the code  in `Element::ToString`: https://github.com/osrf/sdformat/blob/54d9018599ce6605cee13d372c7dc51f0c0fcd2d/src/Element.cc#L528-L537
I think these functions were added a long time ago when the syntax for including files was `<include filename="path">`, which is not valid SDFormat anymore.

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -493,7 +493,7 @@ namespace sdf
 
     /// \brief The <include> element that was used to load this entity. For
     /// example, given the following SDFormat:
-    /// <sdf version = '1.8'>
+    /// <sdf version='1.8'>
     ///   <world name='default'>
     ///     <include>
     ///       <uri>model_uri</uri>
@@ -501,7 +501,7 @@ namespace sdf
     ///     </include>
     ///   </world>
     /// </sdf>
-    /// The ElementPtr associated with the model loaded from 'model_uri' will
+    /// The ElementPtr associated with the model loaded from `model_uri` will
     /// have the includeElement set to
     ///     <include>
     ///       <uri>model_uri</uri>
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -385,8 +385,9 @@ namespace sdf
     public: std::string GetInclude() const SDF_DEPRECATED(11.0);
 
     /// \brief Set the <include> element that was used to load this element.
-    /// This set by the parser on the first element of the included object (eg.
-    /// Model, Actor, Light, etc). It is not passed down to children elements.
+    /// This is set by the parser on the first element of the included object
+    /// (eg. Model, Actor, Light, etc). It is not passed down to children
+    /// elements.
     /// \param[in] _includeELem The <include> Element
     public: void SetIncludeElement(sdf::ElementPtr _includeElem);
 
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -378,11 +378,23 @@ namespace sdf
 
     /// \brief Set the include filename to the passed in filename.
     /// \param[in] _filename the filename to set the include filename to.
-    public: void SetInclude(const std::string &_filename);
+    public: void SetInclude(const std::string &_filename) SDF_DEPRECATED(11.0);
 
     /// \brief Get the include filename.
     /// \return The include filename.
-    public: std::string GetInclude() const;
+    public: std::string GetInclude() const SDF_DEPRECATED(11.0);
+
+    /// \brief Set the <include> element that was used to load this element.
+    /// This set by the parser on the first element of the included object (eg.
+    /// Model, Actor, Light, etc). It is not passed down to children elements.
+    /// \param[in] _includeELem The <include> Element
+    public: void SetIncludeElement(sdf::ElementPtr _includeElem);
+
+    /// \brief Get the <include> element that was used to load this element.
+    /// \return The Element pointer to the <include> element, if this element
+    /// was loaded from an included file. Otherwise, nullptr. This is also
+    /// nullptr for Elements that cannot be included.
+    public: sdf::ElementPtr GetIncludeElement() const;
 
     /// \brief Set the path to the SDF document where this element came from.
     /// \param[in] _path Full path to SDF document.
@@ -478,6 +490,28 @@ namespace sdf
 
     // The possible child elements
     public: ElementPtr_V elementDescriptions;
+
+    /// \brief The <include> element that was used to load this entity. For
+    /// example, given the following SDFormat:
+    /// <sdf version = '1.8'>
+    ///   <world name='default'>
+    ///     <include>
+    ///       <uri>model_uri</uri>
+    ///       <pose>1 2 3 0 0 0</pose>
+    ///     </include>
+    ///   </world>
+    /// </sdf>
+    /// The ElementPtr associated with the model loaded from 'model_uri' will
+    /// have the includeElement set to
+    ///     <include>
+    ///       <uri>model_uri</uri>
+    ///       <pose>1 2 3 0 0 0</pose>
+    ///     </include>
+    ///
+    /// This can be used to retrieve additional information available under the
+    /// <include> tag after the entity has been loaded. An example use case for
+    /// this is when saving a loaded world back to SDFormat.
+    public: ElementPtr includeElement;
 
     /// name of the include file that was used to create this element
     public: std::string includeFilename;
--- a/src/Element.cc
+++ b/src/Element.cc
@@ -195,6 +195,11 @@ ElementPtr Element::Clone() const
     clone->dataPtr->value = this->dataPtr->value->Clone();
   }
 
+  if (this->dataPtr->includeElement)
+  {
+    clone->dataPtr->includeElement = this->dataPtr->includeElement->Clone();
+  }
+
   return clone;
 }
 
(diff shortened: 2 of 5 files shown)


src/Element_TEST.cc
```diff
@@ -158,11 +158,28 @@ TEST(Element, Include)
 {
   sdf::Element elem;
 
+  SDF_SUPPRESS_DEPRECATED_BEGIN
```
EricCousineau-TRI, line 161: BTW Ooh, fancy! Much easier to read than `#pramga RAWRRR` stuff.

Less silly, more concrete example:
https://github.com/RobotLocomotion/drake/blob/v0.27.0/bindings/pydrake/common/test/deprecation_example/cc_module_py.cc#L57-L62

include/sdf/Element.hh
```diff
@@ -479,6 +491,28 @@ namespace sdf
...
+    /// example, given the following SDFormat:
+    /// <sdf version = '1.8'>
+    ///   <world name='default'>
+    ///     <include>
+    ///       <uri>model_uri</uri>
+    ///       <pose>1 2 3 0 0 0</pose>
+    ///     </include>
+    ///   </world>
+    /// </sdf>
+    /// The ElementPtr associated with the model loaded from 'model_uri' will
```
EricCousineau-TRI, line 504: nit Consider surrounding actual code elements with backticks, e.g. `` `module_uri` ``.

Doxygen supports its own flavor of Markdown.

include/sdf/Element.hh
```diff
@@ -479,6 +491,28 @@ namespace sdf
     // The possible child elements
     public: ElementPtr_V elementDescriptions;
 
+    /// \brief The <include> element that was used to load this entity. For
+    /// example, given the following SDFormat:
+    /// <sdf version = '1.8'>
```
EricCousineau-TRI, line 496: nit Why the space between ` = `?

What does `xmllint` say about that?

test/integration/nested_model.cc
```diff
@@ -1371,3 +1371,87 @@ TEST(NestedReference, PlacementFrameElement)
...
   }
 }
+
+/////////////////////////////////////////////////
+TEST(NestedModel, IncludeElements)
+{
+  const std::string modelRootPath = sdf::filesystem::append(
+      PROJECT_SOURCE_PATH, "test", "integration", "model");
+
+  sdf::setFindCallback(
```
EricCousineau-TRI, line 1381: nit Consider using non-global version? ;)

test/integration/nested_model.cc
```diff
@@ -1371,3 +1371,87 @@ TEST(NestedReference, PlacementFrameElement)
...
+    EXPECT_EQ("test_plugin", pluginElem->Get<std::string>("name"));
+    ASSERT_TRUE(pluginElem->HasAttribute("filename"));
+    EXPECT_EQ("test_plugin_file", pluginElem->Get<std::string>("filename"));
+  };
+
+  // Test with //world/include
+  {
+    std::ostringstream stream;
+    stream << R"(
+    <sdf version='1.8'>
```
EricCousineau-TRI, line 1416: nit (minor) Inconsistent usage of quotes, `'` and `"`

test/integration/nested_model.cc
```diff
@@ -1371,3 +1371,87 @@ TEST(NestedReference, PlacementFrameElement)
...
+        box->Element()->GetIncludeElement(), "box", "test_box");
+
+    auto *test2 = world->ModelByIndex(1);
+    ASSERT_NE(nullptr, test2);
+    EXPECT_EQ(nullptr, test2->Element()->GetIncludeElement());
+
+    auto *testModel2 = test2->ModelByIndex(0);
+    ASSERT_NE(nullptr, testModel2);
+    checkIncludeElem(
+        testModel2->Element()->GetIncludeElement(), "test_model", "test_model");
```
EricCousineau-TRI, line 1455: I'm not fully 1000% convinced that this preserves the information across load str -> element -> dump str.

Can you add a quick idempotent check that the include markup is preserved as-is?

EricCousineau-TRI (approved):
Sweet! Minor nits, and a testing update.

test/integration/nested_model.cc
EricCousineau-TRI: Per f2f, this implies that the `//include` tags would be preserved, which *should not* be the default libsdformat behavior.

Perhaps there should be an option (API + cmdline) to preserve includes?

src/Element.cc
```diff
@@ -912,6 +929,18 @@ std::string Element::GetInclude() const
   return this->dataPtr->includeFilename;
 }
 
+/////////////////////////////////////////////////
+void Element::SetIncludeElement(sdf::ElementPtr _includeElem)
+{
+  this->dataPtr->includeElement = _includeElem;
+}
```
jennuine, line 936: Should this also update `this->dataPtr->includeFilename`?

jennuine (commented):
Looks good! Did a first pass and left some minor comments. I'm curious, is `includeFilename` needed anymore?

If so, should `Element::ToString()` be updated since `<include filename="path">` is no longer valid? (or removed)
https://github.com/osrf/sdformat/blob/69dc1542641df2586d2e8dcfb02e307a8a865fdb/src/Element.cc#L535-L536

aaronchongth (commented):
This is awesome!

On the topic of the `ToString` function, would a good way of re-implementing it be to use the output of `PrintValuesImpl`, but modify the fields depending on what is stored in `includeElement`, the `pose`, `static` flags, etc?

src/Element.cc
azeey: My intention was to deprecate `SetInclude` and along with it, `includeFileName`. We already have `Element::FilePath` so, `includeFileName` is redundant.

src/Element_TEST.cc
azeey: For sure! Stole it from DART https://github.com/dartsim/dart/blob/master/dart/common/Deprecated.hpp

test/integration/nested_model.cc
azeey: I created an issue to capture this: https://github.com/osrf/sdformat/issues/522. How about we tackle that in a follow up PR?

test/integration/nested_model.cc
azeey: I'm terrible at staying consistent with this. Changed to `'`

test/integration/nested_model.cc
azeey: Using non-global version in https://github.com/osrf/sdformat/pull/509/commits/2631da14dff94ea7c66e2480be626ffd3e33e72e

include/sdf/Element.hh
azeey: hmm, just a force of habit for other times I use `=` in code. `xmllint` does remove all spaces. Removed in https://github.com/osrf/sdformat/pull/509/commits/2631da14dff94ea7c66e2480be626ffd3e33e72e.

include/sdf/Element.hh
azeey: Done in https://github.com/osrf/sdformat/pull/509/commits/2631da14dff94ea7c66e2480be626ffd3e33e72e

azeey:
> On the topic of the ToString function, would a good way of re-implementing it be to use the output of PrintValuesImpl, but modify the fields depending on what is stored in includeElement, the pose, static flags, etc?

I think this is actually how it works currently. The parser overrides all those parameters of the included model based on the contents of the `<include>` element when it first loads the include model: https://github.com/osrf/sdformat/blob/2631da14dff94ea7c66e2480be626ffd3e33e72e/src/parser.cc#L1225-L1291

test/integration/nested_model.cc
EricCousineau-TRI: Sounds great!

include/sdf/Element.hh
```diff
@@ -378,11 +378,23 @@ namespace sdf
...
-    public: void SetInclude(const std::string &_filename);
+    public: void SetInclude(const std::string &_filename) SDF_DEPRECATED(11.0);
 
     /// \brief Get the include filename.
     /// \return The include filename.
-    public: std::string GetInclude() const;
+    public: std::string GetInclude() const SDF_DEPRECATED(11.0);
+
+    /// \brief Set the <include> element that was used to load this element.
+    /// This set by the parser on the first element of the included object (eg.
```
jennuine, line 388:
```suggestion
    /// This is set by the parser on the first element of the included object (eg.
```

jennuine (approved):
LGTM! Minor nit: I think one of your comments is missing "is" in the statement but feel free to disregard

include/sdf/Element.hh
azeey: Fixed in https://github.com/osrf/sdformat/pull/509/commits/51d0c0f175171d33faf032f8eac37532697efd50

PR #515, opened 2021-03-16
Error when delimiter "::" found in element name in SDFormat 1.8

### LINKED ISSUE #420: Using the delimiter "::" in a name should be an error in SDFormat 1.8
Per the SDFormat 1.8 composition [proposal](http://sdformat.org/tutorials?tut=composition_proposal&cat=pose_semantics_docs&#1-3-1-reserved-delimiter-token), the `::` delimiter is reserved for forming scopes, thus, it should not be used in `name` attributes. However, SDFormat 1.7 files can contain `::` in `name` attributes as a result of model flattening. So, we should only emit an error if the file is SDFormat 1.8.

jennuine:
Closes #420 

## Summary
Error when the `::` delimiter is found in `name` attributes for files that are SDFormat >= 1.8

## Checklist
- [X] Added tests
- [X] Code check passed (In source directory, run `sh tools/code_check.sh`)
- [x] While waiting for a review on your PR, please help review
to support the maintainers

### DIFF
--- a/include/sdf/parser.hh
+++ b/include/sdf/parser.hh
@@ -417,10 +417,11 @@ namespace sdf
   SDFORMAT_VISIBLE
   bool recursiveSiblingUniqueNames(sdf::ElementPtr _elem);
 
-  /// \brief Check that all sibling elements do not the delimiter double colons
-  /// '::' in element names which is reserved for forming scopes in SDFormat 1.8
-  /// This checks recursively and should check the files exhaustively rather
-  /// than terminating early when the first name containing '::' is found.
+  /// \brief Check that all sibling elements do not contain the delimiter
+  /// double colons '::' in element names, which is reserved for forming scopes
+  /// in SDFormat 1.8. This checks recursively and should check the files
+  /// exhaustively rather than terminating early when the first name
+  /// containing '::' is found.
   /// \param[in] _elem SDF Element to check recursively.
   /// \return True if all contained element names do not have the delimiter '::'
   SDFORMAT_VISIBLE
--- a/include/sdf/parser.hh
+++ b/include/sdf/parser.hh
@@ -417,6 +417,15 @@ namespace sdf
   SDFORMAT_VISIBLE
   bool recursiveSiblingUniqueNames(sdf::ElementPtr _elem);
 
+  /// \brief Check that all sibling elements do not the delimiter double colons
+  /// '::' in element names which is reserved for forming scopes in SDFormat 1.8
+  /// This checks recursively and should check the files exhaustively rather
+  /// than terminating early when the first name containing '::' is found.
+  /// \param[in] _elem SDF Element to check recursively.
+  /// \return True if all contained element names do not have the delimiter '::'
+  SDFORMAT_VISIBLE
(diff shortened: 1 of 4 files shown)


src/parser.cc
```diff
@@ -685,6 +685,17 @@ bool readDoc(tinyxml2::XMLDocument *_xmlDoc, SDFPtr _sdf,
           "Error reading element <" + _sdf->Root()->GetName() + ">"});
       return false;
     }
+
+    // delimiter '::' in element names not allowed in SDFormat >= 1.8
+    if (std::stof(_sdf->Root()->OriginalVersion()) >= 1.8f
```
scpeters, line 690: this logic won't work when we reach specification version `1.10`. I recommend using `ignition::math::SemanticVersion`

* https://github.com/osrf/sdformat/blob/sdformat9_9.5.0/src/Model.cc#L161
* https://github.com/osrf/sdformat/blob/sdformat9_9.5.0/src/Model.cc#L247

scpeters (changes requested):

src/parser.cc
jennuine: Thanks! 56e47a73303264cfeab1c7ee4aebd14083cee0dd

include/sdf/parser.hh
```diff
@@ -417,6 +417,15 @@ namespace sdf
   SDFORMAT_VISIBLE
   bool recursiveSiblingUniqueNames(sdf::ElementPtr _elem);
 
+  /// \brief Check that all sibling elements do not the delimiter double colons
```
scpeters, line 420: missing a verb here: "Check that all sibling elements do not the delimiter double colons"

do not *contain*?

src/parser_TEST.cc
```diff
@@ -353,6 +353,31 @@ TEST(Parser, PlacementFrameMissingPose)
...
+  sdf::SDFPtr sdf = InitSDF();
+  std::ostringstream stream;
+  stream << "<?xml version=\"1.0\"?>"
+         << "<sdf version='1.8'>"
+         << "  <model name='test'>"
+         << "    <link name='A::B'/>"
+         << "  </model>"
+         << "</sdf>";
+
+  EXPECT_FALSE(sdf::readString(stream.str(), sdf));
```
scpeters, line 369: nit: you could pass an `sdf::Errors` object to `sdf::readString` and check its contents

scpeters (approved):
looks good; I just left some minor comments

nice work!

include/sdf/parser.hh
jennuine: Yup, thanks! ac6ec01a8bfde7cc3e36fdac8c3f3d6b6ec4ac52

src/parser_TEST.cc
jennuine: ac6ec01a8bfde7cc3e36fdac8c3f3d6b6ec4ac52

scpeters: looks great!

jennuine:
Waiting for https://github.com/osrf/sdformat/pull/504 to be merged before merging this one

PR #525, opened 2021-03-30
Fixed application of <sensor><pose> tags in lumped linkes during URDF conversion

### LINKED ISSUE #67: Pose tag in sensors ignored with fixed links
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/67)) by adhenning (Bitbucket: [adhenning](https://bitbucket.org/%7Baa99a57e-7bf1-4a88-ad27-a3cdd36e629c%7D/)).**

The original report had attachments: [GearsBot.URDF](https://osrf-migration.github.io/sdformat-gh-pages/data/repositories/osrf/sdformat/issues/67/attachments/GearsBot.URDF)

When a sensor (or at least camera's in this case) is attached to a fixed link, it's pose tag is ignored and overwritten with the computed pose for the fixed link when the pose tag should be transformed by the computed pose. I've attach the file to reproduce this issue. The following cases show the error. The pose should change, but doesn't.

No pose (0, 0, 0, 0, 0, 0):

    <sensor name="robotCamera" type="camera">
      <visualize>true</visualize>
      
      <camera></camera>
    </sensor>

Output:

    $ gz sdf -p /usr/share/frcsim/models/GearsBot/robots/GearsBot.URDF | awk "/<sensor name='robotCamera/,/<\/sensor>/"
      <sensor name='robotCamera' type='camera'>
        <visualize>1</visualize>
        <camera name='__default__'>
          <horizontal_fov>1.047</horizontal_fov>
          <image>
            <width>320</width>
            <height>240</height>
          </image>
          <clip>
            <near>0.1</near>
            <far>100</far>
          </clip>
        </camera>
        <pose>-0.20115 0.42488 0.30943 1.5708 -0.89012 1.5708</pose>
      </sensor>

Pose:

    <sensor name="robotCamera" type="camera">
      <visualize>true</visualize>
      <pose>1 1 1 1.570796 1.570796 1.570796</pose>
      <camera></camera>
    </sensor>

Output:

      $ gz sdf -p /usr/share/frcsim/models/GearsBot/robots/GearsBot.URDF | awk "/<sensor name='robotCamera/,/<\/sensor>/"
      <sensor name='robotCamera' type='camera'>
        <visualize>1</visualize>
        <camera name='__default__'>
          <horizontal_fov>1.047</horizontal_fov>
          <image>
            <width>320</width>
            <height>240</height>
          </image>
          <clip>
            <near>0.1</near>
            <far>100</far>
          </clip>
        </camera>
        <pose>-0.20115 0.42488 0.30943 1.5708 -0.89012 1.5708</pose>
      </sensor>

peci1:
# 🦟 Bug fix

Fixes #67 and #378.

## Summary
There is a TODO in

https://github.com/osrf/sdformat/blob/8496164cac259e13083023efb06b76a04b1f8d3e/src/parser_urdf.cc#L3374-L3375 

which states that `<sensor><pose>` tags are being silently ignored. This is the culprit of bugs #67 and #378. And of many headaches people spend converting their sensors, finding out after hours that they have to attach them to non-lumped links without any offset for the conversion to work.

This PR fixes the behavior and adds correct processing of `<sensor><pose>` and `<projector><pose>` for lumped links.

In addition to ignoring the `<sensor><pose>` values, there is also weird behavior. First,

https://github.com/osrf/sdformat/blob/8496164cac259e13083023efb06b76a04b1f8d3e/src/parser_urdf.cc#L3378

thinks it deletes the user-specified `<sensor><pose>` tag.

Second,

https://github.com/osrf/sdformat/blob/8496164cac259e13083023efb06b76a04b1f8d3e/src/parser_urdf.cc#L3406

thinks it adds the computed transform to `<sensor><pose>`.

The problem is that both of these lines of code work one level higher than they should. `(*_blobIt)` is a XML snippet containing the `<sensor>` tag, so there is no direct child `<pose>` of `(*_blobIt)`. The second line then adds a new element `<pose>`, but again, as a sibling of `<sensor>`, not as a child. This is interesting and might have had unforseen consequences (as multiple pose tags are allowed, at least syntactically). Fortunately, there is a guard at 

https://github.com/osrf/sdformat/blob/8496164cac259e13083023efb06b76a04b1f8d3e/src/parser_urdf.cc#L2128

that only allows the first child of `(*_blobIt)` to be passed further. This might look like another bug, but I actually utilized this behavior sealing it for the future (it has been this way for a decade, so...).

Currently, when there is no `<sensor><pose>` specified, the conversion process doesn't emit any `<sensor><pose>` in the end, even though it should accumulate something and I see it adding and removing the XML fragments. That is because of the fact it adds the `<pose>` at wrong level, and in the next level, it deletes/re-adds this sibling of `<sensor>`. With `<sensor><pose>` manually specified, it makes it appear in the result (but without the accumulated transform). That's because the `DeleteChild` function has no effect on `<sensor><pose>` but deletes the sibling `<pose>` instead (TinyXML2 doesn't fail when deleting a wrong child).

The fix I propose in this PR is somewhat complicated to grasp on. First idea of just storing `_reductionTransform * sensorPose` in the `<sensor><pose>` field does not work as all but the first lump transforms get doubled. What I do is I take the `<sensor><pose>` when first encountering it and store the `<pose>` as a sibling of `<sensor>`. As said earlier, this sibling will not bubble up anywhere, but it is the scratchpad that is needed. `<sensor><pose>` is then used for storing the actual `_reductionTransform * sensorPose`. What is important is that the original `<sensor><pose>` is not updated and we always have access to it in the original form, so that we can take the recursively computed reduction transform and add the sensor pose exactly once.

I was also looking for a nicer way to parse the 6-tuple transform string, but haven't found any usable part of the library. If you know how e.g. `sdf::Param` could be used to simplify and reuse-ify the code, let me know.

I also added a comprehensive test suite based on both mentioned issues and my experiments. I also verified the change running `ign sdf -p` on https://github.com/osrf/subt/tree/master/submitted_models/ctu_cras_norlab_absolem_sensor_config_1 model which might go as deep as 10 levels of joints/links (or maybe even more) with sensors at the very tips. I tried exchanging the last link transform for a sensor pose and the result was the same.

---

As a workaround until this PR is merged, this is what can be done. 

1. Attach the sensor to a non-lumped link and make the origins of the link and the sensor identical. Non-lumped links are set by `<preserveFixedJoint>true</preserveFixedJoint>`
2. If you want to have the sensor on a lumped link, you have to manually compute the transformation of the sensor to its nearest non-lumped link. Once you have this transformation, add it in `<sensor><pose>`. Even though it seems you're duplicating the tree-induced transform by the `<pose>` tag, you are not because of the reported bugs. 

---

Interestingly, the part of code I fixed was mostly untouched for 8 years. However, it seems that `gz sdf -p` from Gazebo 9 behaves differently.

## Checklist
- [x] Added tests
- [x] Updated migration guide (as needed) **(how should I do it before submitting the PR when there is need to write down the PR number I don't yet have?)**
- [x] All tests passed (See
to support the maintainers

### DIFF
--- a/test/integration/urdf_gazebo_extensions.cc
+++ b/test/integration/urdf_gazebo_extensions.cc
@@ -99,4 +99,163 @@ TEST(SDFParser, UrdfGazeboExtensionURDFTest)
       EXPECT_TRUE(physics->Get<bool>("implicit_spring_damper"));
     }
   }
+
+  sdf::ElementPtr link0;
+  for (sdf::ElementPtr link = model->GetElement("link"); link;
+       link = link->GetNextElement("link"))
+  {
+    const auto& linkName = link->Get<std::string>("name");
+    if (linkName == "link0")
+    {
+      link0 = link;
+      break;
+    }
+  }
+  ASSERT_TRUE(link0);
+
+  bool foundSensorNoPose {false};
+  bool foundSensorPose {false};
+  bool foundSensorPoseRelative {false};
+  bool foundSensorPoseTwoLevel {false};
+  bool foundIssue378Sensor {false};
+  bool foundIssue67Sensor {false};
+
+  for (sdf::ElementPtr sensor = link0->GetElement("sensor"); sensor;
+       sensor = sensor->GetNextElement("sensor"))
+  {
+    const auto& sensorName = sensor->Get<std::string>("name");
+    if (sensorName == "sensorNoPose")
+    {
+      foundSensorNoPose = true;
+      ASSERT_TRUE(sensor->HasElement("pose"));
+      const auto poseElem = sensor->GetElement("pose");
+
+      const auto& posePair = poseElem->Get<ignition::math::Pose3d>(
+        "", ignition::math::Pose3d::Zero);
+      ASSERT_TRUE(posePair.second);
+
+      const auto& pose = posePair.first;
+
+      EXPECT_FLOAT_EQ(pose.X(), 333.0);
+      EXPECT_FLOAT_EQ(pose.Y(), 0.0);
+      EXPECT_FLOAT_EQ(pose.Z(), 0.0);
+      EXPECT_FLOAT_EQ(pose.Roll(), 0.0);
+      EXPECT_FLOAT_EQ(pose.Pitch(), 0.0);
+      EXPECT_NEAR(pose.Yaw(), IGN_PI_2, 1e-5);
+
+      EXPECT_FALSE(poseElem->GetNextElement("pose"));
+    }
+    else if (sensorName == "sensorPose")
+    {
+      foundSensorPose = true;
+      ASSERT_TRUE(sensor->HasElement("pose"));
+      const auto poseElem = sensor->GetElement("pose");
+
+      const auto& posePair = poseElem->Get<ignition::math::Pose3d>(
+        "", ignition::math::Pose3d::Zero);
+      ASSERT_TRUE(posePair.second);
+
+      const auto& pose = posePair.first;
+
+      EXPECT_FLOAT_EQ(pose.X(), 333.0);
+      EXPECT_FLOAT_EQ(pose.Y(), 111.0);
+      EXPECT_FLOAT_EQ(pose.Z(), 0.0);
+      EXPECT_FLOAT_EQ(pose.Roll(), 0.0);
+      EXPECT_FLOAT_EQ(pose.Pitch(), 0.0);
+      EXPECT_NEAR(pose.Yaw(), IGN_PI_2 - 1, 1e-5);
+
+      EXPECT_FALSE(poseElem->GetNextElement("pose"));
+    }
+    else if (sensorName == "sensorPoseRelative")
+    {
+      foundSensorPoseRelative = true;
+      ASSERT_TRUE(sensor->HasElement("pose"));
+      const auto poseElem = sensor->GetElement("pose");
+
+      const auto& posePair = poseElem->Get<ignition::math::Pose3d>(
+        "", ignition::math::Pose3d::Zero);
+      ASSERT_TRUE(posePair.second);
+
+      const auto& pose = posePair.first;
+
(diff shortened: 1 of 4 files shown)


nkoenig (approved):
Looks good to me. Thanks for the tests.

test/integration/urdf_gazebo_extensions.cc
```diff
@@ -99,4 +99,163 @@ TEST(SDFParser, UrdfGazeboExtensionURDFTest)
...
+      ASSERT_TRUE(sensor->HasElement("pose"));
+      const auto poseElem = sensor->GetElement("pose");
+
+      const auto& posePair = poseElem->Get<ignition::math::Pose3d>(
+        "", ignition::math::Pose3d::Zero);
+      ASSERT_TRUE(posePair.second);
+
+      const auto& pose = posePair.first;
+
+      EXPECT_FLOAT_EQ(pose.X(), 333.0);
```
nkoenig, line 139: Can you change all of the `EXPECT_FLOAT_EQ` to `EXPECT_DOUBLE_EQ`. It will make windows happy, see https://build.osrfoundation.org/job/sdformat-ci-pr_any-windows7-amd64/2062/msbuild/new/category.63474753/

nkoenig (commented):
Minor CI fix needed.

test/integration/urdf_gazebo_extensions.cc
peci1: Okay, I fixed all the equality tests.

doisyg:
Is it possible to merge also in edifice/sdf11 ? (as it should fix https://github.com/ignitionrobotics/ros_ign/issues/147)

chapulina:
> Is it possible to merge also in edifice/sdf11 ?

Being forward-ported on https://github.com/osrf/sdformat/pull/539

PR #519, opened 2021-03-18
updated loading //material colors

jennuine:
# 🦟 Bug fix

Fixes #193

## Summary
This fix parses `//material` color components (i.e., ambient, diffuse, specular, emissive) from a SDFormat file in `Material::Load` by checking that values are either `r,g,b` or `r,g,b,a` (expecting 3 or 4 values) and that each value is between [0,1]. If the checks do not pass, the default color `0,0,0,1` is used.

## Checklist
- [X] Signed all commits for DCO
- [X] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] Code check passed (In source directory, run `sh tools/code_check.sh`)
- [x] All tests passed (See
  [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review
[another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+)
to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #526, opened 2021-03-30
Allow to convert URDF color to SDF material tag

ahcorde:

# 🦟 Bug fix

This PR tries to fix this issue https://github.com/osrf/sdformat/issues/513 where the URDF gazebo material tags are incorrectly translated

As a general overview:

 - URDF defined the color as `rgba` but in ignition the color is defined as ambient, specular and diffuse. [URDF definition](http://wiki.ros.org/urdf/XML/link)
 - The other way to use a color in the URDF format is with OGRE materials, but according with the migration guide there is no support for this.

This solution will not break the `old` behaviour, if there is any `<material>` tag under a `<gazebo>` tag, then the tag will be replaced. 

## Checklist
- [ ] Added tests
- [ ] All tests passed (See
to support the maintainers


nkoenig:
Is it possible to add a test?

peci1:
Will this be compatible with Gazebo Classic?

ahcorde:
> Will this be compatible with Gazebo Classic in case only the URDF color is defined?

Gazebo11 (classic) is using `sdformat9`, this PR is targeting `sdformat11`. This should not affect Gazebo classic: `Gazebo9` or `Gazebo11`.

What do you think @nkoenig ? should I retarget this to `sdformat9`? and make it compatible with `Citadel` ?

nkoenig:
> > Will this be compatible with Gazebo Classic in case only the URDF color is defined?
> 
> Gazebo11 (classic) is using `sdformat9`, this PR is targeting `sdformat11`. This should not affect Gazebo classic: `Gazebo9` or `Gazebo11`.
> 
> What do you think @nkoenig ? should I retarget this to `sdformat9`? and make it compatible with `Citadel` ?

I don't see a problem leaving this PR targeted toward sdformat11. No reason to potentially upset Gazebo classic.

ahcorde:
@nkoenig  Test added

ahcorde:
I'm going to close this PR because there is a way to include the color without adding new code, for example:

```
<link name='cart'>
...
</link>
<gazebo reference="cart">
    <visual>
      <material>
        <ambient>0 0 1 1</ambient>
        <diffuse>0 0 1 1</diffuse>
        <specular>0 0 1 1</specular>
      </material>
    </visual>
 </gazebo>
```

peci1:
>  there is a way to include the color without adding new code

That's only part-wise solution. This way, you can only set a single material to all `<visual>` tags of the link. Using the URDF color tags, you set the color per-`<visual>`. This can be overcome by splitting the visual elements to multiple lumped links, but it's impractical at least...

ahcorde:
friendly ping @nkoenig

src/parser_urdf.cc
```diff
@@ -3200,6 +3200,35 @@ void CreateVisual(tinyxml2::XMLElement *_elem, urdf::LinkConstSharedPtr _link,
...
+    auto materialTag = sdfVisual->FirstChildElement("material");
+    // If the material is not included by an extension then create it
+    if (materialTag == nullptr)
+    {
+      AddKeyValue(sdfVisual, "material", "");
+      materialTag = sdfVisual->FirstChildElement("material");
+    }
+    // If the specular and diffuse are defined by an extension then don't
+    // use the color values
+    if (materialTag->FirstChildElement("diffuse") == nullptr &&
```
azeey, line 3219: I recently had a discussion about how material tags are translated to urdf color. In the [sdformat_urdf](https://github.com/ros/sdformat_urdf/blob/ros2/sdformat_urdf/README.md#material-limitations) package, this formula was used `urdf color = 0.4 * sdf ambient + 0.8 * sdf diffuse` and the related discussion is here: https://github.com/ros/sdformat_urdf/pull/1#discussion_r509709647

I wonder if we could do the reverse here. I don't have a good suggestion on how though.

ahcorde: Yep, there are 3 equations, each one of the equations have two unknown variables, and there is no relation between them. I neither have a good suggestion. We only know that this variables should live in the interval [0, 1].

peci1: I'm against adding any ambient color by default. That would cause problems in dark areas where the models would glow. Imagine a SubT-like scenario - you don't want any ambient color there...

peci1: That would mean `sdf diffuse = 1.25 * urdf color, sdf ambient = 0`

azeey: @peci1 I don't think adding ambient would cause objects to glow. That would be emissive. Citing http://gazebosim.org/tutorials?tut=color_model:
> The final color of an object depends both on the material and the lights shining on it. Values on lights represent the intensity of light emitted. Values on a material represent the percentage of light an object reflects. The same material may be different colors depending on the light that hits it.

So, if the ambient of the scene is [0, 0, 0], then the ambient component of the objects material won't affect the final color of the object. For SubT, the scene's ambient is [0.1, 0.1, 0.1], so, each of the object's ambient channels would be multiplied by 0.1.

peci1: Hmm, so the influence of a material's `<ambient>` color depends on the global `<ambient>`? Then it would be okay to set some - if somebody wanted to make a totally dark world, it would be possible by setting global ambient to black.

chapulina: +1 to setting both ambient and diffuse to the URDF color, for a "plain color" look. Specular and emissive feel like "extra" features. Users can always use the extension if they need finer control.

ahcorde: https://github.com/osrf/sdformat/pull/526/commits/c58aa5f4bcf5191fc7843fc45005c36a7a09478b

azeey: Thinking through [this](https://github.com/ros/sdformat_urdf/pull/1#discussion_r514517271) comment a little more, I think we can assign `sdf diffuse = 1.25 * urdf` and `sdf ambient = 1.25 * urdf`. 

My reasoning is that [Rviz](https://github.com/ros-visualization/rviz/blob/8bd56fb928b255e095424a5f1fb3ff8a5274602e/src/rviz/robot/robot_link.cpp#L472-L473) takes the urdf color and sets ` urdf diffuse = urdf` and `urdf ambient = 0.5 * urdf` with an ambient light of `rgba(1 1 1 1)` per [this](https://github.com/ros/sdformat_urdf/pull/1#discussion_r514517271) comment, and diffuse light of `rgba (1 1 1 1)` [as well](https://github.com/ros-visualization/rviz/blob/8bd56fb928b255e095424a5f1fb3ff8a5274602e/src/rviz/visualization_manager.cpp#L175). This means, the final diffuse and ambient colors are set by `urdf diffuse` and `urdf ambient`. In Gazebo, let's assume the default ambient light of `rgba(0.4 0.4 0.4 1)` and diffuse light of `rgba(0.8 0.8 0.8 1)`. The material colors are multiplied by these to obtain the final colors, so to get the same final colors as in Rviz, we should do `sdf ambient = urdf ambient / 0.4` and `sdf diffuse = urdf diffuse / 0.8`. But since  ` urdf diffuse = urdf` and `urdf ambient = 0.5 * urdf` , this results in `sdf ambient = 0.5 * urdf / 0.4` and `sdf diffuse = urdf / 0.8`.

We do need to clamp the values to 1.0 after doing the math because values > 1.0 will result in an error (see #519).

ahcorde:
friendly ping @azeey

azeey (approved):
I tried to using an example to compare the colors between Rviz and ign-gazebo (with `examples/world/empty.sdf`). It still doesn't match quite right. ign-gazebo is much brighter. 
image. I don't know if this an issue with ign-gazebo or not. I'll go ahead and approve since the values used for the conversion seem reasonable to me.

ahcorde:
Merging I ticketed an issue

peci1:
Are there plans for backporting to sdf10?

Issue #536, opened 2021-04-07
String parsing results in FilePath as `data-string` when error occurs


aaronchongth:
## Environment
* OS Version: Ubuntu 20.04
* Source or binary build?
  * Source build, branch `sdf11`
  
  

## Description
Adding this test in `parser_TEST.cc`,
```cpp
TEST(Parser, ReadStringError)
{
  sdf::SDFPtr sdf = InitSDF();
  std::ostringstream stream;
  stream
    << "<sdf version='1.8'>"
    << "<model name=\"test\">"
    << "  <link name=\"test1\">"
    << "    <visual name=\"good\">"
    << "      <geometry>"
    << "        <box><size>1 1 1</size></box>"
    << "      </geometry>"
    << "    </visual>"
    << "  </link>"
    << "  <link>"
    << "    <visual name=\"good2\">"
    << "      <geometry>"
    << "        <box><size>1 1 1</size></box>"
    << "      </geometry>"
    << "    </visual>"
    << "  </link>"
    << "</model>"
    << "</sdf>";
  sdf::Errors errors;
  EXPECT_FALSE(sdf::readString(stream.str(), sdf, errors));
  ASSERT_NE(errors.size(), 0u);

  std::cerr << errors[0] << std::endl;

  EXPECT_EQ(errors[0].Code(), sdf::ErrorCode::ATTRIBUTE_MISSING);
  EXPECT_FALSE(errors[0].FilePath().has_value());
  EXPECT_FALSE(errors[0].LineNumber().has_value());
}
```

* Expected behavior: 

Expected `FilePath` and `LineNumber` to be missing.

```bash
Error Code 4: Msg: Required attribute[name] in element[link] is not specified in SDF.
```

* Actual behavior: 

However due to the source being set to `data-string`, we get this.

```bash
Error Code 4: [data-string:L1]:  Msg: Required attribute[name] in element[link] is not specified in SDF.
```

## Steps to reproduce

1. Add code block into `parser_TEST.cc`
2. Build.
3. Run `./build/sdformat11/src/UNIT_parser_TEST`

## Output

Gist, https://gist.github.com/aaronchongth/74c9641202dbdc4e60142a8547449a1f

EricCousineau-TRI:
I'd propose keeping the `<fake token>:L#` approach. Just change `data-string` to `<data-string>` to make it more clear that it's not a real file.

[...]

PR #537, opened 2021-04-07
Fixed file path in errors when parsing sdf or urdf from string

aaronchongth:
Signed-off-by: Aaron Chong <aaronchongth@gmail.com>

# 🦟 Bug fix

Fixes https://github.com/osrf/sdformat/issues/536

## Summary

* Using fixed strings to highlight string inputs for sdf or urdf
* Error checks FilePath during construction
* Added tests

## Checklist
- [x] Added tests
- [x] All tests passed (See
- [x] While waiting for a review on your PR, please help review
to support the maintainers

### DIFF
--- a/include/sdf/Filesystem.hh
+++ b/include/sdf/Filesystem.hh
@@ -128,6 +128,15 @@ namespace sdf
       IGN_UTILS_UNIQUE_IMPL_PTR(dataPtr)
     };
   }
+
+  /// \brief The source path replacement if it was parsed from a string,
+  /// instead of a file.
+  const std::string sdfStringSource = "data-string";
+
+  /// \brief The source path replacement if the urdf was parsed from a string,
+  /// instead of a file.
+  const std::string urdfStringSource = "urdf string";
+
   }
 }
 
--- a/include/sdf/Filesystem.hh
+++ b/include/sdf/Filesystem.hh
@@ -136,7 +136,6 @@ namespace sdf
   /// \brief The source path replacement if the urdf was parsed from a string,
   /// instead of a file.
   const std::string urdfStringSource = "urdf string";
-
   }
 }
 
--- a/include/sdf/Filesystem.hh
+++ b/include/sdf/Filesystem.hh
@@ -128,14 +128,6 @@ namespace sdf
       IGN_UTILS_UNIQUE_IMPL_PTR(dataPtr)
     };
   }
-
-  /// \brief The source path replacement if it was parsed from a string,
-  /// instead of a file.
-  const std::string sdfStringSource = "<data-string>";
-
-  /// \brief The source path replacement if the urdf was parsed from a string,
-  /// instead of a file.
-  const std::string urdfStringSource = "<urdf-string>";
   }
 }
 
--- a/include/sdf/Filesystem.hh
+++ b/include/sdf/Filesystem.hh
@@ -131,11 +131,11 @@ namespace sdf
 
   /// \brief The source path replacement if it was parsed from a string,
   /// instead of a file.
-  const std::string sdfStringSource = "data-string";
+  const std::string sdfStringSource = "<data-string>";
 
   /// \brief The source path replacement if the urdf was parsed from a string,
   /// instead of a file.
-  const std::string urdfStringSource = "urdf string";
+  const std::string urdfStringSource = "<urdf-string>";
   }
 }
 
--- a/src/Element.cc
+++ b/src/Element.cc
@@ -51,7 +51,7 @@ void Element::SetParent(const ElementPtr _parent)
 
   // If this element doesn't have a path, get it from the parent
   if (nullptr != _parent && (this->FilePath().empty() ||
-      this->FilePath() == "data-string"))
+      this->FilePath() == sdfStringSource))
   {
     this->SetFilePath(_parent->FilePath());
   }
--- a/src/Element.cc
+++ b/src/Element.cc
@@ -22,6 +22,7 @@
 #include "sdf/Assert.hh"
 #include "sdf/Element.hh"
 #include "sdf/Filesystem.hh"
+#include "parser_private.hh"
 
 using namespace sdf;
 
@@ -51,7 +52,7 @@ void Element::SetParent(const ElementPtr _parent)
 
   // If this element doesn't have a path, get it from the parent
   if (nullptr != _parent && (this->FilePath().empty() ||
-      this->FilePath() == sdfStringSource))
+      this->FilePath() == std::string(sdfStringSource)))
   {
     this->SetFilePath(_parent->FilePath());
   }
--- a/src/Element.cc
+++ b/src/Element.cc
@@ -22,7 +22,6 @@
 #include "sdf/Assert.hh"
 #include "sdf/Element.hh"
 #include "sdf/Filesystem.hh"
-#include "parser_private.hh"
 
 using namespace sdf;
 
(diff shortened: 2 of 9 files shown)


EricCousineau-TRI (commented):
I've posted some feedback in #536; can you look at it and let me know what you think?

include/sdf/Filesystem.hh
```diff
@@ -128,6 +128,14 @@ namespace sdf
       IGN_UTILS_UNIQUE_IMPL_PTR(dataPtr)
     };
   }
+
+  /// \brief The source path replacement if it was parsed from a string,
```
EricCousineau-TRI, line 132: BTW Note that `std::string` has a non-trivial destructor (when SSO (short-string opt) ain't happening). Please note that this violates GSG:
https://ignitionrobotics.org/docs/all/contributing#style-guides
https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

See also: Drake's mitigation:
https://drake.mit.edu/code_style_guide.html
https://drake.mit.edu/styleguide/cppguide.html#Static_and_Global_Variables (`drake::never_destroyed<T>`)


This may be an existing defect, though.

include/sdf/Filesystem.hh
```diff
@@ -128,6 +128,14 @@ namespace sdf
...
+  /// instead of a file.
```
EricCousineau-TRI, line 133: nit Instead of `const std::string`, consider instead using `constexpr char[]` since you're in C++17 land.

include/sdf/Filesystem.hh
```diff
@@ -128,6 +128,14 @@ namespace sdf
...
+  const std::string sdfStringSource = "<data-string>";
```
EricCousineau-TRI, line 134: Is there a way to make this private API for now? Or do you see this as an important part of a user's workflow?

include/sdf/Filesystem.hh
EricCousineau-TRI: @azeey Can you advise whether this is important, or can be ignored for now?

include/sdf/Filesystem.hh
```diff
@@ -128,6 +128,14 @@ namespace sdf
...
+
+  /// \brief The source path replacement if the urdf was parsed from a string,
+  /// instead of a file.
+  const std::string urdfStringSource = "<urdf-string>";
```
EricCousineau-TRI, line 138: nit This and above violate GSG for constant names:
https://google.github.io/styleguide/cppguide.html#Constant_Names

@azeey Is this what `libsdformat` normally does (not using `kCameCase`), or should it use GSG recommendations?

EricCousineau-TRI (approved):
LGTM! Just made some comments about GSG violations, but not sure if they're super important w.r.t. current adherence in `libsdformat`. Thanks!

include/sdf/Filesystem.hh
aaronchongth: Oh thanks for pointing this out, I'll be mindful of it going forwards!

include/sdf/Filesystem.hh
aaronchongth: Great point, as far as I know there should not be any public usage, I'll look into into getting it into implementation land.

aaronchongth: https://github.com/osrf/sdformat/pull/537/commits/745ea3bf520de5c16927d2dccade26bd3db73b85

include/sdf/Filesystem.hh
aaronchongth: https://github.com/osrf/sdformat/pull/537/commits/745ea3bf520de5c16927d2dccade26bd3db73b85

include/sdf/Filesystem.hh
aaronchongth: Found some references in `ign-sensors`, https://github.com/ignitionrobotics/ign-sensors/blob/ign-sensors5/src/AirPressureSensor.cc#L42-L51.

I'll wait and see what @azeey thinks

aaronchongth:
@osrf-jenkins run test

azeey (commented):
Looks good. Just a minor comment about where constants are defined.

src/Element.cc
```diff
@@ -22,6 +22,7 @@
 #include "sdf/Assert.hh"
 #include "sdf/Element.hh"
 #include "sdf/Filesystem.hh"
+#include "parser_private.hh"
```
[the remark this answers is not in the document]
aaronchongth: Fixed in https://github.com/osrf/sdformat/pull/537/commits/f8136067bbe0d794201629438b033905877f1e83

include/sdf/Filesystem.hh
azeey: I somehow missed these comments, but yes, we should use `kCamelCase`.

include/sdf/Filesystem.hh
azeey: I found a place outside of libsdformat where this is used: https://github.com/ignitionrobotics/ign-gazebo/blob/57d583390c6c9d3f39f520f0570a9637f5060fcd/src/Util.cc#L289. I suggest:

1. We revert to `data-string` instead of `<data-string>`. We can put the `< >` just in the console message.
2. We revert to `urdf string` instead of `<urdf-string>`.
3. We explicitly make this part of the public API by creating constants in `Types.hh`.

aaronchongth: that works! I've also verified that the paths from `TinyXML2` are absolute and will always be prefixed with a `/`, so there should be no chance for the edge case where the file paths names are `data-string` or `urdf string`

aaronchongth: https://github.com/osrf/sdformat/pull/537/commits/b922496a5ae2dc4acf2c469721678b7952ff8c4d

azeey (approved):

azeey:
oops, @aaronchongth  please use squash merge next time.

aaronchongth:
@azeey I apologize, I thought I selected squash merge for the auto-merging, I'll do it manually in the future just in case

include/sdf/Filesystem.hh
EricCousineau-TRI: Sorry to page in late here; is it hard to fix those downstream usages in subsequent PR?
First off, (1) and (2) are inconsistent. Second off, it's still kinda defective in terms of #536.

Given both are pre-release, I think we could change?

Lemme take a quick whack at this.

EricCousineau-TRI: Filed #551

EricCousineau-TRI: Filed https://github.com/ignitionrobotics/ign-gazebo/pull/794

azeey: Not sure why (1) and (2) are inconsistent. I'm just suggesting we revert to the original strings: `data-string` and `urdf string`.

aaronchongth: Sorry for the lack of visibility, the suggestion to revert back to `data-string` and `urdf string` without the triangle brackets was made during a call. 

https://github.com/osrf/sdformat/pull/537/commits/b922496a5ae2dc4acf2c469721678b7952ff8c4d is the final commit that reverts it and puts together the triangle brackets separately.

EricCousineau-TRI: > Not sure why (1) and (2) are inconsistent. 

Soz, that was poorly worded. I meant to say having one form the file appear with `< >` (console output) and another form without (API) is kinda jarring / inconsistent.

> I'm just suggesting we revert to the original strings [...]

I think this is a poorly motivated solution, tho? (possibly easy to fix downstream rather than have weird "imbalanced" tokens)
Hence the PR + issue.

> Sorry for the lack of visibility [...]

No worries! I think the motivation was sufficiently visible, just didn't agree w/ motivation :sweat_smile:

PR #543, opened 2021-04-21
Add an EnforcementPolicy for deprecated elements

### LINKED ISSUE #541: EnforcementPolicy: Shouldn't strictly turn deprecations into errors?
## Environment
N/A. Source code rather.

## Description
* Expected behavior: Devs / users should have knobs to let deprecations trigger warnings or errors, separate from (or extending) `sdf::EnforcementPolicy::ERR`.
* Actual behavior: Deprecations are warnings, and turn into errors with `sdf::EnforcementPolicy::ERR`, which hurts libs that want to be strict, but still admit deprecations.

See: https://github.com/robotlocomotion/drake/pull/14852#pullrequestreview-635709616

Relates #481 

## Steps to reproduce
N/A

## Output
N/A

azeey:
# 🎉 New feature

Closes #541

## Summary
If the warning policy of the parser is set to `ERR` in the `sdf::ParserConfig`, the parser emits an error message for deprecated elements. This PR adds a separate `EnforcementPolicy` for deprecated elements so that the behavior of the parser can be configured by the user. The API is such, by default `DeprecatedElementsPolicy` reflects the value of `WarninPolicy`,  but the user can call `SetDeprecatedElementsPolicy` to override the value. Note that the order of the function calls does not matter. Once the `DeprecatedElementsPolicy` overridden, calling `WarninPolicy` will not change the value of `DeprecatedElementsPolicy` unless `ResetDeprecatedElementsPolicy` is called. I'm open to feedback on this API.

## Test it
Tests are in `test/integration/deprecated_specs.cc`

## Checklist
- [x] Added tests
to support the maintainers

### DIFF
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -101,6 +101,7 @@ namespace sdf
 
     /// \brief Set the requirement type.
     /// \param[in] _req Requirement type for this element:
+    /// -1: Deprecated.
     /// 0: Not required.
     /// 1: Exactly one element is required.
     /// +: One or more elements are required.
--- a/include/sdf/ParserConfig.hh
+++ b/include/sdf/ParserConfig.hh
@@ -152,7 +152,7 @@ class SDFORMAT_VISIBLE ParserConfig
   /// WarningsPolicy.
   public: void ResetDeprecatedElementsPolicy();
 
-  /// \brief Get the current deperacted elements policy. By default, the policy
+  /// \brief Get the current deprecated elements policy. By default, the policy
   /// is the same as the overall WarningsPolicy, but it can be overriden by
   /// SetDeprecatedElementsPolicy. Once it is overriden, changing
   /// SetWarningsPolicy will not change the value of DeprecatedElementsPolicy
--- a/include/sdf/ParserConfig.hh
+++ b/include/sdf/ParserConfig.hh
@@ -144,6 +144,22 @@ class SDFORMAT_VISIBLE ParserConfig
   /// \return The unrecognized elements policy enum value
   public: EnforcementPolicy UnrecognizedElementsPolicy() const;
 
+  /// \brief Set the policy for deprecated elements.
+  /// \param[in] _policy The deprecated elements enforcement policy
+  public: void SetDeprecatedElementsPolicy(EnforcementPolicy _policy);
+
+  /// \brief Resets the policy for deprecated elements so that it follows
+  /// WarningsPolicy.
+  public: void ResetDeprecatedElementsPolicy();
+
+  /// \brief Get the current deperacted elements policy. By default, the policy
+  /// is the same as the overall WarningsPolicy, but it can be overriden by
+  /// SetDeprecatedElementsPolicy. Once it is overriden, changing
+  /// SetWarningsPolicy will not change the value of DeprecatedElementsPolicy
+  /// unless ResetDeprecatedElementsPolicy is called.
+  /// \return The deperacted elements policy enum value
+  public: EnforcementPolicy DeprecatedElementsPolicy() const;
+
   /// \brief Registers a custom model parser.
   /// \param[in] _modelParser Callback as described in
   /// sdf/InterfaceElements.hh.
--- a/test/integration/deprecated_specs.cc
+++ b/test/integration/deprecated_specs.cc
@@ -78,6 +78,11 @@ TEST(DeprecatedElements, CanEmitWarningWithErrorEnforcmentPolicy)
 
 #ifdef _WIN32
   sdf::Console::Instance()->SetQuiet(false);
+  sdf::testing::ScopeExit revertSetQuiet(
+      []
+      {
+        sdf::Console::Instance()->SetQuiet(true);
+      });
 #endif
 
   sdf::SDFPtr sdf(new sdf::SDF());
@@ -123,7 +128,4 @@ TEST(DeprecatedElements, CanEmitWarningWithErrorEnforcmentPolicy)
     ASSERT_FALSE(errors.empty());
     EXPECT_EQ(sdf::ErrorCode::ELEMENT_DEPRECATED, errors[0].Code());
   }
-#ifdef _WIN32
-  sdf::Console::Instance()->SetQuiet(true);
-#endif
 }
--- a/test/integration/deprecated_specs.cc
+++ b/test/integration/deprecated_specs.cc
@@ -63,6 +63,11 @@ TEST(DeprecatedElements, CanEmitErrors)
   EXPECT_EQ(sdf::ErrorCode::ELEMENT_DEPRECATED, errors[0].Code());
 }
 
+bool contains(const std::string &_a, const std::string &_b)
+{
+  return _a.find(_b) != std::string::npos;
+}
+
 ////////////////////////////////////////////////////
 TEST(DeprecatedElements, CanEmitWarningWithErrorEnforcmentPolicy)
 {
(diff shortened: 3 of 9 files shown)


include/sdf/ParserConfig.hh
```diff
@@ -144,6 +144,22 @@ class SDFORMAT_VISIBLE ParserConfig
...
 
+  /// \brief Set the policy for deprecated elements.
+  /// \param[in] _policy The deprecated elements enforcement policy
+  public: void SetDeprecatedElementsPolicy(EnforcementPolicy _policy);
+
+  /// \brief Resets the policy for deprecated elements so that it follows
+  /// WarningsPolicy.
+  public: void ResetDeprecatedElementsPolicy();
+
+  /// \brief Get the current deperacted elements policy. By default, the policy
```
aaronchongth, line 155:
```suggestion
  /// \brief Get the current deprecated elements policy. By default, the policy
```

aaronchongth (approved):
Looks great! This would be super useful moving forwards.

include/sdf/Element.hh
```diff
@@ -101,6 +101,7 @@ namespace sdf
 
     /// \brief Set the requirement type.
     /// \param[in] _req Requirement type for this element:
+    /// -1: Deprecated.
```
EricCousineau-TRI, line 104: BTW This smells funky. I understand motivation is to have easy translation btw schema / XSD and here, but may be better as an enum at some point. Lemme know if an issue is useful.

test/integration/deprecated_specs.cc
```diff
@@ -38,3 +42,89 @@ TEST(DeprecatedSpecs, Spec1_2)
...
+
+////////////////////////////////////////////////////
+TEST(DeprecatedElements, CanEmitWarningWithErrorEnforcmentPolicy)
+{
+  // Capture sdfwarn output, which is treamed to std::cerr
+  std::stringstream buffer;
+  auto old = std::cerr.rdbuf(buffer.rdbuf());
+
+#ifdef _WIN32
+  sdf::Console::Instance()->SetQuiet(false);
```
EricCousineau-TRI, line 73: nit Ew... this seems nigh defective in terms of public contract.

I don't see any indicator in `Console.hh` that indicates this would be necessary?

test/integration/deprecated_specs.cc
```diff
@@ -38,3 +42,89 @@ TEST(DeprecatedSpecs, Spec1_2)
...
+#endif
+
+  auto contains = [](const std::string &_a, const std::string &_b)
```
EricCousineau-TRI, line 76: nit Benefit as lambda not immediately clearer.
Can you comment as to why? (e.g. "`EXPECT_PRED2` does not functions, so we define as lambda"?)
Also, is my speculation true? (if not, can you just make it a function?)

test/integration/deprecated_specs.cc
```diff
@@ -38,3 +42,89 @@ TEST(DeprecatedSpecs, Spec1_2)
...
+  EXPECT_TRUE(sdf::readString(
+      "<sdf version='1.8'><testElem/></sdf>", config, sdf, errors));
+  ASSERT_FALSE(errors.empty());
+  EXPECT_EQ(sdf::ErrorCode::ELEMENT_DEPRECATED, errors[0].Code());
+}
+
+////////////////////////////////////////////////////
+TEST(DeprecatedElements, CanEmitWarningWithErrorEnforcmentPolicy)
+{
+  // Capture sdfwarn output, which is treamed to std::cerr
```
EricCousineau-TRI, line 68: nit Also ew... it seems odd that the only way to intercept logging is via C++ I/O redirect on `cerr`.

Can't `Console` admit a more intuitive redirect as public API? (via `msgStream` or `logFileStream`?)
(lemme know if issue is better here)

test/integration/deprecated_specs.cc
```diff
@@ -38,3 +42,89 @@ TEST(DeprecatedSpecs, Spec1_2)
...
+    config.ResetDeprecatedElementsPolicy();
+    sdf::Errors errors;
+    EXPECT_TRUE(sdf::readString(
+          "<sdf version='1.8'><testElem/></sdf>", config, sdf, errors));
+    ASSERT_FALSE(errors.empty());
+    EXPECT_EQ(sdf::ErrorCode::ELEMENT_DEPRECATED, errors[0].Code());
+  }
+
+  // Revert cerr rdbug so as to not interfere with other tests
+  std::cerr.rdbuf(old);
```
EricCousineau-TRI, line 126: nit This may not be reachable in test failure which interrupts execution (usage of `ASSERT_*`).

Consider a C++ RAII pattern. Example from Drake:
https://github.com/RobotLocomotion/drake/blob/v0.29.0/common/scope_exit.h

EricCousineau-TRI (approved):
LGTM! Just some minor nits, not acutely aligned w/ aim of this PR, so feel free to ignore 'em!

include/sdf/Element.hh
azeey: Yeah, an enum would be good. We have a lot of places in the code where we're doing string comparison. But, I think that's a bigger change than what I'd like for this PR. An issue would be great!

test/integration/deprecated_specs.cc
azeey: I don't know if tests still fail if we don't disable the output on windows, but that's what the todo says: https://github.com/osrf/sdformat/blob/25004f0bad227cf11854cc208e74be08e8623359/src/Console.cc#L35-L41

test/integration/deprecated_specs.cc
azeey: Just being lazy. Changed to function in https://github.com/osrf/sdformat/pull/543/commits/9592c002b72c1e9d187b4487857e37b92b1d0aec

test/integration/deprecated_specs.cc
azeey: Added a way redirect `Console` in https://github.com/osrf/sdformat/pull/543/commits/c6b9843f2bf067ed404c720e7130040fb945ee79. Let me know if the API makes sense.

test/integration/deprecated_specs.cc
azeey: Added `ScopeExit` in https://github.com/osrf/sdformat/pull/543/commits/eebf49b92f85392d042044151aeb7206c18cf079.

azeey:
I believe I've addressed all the feedback. @EricCousineau-TRI , PTAL.

test/integration/deprecated_specs.cc
EricCousineau-TRI: Sweet!

test/integration/deprecated_specs.cc
EricCousineau-TRI: Yup, awesome!

test/integration/deprecated_specs.cc
EricCousineau-TRI: Excellent!

include/sdf/Element.hh
EricCousineau-TRI: Done! https://github.com/osrf/sdformat/issues/553

test/integration/deprecated_specs.cc
EricCousineau-TRI: Added issue: https://github.com/osrf/sdformat/issues/554

EricCousineau-TRI:
All of this looks great, thanks!

PR #575, opened 2021-05-19
Making PrintValues() and ToString() able to not print default elements

### LINKED ISSUE #287: Permit detecting an element / attribute was explicitly set (user) or implicitly set (conversion / default value)
From discussion on 2020-01-23 on Asana.

Notes:
Default values
* Know if a property was actually defined
* `libsdformat`, in its conversion phase, substitutes in default values
* Odd / non-strict value parsing (e.g. colors, RGBA)
* API: Could add an attribute check, `IsExplicitlySet()` or `UsedDefaultValue()` or something, and have overloads for querying that uses `optional<>` or returns a flag
* May be hard to check if a default value was actually set?
* Could be useful for deprecating attributes that a user set?

\cc @SeanCurtis-TRI

marcoag:
Related to #287

## Summary
Now PrintValues() and ToString() will not output the elements that were added by default during parsing. There's also the option to call them with a boolean flag that will allow the caller to decide weather they want defaults int the output or not.

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -112,7 +112,10 @@ namespace sdf
     /// \sa Element::SetRequired
     public: const std::string &GetRequired() const;
 
-    /// \brief Set if the element and children wer set or default in the file
+    /// \brief Set if the element and children where set or default
+    /// in the original file. This is meant to be used by the parser to
+    /// indicate whether the element was set by the user in the original
+    /// file or added by default during parsing.
     /// \param[in] _value True if the element was set
     public: void SetExplicitlySetInFile(const bool _value);
 
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -152,8 +152,8 @@ namespace sdf
     /// \brief Output Element's values to stdout.
     /// \param[in] _prefix String value to prefix to the output.
     /// \param[in] _includeDefaults flag for default elements and attributes.
-    public: void PrintValues(std::string _prefix,
-                             const bool _includeDefaults) const;
+    public: void PrintValues(const std::string &_prefix,
+                             bool _includeDefaults) const;
 
     /// \brief Helper function for SDF::PrintDoc
     ///
@@ -182,7 +182,7 @@ namespace sdf
     /// \param[in] _includeDefaults flag for default elements and attributes.
     /// \return The string representation.
     public: std::string ToString(const std::string &_prefix,
-                                 const bool _includeDefaults) const;
+                                 bool _includeDefaults) const;
 
     /// \brief Add an attribute value.
     /// \param[in] _key Key value.
@@ -438,7 +438,7 @@ namespace sdf
     /// \param[out] _out the std::ostreamstream to write output to.
     private: void ToString(const std::string &_prefix,
                            std::ostringstream &_out,
-                           const bool _includeDefaults) const;
+                           bool _includeDefaults) const;
 
     /// \brief Generate a string (XML) representation of this object.
     /// \param[in] _prefix arbitrary prefix to put on the string.
@@ -446,7 +446,7 @@ namespace sdf
     /// \param[out] _out the std::ostreamstream to write output to.
     private: void PrintValuesImpl(const std::string &_prefix,
                                   std::ostringstream &_out,
-                                  const bool _includeDefaults) const;
+                                  bool _includeDefaults) const;
 
     /// \brief Create a new Param object and return it.
     /// \param[in] _key Key for the parameter.
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -112,13 +112,13 @@ namespace sdf
     /// \sa Element::SetRequired
     public: const std::string &GetRequired() const;
 
-    /// \brief Set if the element and it's child Elments are set or default
-    /// \param[in] _value True if the element is set
-    public: void SetIsExplicitlySet(const bool _value);
+    /// \brief Set if the element and children wer set or default in the file
+    /// \param[in] _value True if the element was set
+    public: void SetExplicitlySetInFile(const bool _value);
 
-    /// \brief Return true if the element has been explicitly set
-    /// \return True if the element has been set
-    public: bool GetIsExplicitlySet() const;
+    /// \brief Return if the element was been explicitly set in the file
+    /// \return True if the element was set in the file
+    public: bool GetExplicitlySetInFile() const;
 
     /// \brief Set whether this element should copy its child elements
     /// during parsing.
@@ -458,8 +458,8 @@ namespace sdf
     /// \brief True if element is required
     public: std::string required;
 
-    /// \brief True if the element is set.
-    public: bool isExplicitlySet;
+    /// \brief True if the element was set in the SDF file.
+    public: bool explicitlySetInFile;
 
     /// \brief Element description
     public: std::string description;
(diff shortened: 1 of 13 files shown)


include/sdf/Element.hh
```diff
@@ -149,6 +149,12 @@ namespace sdf
     /// \param[in] _prefix String value to prefix to the output.
     public: void PrintValues(std::string _prefix) const;
 
+    /// \brief Output Element's values to stdout.
+    /// \param[in] _prefix String value to prefix to the output.
+    /// \param[in] _includeDefaults flag for default elements and attributes.
+    public: void PrintValues(std::string _prefix,
```
aaronchongth, line 155: We should pass in the `std::string` as a `const std::string &`

include/sdf/Element.hh
```diff
@@ -149,6 +149,12 @@ namespace sdf
...
+                             const bool _includeDefaults) const;
```
aaronchongth, line 156: We won't need `const` when we are passing the parameter by value.

include/sdf/Element.hh
```diff
@@ -171,6 +177,13 @@ namespace sdf
     /// \return The string representation.
     public: std::string ToString(const std::string &_prefix) const;
 
+    /// \brief Convert the element values to a string representation.
+    /// \param[in] _prefix String value to prefix to the output.
+    /// \param[in] _includeDefaults flag for default elements and attributes.
+    /// \return The string representation.
+    public: std::string ToString(const std::string &_prefix,
+                                 const bool _includeDefaults) const;
```
aaronchongth, line 185: We won't need `const` when we are pssing by value.

include/sdf/Element.hh
```diff
@@ -424,13 +437,16 @@ namespace sdf
     /// \param[in] _prefix arbitrary prefix to put on the string.
     /// \param[out] _out the std::ostreamstream to write output to.
     private: void ToString(const std::string &_prefix,
-                           std::ostringstream &_out) const;
+                           std::ostringstream &_out,
+                           const bool _includeDefaults) const;
```
aaronchongth, line 441: We won't need `const` when we are pssing by value.

include/sdf/Element.hh
```diff
@@ -424,13 +437,16 @@ namespace sdf
...
 
     /// \brief Generate a string (XML) representation of this object.
     /// \param[in] _prefix arbitrary prefix to put on the string.
+    /// \param[in] _printDefaults flag for default elements and attributes.
     /// \param[out] _out the std::ostreamstream to write output to.
     private: void PrintValuesImpl(const std::string &_prefix,
-                                  std::ostringstream &_out) const;
+                                  std::ostringstream &_out,
+                                  const bool _includeDefaults) const;
```
aaronchongth, line 449: We won't need `const` when we are pssing by value.

aaronchongth (changes requested):
Just did a quick look through and requested some changes to the API functions. Looks great! Are all the bullet points in #287 addressed at this point, such that it closes it?

include/sdf/Element.hh
marcoag: addressed on https://github.com/osrf/sdformat/pull/575/commits/c9780654c9ba09a09d250b5776141a0dcd428b80

marcoag:
I thought we got to that point, but maybe @azeey should confirm...

azeey:
> I thought we got to that point, but maybe @azeey should confirm...

I think #287 only needed #542, but I remember @marcoag mentioning there were still some issues when upconverting SDFormat files. If that's the case, we should probably keep it open until that's resolved.

aaronchongth (approved):
This looks awesome! Thanks for making the changes!

azeey:
Per VC, we should keep the default behavior (when printing via `ign sdf -p` or using `sdf::Element::ToString()` of printing values and elements added by default for `libsdformat9-libsdformat11`. For `libsdformat12`, we'll change the default behavior to print only user added content.

marcoag:
Now it's a bit of a weird behavior since it will print any default `Element` but not their default `Attributes`. Shall we add another flag to print or not attributes in a similar way to `Element` here?

azeey:
I suggest we have two behaviors: (1) a default behavior that is consistent with what the code did before this PR, (2) a new behavior enabled by a flag that does **not** print default elements or attributes.

marcoag:
As agreed on our previous VC this has been added: 
- [x] kept behavior consistent with the code before this PR (print/toString default elements and not print/toString default attributes). 
- [x] Added two flags: one to print/toString default elements and another one to print/toString default attributes

src/Element.cc
```diff
@@ -442,47 +442,59 @@ void Element::PrintDocLeftPane(std::string &_html, int _spacing,
...
+         aiter != this->dataPtr->attributes.end(); ++aiter)
     {
-      (*eiter)->ToString(_prefix + "  ", _out);
+      // Only print attribute values if they were set
+      // TODO(anyone): GetRequired is added here to support up-conversions where
+      // a new required attribute with a default value is added. We would have
+      // better separation of concerns if the conversion process set the
+      // required attributes with their default values.
+      if ((*aiter)->GetSet() || (*aiter)->GetRequired() ||
+          _includeDefaultAttributes)
```
marcoag, line 463: Still unsure if we should do something about this comment here...

azeey (approved):
LGTM! Just a few minor comments.

src/Element_TEST.cc
```diff
@@ -575,6 +574,71 @@ TEST(Element, ToStringInclude)
...
+  sdf::ElementPtr elem = std::make_shared<sdf::Element>();
+  elem->SetParent(parent);
+  parent->InsertElement(elem);
+  sdf::ElementPtr elem2 = std::make_shared<sdf::Element>();
+  elem2->SetParent(parent);
+  parent->InsertElement(elem2);
+
+  std::ostringstream stream;
+  stream
+    << "<>\n"
```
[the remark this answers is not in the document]
marcoag: Solved in https://github.com/osrf/sdformat/pull/575/commits/f90794ae9cbfb8d2aff768e7b01bd8ca3d541c37#.

src/Element_TEST.cc
```diff
@@ -575,6 +574,71 @@ TEST(Element, ToStringInclude)
...
+{
+  sdf::ElementPtr element = std::make_shared<sdf::Element>();
+  element->AddAttribute("test", "string", "foo", false, "foo description");
+  element->AddAttribute("test2", "string", "bar", true, "bar description");
+
+  EXPECT_EQ(element->ToString(""), element->ToString("", true, false));
+  EXPECT_EQ(element->ToString(""), element->ToString("", false, false));
+
+  std::ostringstream stream;
+  stream << "< test2='bar'/>\n";
```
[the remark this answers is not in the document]
marcoag: Solved in https://github.com/osrf/sdformat/pull/575/commits/f90794ae9cbfb8d2aff768e7b01bd8ca3d541c37#.

include/sdf/Element.hh
```diff
@@ -171,6 +179,15 @@ namespace sdf
     /// \return The string representation.
     public: std::string ToString(const std::string &_prefix) const;
 
+    /// \brief Convert the element values to a string representation.
```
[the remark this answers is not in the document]
marcoag: Added in https://github.com/osrf/sdformat/pull/575/commits/f90794ae9cbfb8d2aff768e7b01bd8ca3d541c37#.

marcoag:
As per VC we agreed to not consider default elements those added when the auto-conversion to the latest sdf version happens.

PR #399, opened 2020-10-29
Prefix nested model names when flattening

scpeters:
Currently the [addNestedModel function in parser.cc](https://github.com/osrf/sdformat/blob/sdformat10_10.0.0/src/parser.cc#L1253) prefixes link, joint, and frame names with the flattened model name delimited by "::". This applies the same prefix to the names of nested models within the flattened model as well. Motivated by https://github.com/osrf/sdformat/pull/355#pullrequestreview-485492932

PR #596, opened 2021-06-16
Translate poses of nested models inside other nested models (sdf6)

j-rivero:
When `addNestedModel` function is called, it processes links and joints to translate the pose accordingly to parent pose. This was not done for nested models inside the SDF being processed by the function, hence they were not aware of parent pose.

This PR includes nested models in the same way that is doing for links and add a test that fails without the change.
The change was proposed by @azeey. If we are not wrong, the change is not present in newer versions of sdformat after the implementation of frame semantics.

### DIFF
--- a/test/integration/model/box_with_submodel/model.config
+++ b/test/integration/model/box_with_submodel/model.config
@@ -1,0 +1,15 @@
+<?xml version="1.0"?>
+<model>
+  <name>BoxWithSubmodel</name>
+  <version>1.0</version>
+  <sdf version="1.5">model.sdf</sdf>
+
+  <author>
+    <name>Jose Luis Rivero</name>
+    <email>jrivero@osrfoundation.org</email>
+  </author>
+
+  <description>
+    A box with a nested model
+  </description>
+</model>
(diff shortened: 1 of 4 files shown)


scpeters (approved):
this looks good to me

I tried running this test on the `sdf9` branch, and it fails since it looks only at the `<pose>` values without considering the `//pose/@relative_to` attribute. I believe `sdf9` is doing the right thing though, and that the test would pass if the pose values are calculated with the `SemanticPose` API

I believe an equivalent fix was merged to `sdf10` in  #399 and then back-ported to `sdf9` in https://github.com/osrf/sdformat/pull/424

test/integration/model/box_with_submodel/model.config
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<model>
+  <name>BoxWithSubmodel</name>
+  <version>1.0</version>
+  <sdf version="1.5">model.sdf</sdf>
```
azeey, line 5: nit: Should match the version in `model.sdf`
```suggestion
  <sdf version="1.6">model.sdf</sdf>
```

azeey (approved):
Looks good to me too! We should merge this along with #597 so it doesn't break models with joints that reference links in nested models.

osrf-triage:

PR #597, opened 2021-06-21
Fix flattening logic for nested model names (sdf6)

### LINKED ISSUE #574: Error in flattening logic for nested model names
## Environment
* OS Version: Ubuntu Bionic
* Binaries: SDFormat 9.5.1 , Gazebo 11.5.1

## Description

In the test case provided in https://github.com/j-rivero/sdf_bug, Gazebo11 is complaining with:

```
[Err] [Joint.cc:297] EXCEPTION: Couldn't Find Child Link[sub_door::door]
``` 

when using an SDF similar to:

```
<?xml version="1.0" ?>
<sdf version="1.5">
  <model name='hinged_door_ticket'>
      <static>0</static>
      <link name='frame'>
      </link>
      <model name='sub_door'>
	<static>0</static>
	<link name='door'>
	  <pose frame=''>-0 0 2.04 3.14159 -0 1.58</pose>	   
	</link>
      </model>
      <joint name='hinge' type='revolute'>
        <parent>frame</parent>
        <child>sub_door::door</child>       
      </joint>
    </model>
</sdf>
```

The great @azeey first impression was: "It looks like the breaking change is that we added a prefix to nested models when flattening (https://github.com/osrf/sdformat/pull/399). Running `gz sdf -p frame.world`  shows that the joint is looking for sub_door::door. In Gazebo7, there is a model called sub_door. In Gazebo11, that model is now hinged_door_frame::sub_door. For some reason, the flattening logic is not handling that properly

* Expected behavior: the name got resolved correctly
* Actual behavior: failing to resolve the name, error message is emitted

## Steps to reproduce

1. Download the example repository in the description
2. Execute `GAZEBO_MODEL_PATH=${GAZEBO_MODEL_PATH}:${PWD} gazebo --verbose frame.world`

j-rivero:
# 🦟 Bug fix

Fixes #574

## Summary
I think that the current flattening logic deal with names using  `<`, `>`, `\` or `'` as separators. That leaves out some cases were the user use composed names (component::element) since the separator here is `::`. I added some logic to deal with use cases were element is present in different situations with other components of the name:

 * case A: `::component1::element::component2`
 * case B. `>element::component`
 * case C. `::component::element<`

## Checklist
- [x] Added tests

### DIFF
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -1117,6 +1117,35 @@ void addNestedModel(ElementPtr _sdf, ElementPtr _includeSDF)
                      "'" + name_with_nested_prefix + "'");
     replace_all(str, ">" + old_name + "<",
                      ">" + name_with_nested_prefix + "<");
+    // Deal with nested model inside other nested model and already with
+    // '::component::' entries in the name
+    // Case A. ::component1::iter.first::component2
+    replace_all(str, "::" + old_name + "::",
+                     "::" + name_with_nested_prefix + "::");
+    // Case B. >iter.first::component
+    replace_all(str, ">" + old_name + "::",
+                     ">" + name_with_nested_prefix + "::");
+    // Case C.::component::iter.first<
+    replace_all(str, "::" + old_name + "<",
+                     "::" + name_with_nested_prefix + "<");
+    // remove duplicate prefixes introduced by the logic above (cases
+    // A,B,C) if the full name (iter.second) was used in the original SDF: (i.e
+    // iter.second::component)
+    std::string::size_type found = name_with_nested_prefix.find(old_name);
+    if (found != std::string::npos)
+    {
+      std::string nested_prefix = name_with_nested_prefix;
+      nested_prefix.erase(found, old_name.length());
+      replace_all(str, nested_prefix + nested_prefix, nested_prefix);
+    }
+    else
+    {
+      // should never happen since name_with_nested_prefix =
+      // nested model name + old_name
+      sdferr << "Nested name '" << name_with_nested_prefix <<
+                "' does not contain the plain name of the element '" <<
+                old_name << "'. Please report an issue. \n";
+    }
   }
 
   _includeSDF->ClearElements();
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -1111,23 +1111,23 @@ void addNestedModel(ElementPtr _sdf, ElementPtr _includeSDF)
   {
     std::string old_name(iter->first);
     std::string name_with_nested_prefix(iter->second);
-    replace_all(str, "\"" + old_name + "\"",
-                     "\"" + name_with_nested_prefix + "\"");
-    replace_all(str, "'" + old_name + "'",
-                     "'" + name_with_nested_prefix + "'");
-    replace_all(str, ">" + old_name + "<",
-                     ">" + name_with_nested_prefix + "<");
+    replace_all(str, std::string("\"") + old_name + "\"",
+                     std::string("\"") + name_with_nested_prefix + "\"");
+    replace_all(str, std::string("'") + old_name + "'",
+                     std::string("'") + name_with_nested_prefix + "'");
+    replace_all(str, std::string(">") + old_name + "<",
+                     std::string(">") + name_with_nested_prefix + "<");
     // Deal with nested model inside other nested model and already with
-    // '::component::' entries in the name
-    // Case A. ::component1::iter.first::component2
-    replace_all(str, "::" + old_name + "::",
-                     "::" + name_with_nested_prefix + "::");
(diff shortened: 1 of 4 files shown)


src/parser.cc
```diff
@@ -1074,11 +1074,12 @@ void addNestedModel(ElementPtr _sdf, ElementPtr _includeSDF)
   std::string modelName = modelPtr->Get<std::string>("name");
   while (elem)
   {
+    std::string elemName = elem->Get<std::string>("name");
```
azeey, line 1077: This will rename all elements with the `name` attribute. That might break things (eg. `<plugin>` tags), so I suggest we limit it to `link`, `joint`, and `model`.

src/parser.cc
```diff
@@ -1113,12 +1109,43 @@ void addNestedModel(ElementPtr _sdf, ElementPtr _includeSDF)
   for (std::map<std::string, std::string>::iterator iter = replace.begin();
        iter != replace.end(); ++iter)
   {
-    replace_all(str, std::string("\"") + iter->first + "\"",
-                std::string("\"") + iter->second + "\"");
-    replace_all(str, std::string("'") + iter->first + "'",
-                std::string("'") + iter->second + "'");
-    replace_all(str, std::string(">") + iter->first + "<",
-                std::string(">") + iter->second + "<");
+    std::string old_name(iter->first);
```
azeey, line 1112: nit: `camelCase` is the preferred style in libsdformat for variable names.

azeey (commented):
I think we only need to handle case `B`. 

Case `A` handles a name that starts with `::` and I don't think that's valid in `SDFormat`. 

Case `C` handles a name where the old name is preceded by some other name, `some_name::old_name`, and it renames it to `some_name::new_prefix::old_name`. But for `some_name::old_name` to exist, there must be a nested model called `some_name` and a child element of that nested model called `old_name`:

```xml
<model name='included_model'>
  <link name='L1'/>
  <model name='some_name'>
     <link name='old_name'/>
  </model>
  <joint name='J1' type='fixed'>
      <parent>some_name::old_name</parent>
      <child>L1</child>
   </joint>
</model>
```
But if this was the case, case B would have already renamed `some_name` to `new_prefix::some_name` and so we would want `<parent>some_name::old_name</parent>` to be renamed to `<parent>new_prefix::some_name::old_name</parent>` **not**`<parent>some_name::new_prefix::old_name</parent>`

j-rivero:
> I think we only need to handle case `B`.

I think you are right, changed in 4b97716

src/parser.cc
j-rivero: changed 38da8ab

src/parser.cc
j-rivero: good catch, added 579da3b

src/parser.cc
```diff
@@ -1113,12 +1115,36 @@ void addNestedModel(ElementPtr _sdf, ElementPtr _includeSDF)
...
+                     std::string(">") + nameWithNestedPrefix + "::");
+    // remove duplicate prefixes introduced by the logic above
+    // if the full name (iter.second) was used in the original SDF: (i.e
+    // iter.second::component)
+    std::string::size_type found = nameWithNestedPrefix.find(oldName);
+    if (found != std::string::npos)
+    {
+      std::string nestedPrefix = nameWithNestedPrefix;
+      nestedPrefix.erase(found, oldName.length());
+      replace_all(str, nestedPrefix + nestedPrefix, nestedPrefix);
```
azeey, line 1138: Is this still needed after removing some of the `replace_all` cases?

src/parser.cc
```diff
@@ -1074,11 +1074,18 @@ void addNestedModel(ElementPtr _sdf, ElementPtr _includeSDF)
...
+    // i.e: plugin
+    if ((elem->GetName() == "link") || (elem->GetName() == "joint") ||
+        (elem->GetName() == "model"))
     {
       std::string elemName = elem->Get<std::string>("name");
       std::string newName =  modelName + "::" + elemName;
       replace[elemName] = newName;
+    }
+
+    if (elem->GetName() == "link")
```
azeey, line 1087: Quick note that this will be in conflict with #596, but we'll definitely need the line from there.

azeey (approved):

src/parser.cc
j-rivero: Umm I think you are right, this is not needed anymore after reducing the cases to just one. Removed 0f7ccf8

src/parser.cc
j-rivero: +1 going to merge both, I'll keep an eye on it, tests should fail we miss it.

scpeters:
should this be merged forward to sdf9?

osrf-triage:

osrf-triage:

PR #601, opened 2021-06-25
Add World::ValidateGraphs method

scpeters:
# 🎉 New feature

I noticed a bug in ign-physics that needs a new API in libsdformat to properly fix it.

## Summary

I noticed a test in ign-physics in which some models do not have properly loaded graphs, but it's hard to detect since the graph data is private. This creates a `World::ValidateGraphs()` method to call the hidden `validate*Graph` functions in FrameSemantics.cc  with the private graph data. It also requires an extra check in the `validate*Graph` functions that the scope points to a valid graph.

## Test it

Run the `UNIT_World_TEST` and `INTEGRATION_world_dom` tests and see that they pass.

## Checklist
- [X] Added tests
to support the maintainers


chapulina (commented):
I just have the same request as in #602 for more testing.

include/sdf/World.hh
```diff
@@ -70,6 +70,12 @@ namespace sdf
     /// an error code and message. An empty vector indicates no error.
     public: Errors Load(sdf::ElementPtr _sdf, const ParserConfig &_config);
 
+    /// \brief Check that the FrameAttachedToGraph and PoseRelativeToGraph
+    /// are valid.
+    /// \return Errors, which is a vector of Error objects. Each Error includes
+    /// an error code and message. An empty vector indicates no error.
+    public: Errors ValidateGraphs() const;
```
azeey, line 77: Should we add a similar function to `sdf::Model` too?

scpeters: good idea: b32275c

azeey (approved):

PR #605, opened 2021-06-28
Fix ABI break

azeey:
# 🦟 Bug fix

## Summary

This moves the recently added data member `explicitlySetInFile` to the
end of the `ElementPrivate` class to an ABI breakage (See https://github.com/ignitionrobotics/ign-gazebo/issues/891).

We are using the Pimpl pattern with the `Element` class to avoid ABI issues, however, we have template functions that use data members from the `ElementPrivate` class in `Element.hh`. These templates can get instantiated in a translation unit outside of the ones used to build `libsdformat` and the offset values for the data members be computed based on the order of the members at the time the translation unit is compiled. When we later change the order of the data members, the offsets in the translation units will become invalid causing a crash as seen in https://github.com/ignitionrobotics/ign-gazebo/issues/891

## Checklist
- [ ] Added tests
- [x] All tests passed (See
to support the maintainers

### DIFF
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -489,9 +489,6 @@ namespace sdf
     /// \brief True if element is required
     public: std::string required;
 
-    /// \brief True if the element was set in the SDF file.
-    public: bool explicitlySetInFile;
-
     /// \brief Element description
     public: std::string description;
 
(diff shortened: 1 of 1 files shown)


adlarkin (approved):

chapulina (approved):

azeey:

chapulina:
That variable was added in #542, which hasn't been released into `sdf9` yet, so Citadel users shouldn't notice this happened.

https://github.com/ignitionrobotics/sdformat/compare/sdformat9_9.5.0...sdf9

scpeters (approved):

chapulina:
@osrf-jenkins run tests you can

chapulina:
I enabled the DCO bot after this PR was open. Let's merge without that check for now, reopening the PR would trigger lots of builds.

PR #620, opened 2021-07-09
Add Element::FindElement as an alternative to Element::GetElement

### LINKED ISSUE #188: Const accessors in sdformat.
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/188)) by Alejandro Castro (Bitbucket: [amcastro0622](https://bitbucket.org/%7Bf78f4af9-eed7-49c4-b222-5b3dc10d5093%7D/)).**

We had a [use case in our Drake code base](https://reviewable.io/reviews/robotlocomotion/drake/8828#-LD2bp_9--hRPdZXTAQL) in which we only needed const access to an `sdf::Element`. Right now the current API only offers `sdf::Element::GetElement()` which returns a mutable shared pointer to a mutable element object.
Ideally we'd also like having `const Element& GetElement(const std::string&)`.
Most likely similar const access would be needed in other places.

azeey:
# 🎉 New feature

Closes #188

## Summary
`Element::GetElement` may create a child element if it doesn't exist.
This makes it a non-const function. Instead of changing this behavior, a
new function `Element::FindElement` is added that returns a `nullptr` if
the child element is not found instead of creating a new element.

## Test it
Run test `UNIT_Element_TEST`.

## Checklist
- [x] Added tests

### DIFF
--- a/src/Element_TEST.cc
+++ b/src/Element_TEST.cc
@@ -827,7 +827,7 @@ TEST(Element, GetNextElementMultiple)
 
 /////////////////////////////////////////////////
 /// Helper function to add child elements without having to create descriptions
-sdf::ElementPtr AddChildElement(sdf::ElementPtr _parent,
+sdf::ElementPtr addChildElement(sdf::ElementPtr _parent,
                                 const std::string &_elementName,
                                 const bool _addNameAttribute,
                                 const std::string &_childName)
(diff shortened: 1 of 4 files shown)


src/Element_TEST.cc
```diff
@@ -825,6 +825,25 @@ TEST(Element, GetNextElementMultiple)
   ASSERT_EQ(child2->GetNextElement(""), nullptr);
 }
 
+/////////////////////////////////////////////////
+/// Helper function to add child elements without having to create descriptions
+sdf::ElementPtr AddChildElement(sdf::ElementPtr _parent,
```
chapulina, line 830: Nit: we usually use `camelCase` for free functions

chapulina (approved):
Omg, this is great!

src/Element_TEST.cc
azeey: Updated in https://github.com/ignitionrobotics/sdformat/pull/620/commits/d51930495c7946659fd5f74131faf613099b4d05. I'm never sure what to do about free functions and static functions because I think we diverge from the [google style guide](https://google.github.io/styleguide/cppguide.html#Function_Names) on this.

chapulina: Yeah we don't have this documented anywhere, it's one of those things that we need to document

PR #628, opened 2021-07-21
Fix urdf link extension tags

scpeters:
# 🦟 Bug fix

Fixes `<gravity/>` and `<velocity_decay/>` tags in URDF `<gazebo/>` extension elements

## Summary

Currently the `parser_urdf` code inserts empty `<velocity_decay>` elements and `<gravity>1</gravity>` elements to links when parsing `<gazebo/>` extensions. I noticed this because I was unable to set `<gravity>0</gravity>` from a `<gazebo/>` tag, since it was being inserted after and only the first one was being read.

This fixes the logic and adds some tests to confirm that these extensions can be properly set.

## Checklist
- [X] Added tests

### DIFF
--- a/src/parser_urdf.cc
+++ b/src/parser_urdf.cc
@@ -1343,6 +1343,7 @@ void URDF2SDF::ParseSDFExtension(TiXmlDocument &_urdfXml)
       else if (childElem->ValueStr() == "turnGravityOff")
       {
         std::string valueStr = GetKeyValueAsString(childElem);
+        sdf->isGravity = true;
 
         // default of gravity is true
         if (lowerStr(valueStr) == "false" || lowerStr(valueStr) == "no" ||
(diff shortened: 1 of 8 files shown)


src/parser_urdf.cc
```diff
@@ -1343,6 +1343,7 @@ void URDF2SDF::ParseSDFExtension(TiXmlDocument &_urdfXml)
       else if (childElem->ValueStr() == "turnGravityOff")
```
jennuine, line 1343: To avoid introducing a new variable `isGravity`, another option could be adding another `if` statement after this one. For example, 

```C++
else if (childElem->ValueStr() == "gravity")
{
  std::string valueStr = GetKeyValueAsString(childElem);

  if (lowerStr(valueStr) == "false" || lowerStr(valueStr) == "no" ||
      valueStr == "0")
  {
    sdf->gravity = false;
  }
  else
  {
    sdf->gravity = true;
  }
}
```

jennuine (commented):
LGTM, one minor suggestion

azeey (approved):
LGTM!

src/parser_urdf.cc
scpeters: That is true; if we wanted to avoid adding a new variable, we could add some extra code instead. I personally lean towards less code, but I'm open to change this if there is consensus for that

jennuine: Less code sounds good to me :)

jennuine: Less code sounds good to me :)

jennuine (approved):

PR #638, opened 2021-07-28
Support parsing elements that are not part of the schema

### LINKED ISSUE #603: Support parsing "inf", "+inf", "-inf" for elements that are not part of the schema
## Desired behavior
Motivated by https://github.com/osrf/sdformat/issues/261#issuecomment-863585816, `sdf::Param::Get` should behave consistently when parsing a value to a given c++ type regardless of the `typeName` of the `sdf::Param`. Currently, `sdf::Param::Get<double>("inf")` returns the expected value if the `typeName` is `double` of `float`, but not if it's `string`.

## Implementation suggestion
In https://github.com/osrf/sdformat/blob/a797e18eeb012012f02f194178b16b5bf847fe39/include/sdf/Param.hh#L385-L394

we can try to refactor `sdf::Param::ValueFromString` and use it here if the value held by `this->dataPtr->value` is different from `T` instead of directly parsing via `std::stringstream`.

jennuine:
# 🦟 Bug fix

Fixes #603

## Summary

`sdf::Param::Get` parses a value to a given type regardless of the `typeName` of the `sdf::Param`. Previously when `typeName = "string"` then `sdf::Param::Get<double>("inf")` would return `0`, now it returns the expected value `inf`.

Test:
```bash
./build/sdformat9/src/UNIT_Param_TEST --gtest_filter=SetFromString.DoublePositiveInf
./build/sdformat9/src/UNIT_Param_TEST --gtest_filter=SetFromString.DoubleNegativeInf
```

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/Param.hh
+++ b/include/sdf/Param.hh
@@ -260,13 +260,14 @@ namespace sdf
     /// \brief This parameter's default value
     public: ParamVariant defaultValue;
 
-    /// \brief Method used to set the Element from a passed-in string
+    /// \brief Method used to set the Param from a passed-in string
     /// \param[in] _typeName The data type of the value to set
     /// \param[in] _valueStr The value as a string
     /// \param[out] _valueToSet The value to set
-    public: bool ValueFromStringImpl(const std::string &_typeName,
-                                     const std::string &_valueStr,
-                                     ParamVariant &_valueToSet);
+    public: bool SDFORMAT_VISIBLE ValueFromStringImpl(
+                                    const std::string &_typeName,
+                                    const std::string &_valueStr,
+                                    ParamVariant &_valueToSet) const;
   };
 
   ///////////////////////////////////////////////
--- a/include/sdf/Param.hh
+++ b/include/sdf/Param.hh
@@ -349,7 +349,9 @@ namespace sdf
   {
     T *value = std::get_if<T>(&this->dataPtr->value);
     if (value)
+    {
       _value = *value;
+    }
     else
     {
       std::string typeStr = this->dataPtr->TypeToString<T>();
@@ -364,7 +366,22 @@ namespace sdf
       bool success = this->dataPtr->ValueFromStringImpl(typeStr, valueStr, pv);
 
       if (success)
+      {
         _value = std::get<T>(pv);
+      }
+      else if (typeStr == "bool" && this->dataPtr->typeName == "string")
+      {
+        valueStr = lowercase(valueStr);
+
+        std::stringstream tmp;
+        if (valueStr == "true" || valueStr == "1")
+          tmp << "1";
+        else
+          tmp << "0";
+
+        tmp >> _value;
+        return true;
+      }
 
       return success;
     }
--- a/include/sdf/Param.hh
+++ b/include/sdf/Param.hh
@@ -264,11 +264,57 @@ namespace sdf
     /// \param[in] _typeName The data type of the value to set
     /// \param[in] _valueStr The value as a string
     /// \param[out] _valueToSet The value to set
+    /// \return True if the value was successfully set, false otherwise
     public: bool SDFORMAT_VISIBLE ValueFromStringImpl(
                                     const std::string &_typeName,
                                     const std::string &_valueStr,
                                     ParamVariant &_valueToSet) const;
+
+    /// \brief Data type to string mapping
+    /// \return The type as a string, empty string if unknown type
+    public: template<typename T>
+            std::string TypeToString() const;
   };
+
+  ///////////////////////////////////////////////
+  template<typename T>
+  std::string ParamPrivate::TypeToString() const
+  {
+    if constexpr (std::is_same_v<T, bool>)
+      return "bool";
+    else if constexpr (std::is_same_v<T, char>)
+      return "char";
+    else if constexpr (std::is_same_v<T, std::string>)
+      return "string";
+    else if constexpr (std::is_same_v<T, int>)
+      return "int";
+    else if constexpr (std::is_same_v<T, std::uint64_t>)
+      return "uint64_t";
+    else if constexpr (std::is_same_v<T, unsigned int>)
+      return "unsigned int";
+    else if constexpr (std::is_same_v<T, double>)
+      return "double";
+    else if constexpr (std::is_same_v<T, float>)
+      return "float";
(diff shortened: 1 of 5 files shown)


azeey (changes requested):
Looks great! I have some comments to help with the CI failure.

include/sdf/Param.hh
```diff
@@ -291,50 +300,62 @@ namespace sdf
...
+    if (value)
+      _value = *value;
+    else
     {
-      if (typeid(T) == typeid(bool) && this->dataPtr->typeName == "string")
+      std::string valueStr = this->GetAsString();
+      bool success = false;
+      ParamPrivate::ParamVariant pv;
+
+      if constexpr (std::is_same_v<T, bool>)
```
azeey, line 312: While reviewing https://github.com/ignitionrobotics/sdformat/pull/589/ I realized once this is forward ported, we'll have to update all the `ValueFromStringImpl` calls with an extra argument. Do you mind refactoring this function so that the C++ type to string mapping happens in a separate function and then we can have just one call to `ValueFromStringImpl` here?

include/sdf/Param.hh
```diff
@@ -258,6 +259,14 @@ namespace sdf
 
     /// \brief This parameter's default value
     public: ParamVariant defaultValue;
+
+    /// \brief Method used to set the Element from a passed-in string
```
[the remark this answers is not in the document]
jennuine: Used "Param" 6e4fe437972082d6e10aeda53c1d86b3a5a60396

src/Param.cc
```diff
@@ -273,14 +273,24 @@ std::string Param::GetDefaultAsString() const
...
 bool Param::ValueFromString(const std::string &_value)
+{
+  return this->dataPtr->ValueFromStringImpl(this->dataPtr->typeName,
+                                           _value,
+                                           this->dataPtr->value);
+}
+
+//////////////////////////////////////////////////
+bool ParamPrivate::ValueFromStringImpl(const std::string &_typeName,
+                                      const std::string &_valueStr,
```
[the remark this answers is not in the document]
jennuine: 6e4fe437972082d6e10aeda53c1d86b3a5a60396

include/sdf/Param.hh
```diff
@@ -258,6 +259,14 @@ namespace sdf
...
+    /// \param[in] _typeName The data type of the value to set
+    /// \param[in] _valueStr The value as a string
+    /// \param[out] _valueToSet The value to set
+    public: bool ValueFromStringImpl(const std::string &_typeName,
+                                     const std::string &_valueStr,
+                                     ParamVariant &_valueToSet);
```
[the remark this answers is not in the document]
jennuine: Thanks! 6e4fe437972082d6e10aeda53c1d86b3a5a60396

include/sdf/Param.hh
```diff
@@ -258,6 +259,14 @@ namespace sdf
 
     /// \brief This parameter's default value
     public: ParamVariant defaultValue;
+
+    /// \brief Method used to set the Element from a passed-in string
+    /// \param[in] _typeName The data type of the value to set
+    /// \param[in] _valueStr The value as a string
+    /// \param[out] _valueToSet The value to set
+    public: bool ValueFromStringImpl(const std::string &_typeName,
```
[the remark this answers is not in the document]
jennuine: 6e4fe437972082d6e10aeda53c1d86b3a5a60396

include/sdf/Param.hh
jennuine: f0f9759ec9d4572db9c5fd9e0dd861e6f679a7cf

src/Param_TEST.cc
```diff
@@ -75,11 +77,6 @@ TEST(Param, Bool)
   strParam.Get<bool>(value);
   EXPECT_TRUE(value);
 
-  // Anything other than 1 or true is treated as a false value
```
azeey, line 78: Is there a reason for removing this test?

src/Param.cc
```diff
@@ -304,140 +314,149 @@ bool Param::ValueFromString(const std::string &_value)
...
+             _typeName == "time")
     {
       StringStreamClassicLocale ss(tmp);
       sdf::Time timetmp;
 
       ss >> timetmp;
-      this->dataPtr->value = timetmp;
+      _valueToSet = timetmp;
+    }
+    else if (_typeName == "ignition::math::Angle" ||
```
azeey, line 376: Glad you caught that this was missing. It's weird that "angle" is never used by any of the description files but it's one of the types in the variant though. Mind creating an issue?

src/Param_TEST.cc
jennuine: Since this section was removed:
https://github.com/ignitionrobotics/sdformat/blob/f3dbe1d225939ff1bc038b0789fff884d763cfa7/include/sdf/Param.hh#L296-L315

It now reaches this error: https://github.com/ignitionrobotics/sdformat/blob/f0f9759ec9d4572db9c5fd9e0dd861e6f679a7cf/src/Param.cc#L329

To me, it made more sense to error when a random value such as `%` was set rather than defaulting to `false`. Should I change this back?

src/Param.cc
jennuine: How's this? https://github.com/ignitionrobotics/sdformat/issues/654

azeey: Thanks!

src/Param_TEST.cc
jennuine: How's this? 650b544c92c2e1abfb4b7adf6198e7a21b97187f

This seemed to be the easiest fix. Changing `ParamPrivate::ValueFromStringImpl` bool section to assign the value regardless of the error and return true (since `_valueToSet` was successfully set) caused `Param.InvalidBool` test to fail

azeey: That works. Can you add a comment there saying that the piece of code handling bool types is there to keep backward behavior?

I'm thinking we should change the behavior for Fortress. Would you agree?

azeey (approved):

src/Param_TEST.cc
jennuine: I agree. I'll add the comment and put a `TODO` to remove for Fortress

jennuine: d3532946bc64fef713b30fbb1ab21a33603bf9dd

PR #647, opened 2021-07-29
added light_map to material spec

jennuine:
# 🦟 Bug fix

Related to https://github.com/ignitionrobotics/sdformat/issues/639, changes will need to be ported to later versions to fully resolve.

## Summary
Added missing `//light_map` from `//material/pbr/metal` and `//material/pbr/specular` specification.

## Checklist
- [ ] Added tests

### DIFF
--- a/sdf/1.6/material.sdf
+++ b/sdf/1.6/material.sdf
@@ -90,6 +90,14 @@
       <element name="emissive_map" type="string" default="" required="0">
         <description>Filename of the emissive map.</description>
       </element>
+
+      <element name="light_map" type="string" default="" required="0">
+        <attribute name="uv_set" type="unsigned int" default="0" required="0">
+          <description>Index of the texture coordinate set to use.</description>
+        </attribute>
+        <description>Filename of the light map. The light map is a prebaked light texture that is applied over the albedo map</description>
+      </element>
+
     </element>
 
     <element name="specular" required="0">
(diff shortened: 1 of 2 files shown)


iche033:
note the matching DOM classes are only available in `sdf10`, added in https://github.com/ignitionrobotics/sdformat/pull/429. It may be ok - it just means that the field just won't be accessible through the [PbrWorkflow](https://github.com/ignitionrobotics/sdformat/blob/sdf10/include/sdf/Pbr.hh#L186) class in `sdf9`.

chapulina:
>  it just means that the field just won't be accessible through the PbrWorkflow class in sdf9.

I think this can become confusing for downstream users. We should be moving towards parity between the XML spec and the C++ classes.

jennuine:
> note the matching DOM classes are only available in sdf10, added in #429.

I looked at the wrong branch earlier, thank you for pointing that out. 

> I think this can become confusing for downstream users. We should be moving towards parity between the XML spec and the C++ classes.

I think it's currently confusing for downstream users now since `//light_map` is in the 1.7 spec but only in `sdf10` and not `sdf9`. I think we should backport #429 to `sdf9` for consistency. 

I'll close this PR and open a new one with the backport.

jennuine:
Backport PR: https://github.com/ignitionrobotics/sdformat/pull/648

PR #668, opened 2021-08-12
BUG: add missing //plugin to //include

FirefoxMetzger:
# 🦟 Bug fix

Partially addresses #641 . It adds the missing element, but doesn't do anything about additional namespaced elements.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge*

PR #675, opened 2021-08-25
BUG: add missing plugin element to include

FirefoxMetzger:
As per comment in #668 ; apply the same changes to sdf9

(cherry picked from commit 232ac5c985d7cc9e94454188c0fbf7159f4a876f)


azeey (approved):
Tested locally on Gazebo classic and didn't find any regression

PR #680, opened 2021-08-31
Upgrade cpplint and fix new errors

azeey:
## Summary
Upgrade cpplint to the version found in `ign-cmake2`. @scpeters and I had discussed using ign-cmake to create a `codecheck` target, but it looks like we don't directly depend on `ign-cmake` in the `sdf9` branch. So instead, I simply copied `cpplint.py` from `ign-cmake`. We might be able to add the `codecheck` target for `sdf10`

## Test it
`sh tools/code_check.sh`

## Checklist
- [ ] Added tests

### DIFF
--- a/src/Converter_TEST.cc
+++ b/src/Converter_TEST.cc
@@ -1885,10 +1885,16 @@ TEST(Converter, MuchNewerVersion)
   ASSERT_TRUE(sdf::Converter::Convert(&xmlDoc, "1.6"));
 }
 
-const std::string CONVERT_DOC_15_16 =
-  sdf::filesystem::append(PROJECT_SOURCE_PATH, "sdf", "1.6", "1_5.convert");
-const std::string CONVERT_DOC_16_17 =
-  sdf::filesystem::append(PROJECT_SOURCE_PATH, "sdf", "1.7", "1_6.convert");
+static std::string ConvertDoc_15_16()
+{
+  return sdf::filesystem::append(PROJECT_SOURCE_PATH, "sdf", "1.6",
+                                 "1_5.convert");
+}
+static std::string ConvertDoc_16_17()
+{
+  return sdf::filesystem::append(PROJECT_SOURCE_PATH, "sdf", "1.7",
+                                 "1_6.convert");
+}
 
 /////////////////////////////////////////////////
 /// Test conversion of imu in 1.5 to 1.6
@@ -1930,7 +1936,7 @@ TEST(Converter, IMU_15_to_16)
 
   // Convert
   TiXmlDocument convertXmlDoc;
-  convertXmlDoc.LoadFile(CONVERT_DOC_15_16);
+  convertXmlDoc.LoadFile(ConvertDoc_15_16());
   sdf::Converter::Convert(&xmlDoc, &convertXmlDoc);
 
   // Check some basic elements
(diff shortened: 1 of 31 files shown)


src/Converter_TEST.cc
```diff
@@ -1885,10 +1885,16 @@ TEST(Converter, MuchNewerVersion)
   ASSERT_TRUE(sdf::Converter::Convert(&xmlDoc, "1.6"));
 }
 
-const std::string CONVERT_DOC_15_16 =
-  sdf::filesystem::append(PROJECT_SOURCE_PATH, "sdf", "1.6", "1_5.convert");
-const std::string CONVERT_DOC_16_17 =
-  sdf::filesystem::append(PROJECT_SOURCE_PATH, "sdf", "1.7", "1_6.convert");
+static std::string ConvertDoc_15_16()
```
chapulina, line 1888: So did cpplint actually catch this?

chapulina (approved):

scpeters:
> it looks like we don't directly depend on `ign-cmake` in the `sdf9` branch

we depend on ignition-math6, which depends on ignition-cmake2, so I consider it a valid dependency. It would only be used as a build dependency anyway

scpeters (approved):

src/Converter_TEST.cc
azeey: I meant to mention this in the description, but yes, cpplint caught the global string as a `runtime/string` error:
https://github.com/azeey/sdformat/blob/f5c8bb383fe13d2417f960afb261d626f1cbd40c/tools/cpplint.py#L5455. This helps some with the GSG violations we talked about, but it only works for strings.

azeey:
> we depend on ignition-math6, which depends on ignition-cmake2, so I consider it a valid dependency. It would only be used as a build dependency anyway

Okay, I knew about the dependency, but I wasn't sure if it would be possible to build `sdformat` with just `ign-math` installed and `ign-cmake` not installed. If it's okay to add the dependency, I'll do a followup PR that adds the `codecheck` target.

scpeters:
> > we depend on ignition-math6, which depends on ignition-cmake2, so I consider it a valid dependency. It would only be used as a build dependency anyway
> 
> Okay, I knew about the dependency, but I wasn't sure if it would be possible to build `sdformat` with just `ign-math` installed and `ign-cmake` not installed. If it's okay to add the dependency, I'll do a followup PR that adds the `codecheck` target.

cmake fails to find `ignition-math6` without `ignition-cmake2`. I used `brew unlink ignition-cmake2` on my macOS machine and then get the following configuration error with `sdf9`:

~~~
-- Looking for ignition-math6-config.cmake - not found
-- 	Missing: Ignition math (libignition-math6-dev)
...
CMake Error at CMakeLists.txt:240 (message):
  -- BUILD ERRORS: These must be resolved before compiling.


CMake Error at CMakeLists.txt:242 (message):
  -- 	Missing: Ignition math (libignition-math6-dev)


CMake Error at CMakeLists.txt:244 (message):
  -- END BUILD ERRORS



CMake Error at CMakeLists.txt:245 (message):
  Errors encountered in build.  Please see the BUILD ERRORS above.
~~~

so I think it's safe to add an explicit dependency

azeey:
Sounds good. I've created a follow up PR https://github.com/ignitionrobotics/sdformat/pull/682

PR #682, opened 2021-09-01
Add a codecheck make target

azeey:
# 🎉 New feature

## Summary
Follow up from https://github.com/ignitionrobotics/sdformat/pull/680#issuecomment-910467529

Utilizes ignition-cmake's IgnCodecheck module to create a `codecheck` target.
Needs https://github.com/ignitionrobotics/ign-cmake/pull/186 to eliminate some false positives.

## Test it
Run `make codecheck`

## Checklist
- [ ] Added tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,6 +281,28 @@ else (build_errors)
   endif(BUILD_SDF)
 
   ########################################
+  # Setup Codecheck
+  include (IgnCodeCheck)
+  set(CPPCHECK_DIRS
+    ${PROJECT_SOURCE_DIR}/src
+    ${PROJECT_SOURCE_DIR}/include
+    ${PROJECT_SOURCE_DIR}/test/integration
+    ${PROJECT_SOURCE_DIR}/test/performance)
+
+  set(CPPCHECK_INCLUDE_DIRS
+    ${PROJECT_BINARY_DIR}
+    ${PROJECT_SOURCE_DIR}/include
+    ${PROJECT_SOURCE_DIR}/test/integration
+    ${PROJECT_SOURCE_DIR}/test/performance)
+
+  # Ignore vendored directories.
+  file(WRITE ${PROJECT_BINARY_DIR}/cppcheck.suppress
+    "*:${PROJECT_SOURCE_DIR}/src/win/*\n"
+    "*:${PROJECT_SOURCE_DIR}/src/urdf/*\n"
+    )
+  ign_setup_target_for_codecheck()
+
+  ########################################
   # Make the package config file
   configure_file(${CMAKE_SOURCE_DIR}/cmake/sdformat_pc.in
     ${CMAKE_CURRENT_BINARY_DIR}/cmake/${PROJECT_NAME_LOWER}.pc @ONLY)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -280,6 +280,28 @@ else (build_errors)
...
+    ${PROJECT_SOURCE_DIR}/include
+    ${PROJECT_SOURCE_DIR}/test/integration
+    ${PROJECT_SOURCE_DIR}/test/performance)
+
+  # Ignore vendored directories.
+  file(WRITE ${PROJECT_BINARY_DIR}/cppcheck.suppress
+    "*:${PROJECT_SOURCE_DIR}/src/win/*\n"
+    "*:${PROJECT_SOURCE_DIR}/src/urdf/*\n"
+    )
+  ign_setup_target_for_codecheck()
```
adlarkin, line 303: Is this needed? Looking at other repositories (`ign-gazebo`, `ign-launch`), it looks like these calls are not used in the `CMakeLists.txt`, but I'm not sure how codecheck is being setup/configured on these other repositories.

cmake/SearchForStuff.cmake
```diff
@@ -126,6 +126,13 @@ macro (check_gcc_visibility)
   check_cxx_compiler_flag(-fvisibility=hidden GCC_SUPPORTS_VISIBILITY)
 endmacro()
 
+########################################
+# Find ignition cmake2
+# Only for using the testing macros and creating the codecheck target, not 
+# really being use to configure the whole project
+find_package(ignition-cmake2 2.3 REQUIRED)
```
adlarkin, line 133: Is there a particular reason why version `2.3` is the minimum required version? It looks like `ign_setup_target_for_codecheck` exists before `2.3`:  https://github.com/ignitionrobotics/ign-cmake/blob/ignition-cmake2_2.2.0/cmake/IgnCodeCheck.cmake#L1-L2

adlarkin (approved):
LGTM. I tested it with https://github.com/ignitionrobotics/ign-cmake/pull/186 and had no issues. I just have one minor question.

cmake/SearchForStuff.cmake
azeey: This will actually need to be updated since this PR needs a new release of ign-cmake to ensure `make codecheck` is clean.

scpeters: I wouldn't add that as a version requirement, since that will cause build failures if people have an outdated version of ign-cmake, which is a bit extreme to ensure correct linting. I would recommend a `find_package` call with no version requirement, since `IgnCodeCheck` was present in ign-cmake's 2.0.0 release

adlarkin: That's a good point; I think I agree with @scpeters comments.

azeey: Makes sense. If it won't cause failure in CI, since we assume CI will have the latest ign-cmake, we can remove the version requirement.

Also, I originally thought https://github.com/ignitionrobotics/ign-cmake/pull/187 would be needed to make CI pass, but it turns out without that PR, we do get output from codechech, but they are just warnings, not errors. So technically, this PR can go in without the ign-cmake release.

azeey: https://github.com/ignitionrobotics/sdformat/pull/682/commits/9a6d507226b7ccba33011b4a59043a28499ad03a

scpeters (approved):

azeey:
@osrf-jenkins run test

PR #695, opened 2021-09-08
Fix world-complete.sdf and add particle_scatter_ratio to v1.8

### LINKED ISSUE #670: SDF v1.7 is missing //light/intensity but `/test/sdf/world_complete.sdf` uses it
Here is the relevant snippet of the SDF test file:

https://github.com/ignitionrobotics/sdformat/blob/5a8e796c1c8c8c8c51fd628786e726631f83e8a4/test/sdf/world_complete.sdf#L53-L62

However, the spec mentions //light/intensity starting with SDFv1.8 not SDFv1.7. Is this an omission of the spec, or an invalid test file? 

If the former, I can submit a PR that adds it once someone lets me know which version it was introduced in.

In the latter case, should the version of the file be bumped to v1.8? In this case, //particle_emitter/particle_scatter_ratio has been removed in v1.8 and needs fixing in the test file. Alternatively, the //light/intensity element could be removed (fixes the problem).

FirefoxMetzger:
# 🦟 Bug fix

Fixes #670 

As discussed in #670 this PR bumps the SDF version of the test to 1.8 to allow using the //light/intensity element. It also adds the //particle_emitter/particle_scatter_ratio to v1.8 which appears to have been forgotten in the forward port.

### DIFF
--- a/sdf/1.8/particle_emitter.sdf
+++ b/sdf/1.8/particle_emitter.sdf
@@ -104,6 +104,17 @@
     </description>
   </element>
 
+  <element name="particle_scatter_ratio" type="float" default="0.65" required="0">
+    <description>
+    This is used to determine the ratio of particles that will be detected
+    by sensors. Increasing the ratio means there is a higher chance of
+    particles reflecting and interfering with depth sensing, making the
+    emitter appear more dense. Decreasing the ratio decreases the chance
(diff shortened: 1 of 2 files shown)


azeey (approved):
Thanks!

PR #659, opened 2021-08-06
Support merge-include of nested models

azeey:
# 🎉 New feature

## Summary
This adds a new attribute, `merge`, to `//include` that when set to true copies all the children elements of the nested model into the parent model without introducing a new scope. Currently, we only merge named elements such as `//link`, `//frame` , and `//joint` as well as custom elements. This means we ignore elements like `//static` and `//enable_wind`.
 
For example, given the parent model:
```xml
<?xml version="1.0" ?>
<sdf version="1.9">
  <world name="world_model">
    <model name="robot">
      <include merge="true">
        <uri>test_model</uri>
        <pose>100 0 0 0 0 0</pose>
      </include>
    </model>
  </world>
</sdf>
```
and the included model:
```xml
<sdf version="1.9">
  <model name="test_model">
    <link name="L1" />
    <frame name="F1" />
  </model>
</sdf>
```

The result would be:
```xml
<sdf version='1.9'>
  <world name='world_model'>
    <model name='robot'>
      <frame name='test_model__model__' attached_to='L1'>
        <pose relative_to='__model__'>100 0 0 0 -0 0</pose>
      </frame>
      <link name='L1'>
        <pose relative_to='test_model__model__'>0 0 0 0 -0 0</pose>
      </link>
      <frame name='F1' attached_to='test_model__model__'/>
    </model>
  </world>
```

## Test it
Set `SDF_PATH` to the absolute path of `test/integration/model` and run `ign sdf -p test/integration/merge_include_model.sdf`

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

---
This change is [Reviewable](https://reviewable.io/reviews/ignitionrobotics/sdformat/659)

PR #711, opened 2021-09-23
Prefix merged frames with an underscore

azeey:
# 🦟 Bug fix

## Summary
This would avoid confusion and might help users recognize frames created during include-merge as special frames that are only there as an implementation detail.

## Checklist
- [x] Added tests

### DIFF
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -247,7 +247,7 @@ static void insertIncludedElement(sdf::SDFPtr _includeSDF,
 
   ElementPtr proxyModelFrame = _parent->AddElement("frame");
   const std::string proxyModelFrameName =
-      "merged__" + model->Name() + "__model__";
+      "_merged__" + model->Name() + "__model__";
 
   proxyModelFrame->GetAttribute("name")->Set(proxyModelFrameName);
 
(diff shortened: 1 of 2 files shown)


src/parser.cc
```diff
@@ -247,7 +247,7 @@ static void insertIncludedElement(sdf::SDFPtr _includeSDF,
 
   ElementPtr proxyModelFrame = _parent->AddElement("frame");
   const std::string proxyModelFrameName =
-      "merged__" + model->Name() + "__model__";
+      "_merged__" + model->Name() + "__model__";
```
caguero, line 250: Just double checking, single underscore as a prefix, not two underscores, right?

azeey: Yes, just single. Names that start and end with a double underscore are not permitted per http://sdformat.org/tutorials?tut=pose_frame_semantics&cat=specification&#unique-names-and-reserved-names, but we still wanted this name to have a special meaning.

scpeters:
makes sense to me. should we add a test expectation to one of the tests from #659 about the name of these special frames?

jennuine:
> makes sense to me. should we add a test expectation to one of the tests from #659 about the name of these special frames?

e40258c

src/parser.cc
caguero: I see, thanks for the clarification!

scpeters (approved):

caguero (approved):

azeey:
Thanks for adding the test @jennuine!

PR #720, opened 2021-09-29
Remove empty //inertial/pose/@relative_to during 1_7->1.8 conversion

scpeters:
# 🦟 Bug fix

Fixes warnings when loading models from 1.7 or earlier with `//inertial/pose/@relative_to`, part of https://github.com/ignitionrobotics/ign-gazebo/issues/1056

## Summary

The `//inertial/pose/@relative_to` attribute was removed from the 1.8 spec in 93f8c56b6a9c889500db17654fcb83c2f825c598, which causes warnings when loading a file from 1.7 or earlier that contains this attribute. This PR removes this attribute during conversion from 1.7 -> 1.8 if the attribute is empty, which should silence many of the warnings, while retaining those warnings that may require action from the user.

I added an empty attribute to a `test/sdf` file in https://github.com/ignitionrobotics/sdformat/pull/720/commits/34d95879c2e620ed10df69515e4ad211746f6031, which causes a warning that causes `UNIT_ign_TEST` to fail. The failure is fixed by new functionality in the `Converter` class, which allows removal of empty attributes or elements.

## Checklist
- [X] Added tests

### DIFF
--- a/src/Converter.cc
+++ b/src/Converter.cc
@@ -645,7 +645,7 @@ void Converter::Add(tinyxml2::XMLElement *_elem, tinyxml2::XMLElement *_addElem)
 /////////////////////////////////////////////////
 void Converter::Remove(tinyxml2::XMLElement *_elem,
                        tinyxml2::XMLElement *_removeElem,
-                       bool _removeEmpty)
+                       bool _removeOnlyEmpty)
 {
   SDF_ASSERT(_elem != NULL, "SDF element is NULL");
   SDF_ASSERT(_removeElem != NULL, "remove element is NULL");
@@ -663,7 +663,7 @@ void Converter::Remove(tinyxml2::XMLElement *_elem,
   if (attributeName)
   {
     const char * attributeValue = _elem->Attribute(attributeName);
-    if (!_removeEmpty || (attributeValue && strlen(attributeValue) == 0))
+    if (!_removeOnlyEmpty || (attributeValue && strlen(attributeValue) == 0))
     {
       _elem->DeleteAttribute(attributeName);
     }
@@ -675,7 +675,8 @@ void Converter::Remove(tinyxml2::XMLElement *_elem,
     while (childElem)
     {
       auto nextSibling = childElem->NextSiblingElement(elementName);
-      if (!_removeEmpty || (childElem->NoChildren() && !childElem->GetText()))
+      if (!_removeOnlyEmpty ||
+          (childElem->NoChildren() && !childElem->GetText()))
       {
         _elem->DeleteChild(childElem);
       }
--- a/src/Converter.cc
+++ b/src/Converter.cc
@@ -675,8 +675,8 @@ void Converter::Remove(tinyxml2::XMLElement *_elem,
     while (childElem)
     {
       auto nextSibling = childElem->NextSiblingElement(elementName);
-      if (!_removeOnlyEmpty ||
-          (childElem->NoChildren() && !childElem->GetText()))
+      if (!_removeOnlyEmpty || (!childElem->FirstAttribute() &&
+          childElem->NoChildren() && !childElem->GetText()))
       {
         _elem->DeleteChild(childElem);
       }
--- a/src/Converter.cc
+++ b/src/Converter.cc
@@ -663,7 +663,7 @@ void Converter::Remove(tinyxml2::XMLElement *_elem,
   if (attributeName)
   {
     const char * attributeValue = _elem->Attribute(attributeName);
-    if (!_removeEmpty || (attributeValue && attributeValue[0] == '\0'))
+    if (!_removeEmpty || (attributeValue && strlen(attributeValue) == 0))
     {
       _elem->DeleteAttribute(attributeName);
     }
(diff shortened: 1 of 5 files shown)


src/Converter.cc
```diff
@@ -657,16 +662,24 @@ void Converter::Remove(tinyxml2::XMLElement *_elem,
 
   if (attributeName)
   {
-    _elem->DeleteAttribute(attributeName);
+    const char * attributeValue = _elem->Attribute(attributeName);
+    if (!_removeEmpty || (attributeValue && attributeValue[0] == '\0'))
```
azeey, line 666: nit: This works fine, but I think using `std::strlen(attributeValue) == 0 ` would be more easily readable

src/Converter.hh
```diff
@@ -100,9 +100,11 @@ namespace sdf
 
     /// \brief Remove an element.
     /// \param[in] _elem The element that has the _removeElem child.
-    /// \param[in] _removeElem The element to remove.
+    /// \param[in] _removeElem The metadata about what to remove.
+    /// \param[in] _removeEmpty If true, only remove empty nodes.
```
azeey, line 104: Can you document how this works for attributes and elements? From what I can understand, for attributes, it removes just the attribute if it's empty. For elements, it removes the element if it has an empty value and has no children, but it can have attributes.

src/Converter_TEST.cc
```diff
@@ -618,6 +619,118 @@ TEST(Converter, RemoveDescendantElement)
...
+                << "    <convert name='elemC'>"
+                << "      <remove_empty element='elemD'/>"
+                << "    </convert>"
+                << "  </convert>"
+                << "</convert>";
+  tinyxml2::XMLDocument convertXmlDoc;
+  convertXmlDoc.Parse(convertStream.str().c_str());
+  sdf::Converter::Convert(&xmlDoc, &convertXmlDoc);
+
+  tinyxml2::XMLElement *convertedElem =  xmlDoc.FirstChildElement();
```
azeey, line 660: Can you add an assertion on `convertElem` here?

src/Converter_TEST.cc
```diff
@@ -618,6 +619,118 @@ TEST(Converter, RemoveDescendantElement)
...
+  convertStream << "<convert name='elemA'>"
+                << "  <convert descendant_name='elemC'>"
+                << "    <remove_empty element='elemD'/>"
+                << "  </convert>"
+                << "</convert>";
+  tinyxml2::XMLDocument convertXmlDoc;
+  convertXmlDoc.Parse(convertStream.str().c_str());
+  sdf::Converter::Convert(&xmlDoc, &convertXmlDoc);
+
+  tinyxml2::XMLElement *convertedElem = xmlDoc.FirstChildElement();
```
azeey, line 714: Add assertion for `convertedElem`.

azeey (dismissed):
Just a few minor comments.

src/Converter.hh
scpeters: would it be better to change the behavior for removing elements if they also have no attributes? We aren't yet using that functionality, so I couldn't decide. For now, I'll just document it as you suggest

src/Converter_TEST.cc
scpeters: https://github.com/ignitionrobotics/sdformat/pull/720/commits/4df59689dd05d4e8ff20029879777cfc143efacf

src/Converter_TEST.cc
scpeters: https://github.com/ignitionrobotics/sdformat/pull/720/commits/4df59689dd05d4e8ff20029879777cfc143efacf

src/Converter.cc
scpeters: https://github.com/ignitionrobotics/sdformat/pull/720/commits/be159db6bfc592edecc05cfd64bcc16701716eb2

src/Converter.hh
scpeters: updated documentation and renamed the parameter to `_removeOnlyEmpty` in https://github.com/ignitionrobotics/sdformat/pull/720/commits/e6eb29391e12f320619d2ef9d408b847603b2c47

azeey: Is it possible to chain these? By that I mean, would it be possible to remove the attributes first and then remove the element? If so, then I'd say let's make it so that it removes elements only if they have no attributes. This would make it possible to remove just the deprecated attributes first and remove the element if it doesn't contain anything else. For example:
```xml
<test_elem attr1="" attr2="some_val"/>
<test_elem attr1="" />
```

If we deprecate the use of `//test_elem` and know that `//test_elem[@attr1=""]` is the default, we can do 

```xml
<remove_empty attribute='attr1'/>
<remove_empty element='test_elem'/>
```
and it will only delete the second `//test_elem` and we would be able to issue a warning about the first `//test_elem`

scpeters: sure, I'll rework the function logic accordingly

scpeters: done in https://github.com/ignitionrobotics/sdformat/pull/720/commits/9588e0f82f78fa2ddb57e2c8a54339ca1625b55c

azeey (approved):

osrf-triage:

PR #517, opened 2021-03-17
Add Joint DOM API to access joint sensors

azeey:
# 🎉 New feature

## Summary
This adds  `sdf::Joint::SensorCount`, `sdf::Joint::SensorByIndex`, `sdf::Joint::SensorNameExists`, `sdf::Joint::SensorByName`.

## Test it
Run tests

## Checklist
- [ ] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] Code check passed (In source directory, run `sh tools/code_check.sh`)
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review
[another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+)
to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #732, opened 2021-10-14
Backport: Force Torque

azeey:
Backport #393, #517, #669.


chapulina:
Should / can it backported to `sdf9` and then merged forward?

azeey:
> Should / can it backported to `sdf9` and then merged forward?

`sdf9` doesn't have the `ForceTorque` class since it was only added to `sdf11` in #393, so I thought I should backport the noise stuff to just `sdf11`. But looking at the code, I don't see why #393 can't be backported to `sdf9`, so I'll go ahead and do that.

Issue #181, opened 2018-04-30
Use ign-cmake.


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/181)) by Nate Koenig (Bitbucket: [Nathan Koenig](https://bitbucket.org/%7Bc862cdd9-fcc8-4419-9fc7-e20db14b8fcb%7D/)).**

----------------------------------------
Switch this project to use ign-cmake. This will likely require moving osrf/sdf to ignition/sdf

scpeters:
This would be a very big change

* We would probably want to move from the `osrf'` GitHub org to `ignitionrobotics`. This would require communication and agreement but shouldn't be too disruptive overall
* the cmake project name would have to change to start with `ignition-`
* the [code for auto-generating the all-inclusive header file](https://github.com/ignitionrobotics/ign-cmake/blob/3b7870289027ae0fc04c83d5817907ed863a0720/cmake/IgnUtils.cmake#L685-L711) assumes that headers have `ignition/` at the start of the include path, so the include paths would all change from `sdf/*` to `ignition/sdf/*` or `ignition/sdformat/*` or something
* this repository holds both the specification (SDFormat) and our reference parsing library (libsdformat). If we are making big changes, we should consider if we want to split these two into separate repositories so we can factor that in to the overall migration plan

chapulina:
> We would probably want to move from the osrf' GitHub org to ignitionrobotics.

This has been brought up several times and could happen regardless of whether we use `ign-cmake` or not. Some advantages:

* Help with visibility into the SDFormat repo. The `osrf` org is quite bloated and it's hard to find what's important and active. While the `ignitionrobotics` org is more focused.
* SDFormat is being released as part of [Ignition collections](https://ignitionrobotics.org/libs), so it makes sense to move it closer to the rest of the libraries
* It would make a lot of our infra easier, we have to check in several places `if org == "osrf"` just for `sdformat`
* SDF tickets would automatically land in https://github.com/orgs/ignitionrobotics/projects/3 and be tracked by the Ignition team

The transition should also be seamless for both users and tools, because GitHub is pretty good about redirecting.

scpeters:
> * the cmake project name would have to change to start with `ignition-`
[...]

PR #736, opened 2021-10-27
USD to SDF converter

ahcorde:
# 🎉 New feature

## Summary

FYI @adlarkin @koonpeng . Feel free to push code, cleanups or suggestions

This draft PR creates a prototype to convert USD files into SDF files.

For now I already tested with `simple_articulated.usd` and `ur10.usd`. 

Regarding joints there are two fields `physics:body0` and `physics:body1` that indicate the parent and child of the joint, but in usd format there is no distinction between them. We need to rework this part a little bit more. I have to modify `ur10.usda` to set parent as `body1` and child as `body0`. No changes are required in the `simple_articulated.usd` file.

The current implementation supports:
 - **Links**
   - Geometris: cylinders, spheres and boxes
   - Collisions
   - Color
 - **Joints**: fixed, revolute and prismatic
 - **Physics** (only fill data structure)

Code need a huge refactor. We can probably use some ignition library such as ignition-common or ignition-math  in the `usd_parser` folders because I copy-paste the code from urdf. There are object like `Sphere`, poses, Vectors quaternions, links and many other that I'm pretty sure that we can use something from these libraries.

ur10

articulated_arm
## Test it

 - Install [PXR lbrary](https://github.com/PixarAnimationStudios/USD).
    - If you only want the cpp API you should compile the library with `-DPXR_ENABLE_PYTHON_SUPPORT=0`
    - if you compile the library with python support you will be able to use some of the usd tools such us:
        - `usdcat`: It allows to convert `usd` (binary format) into `usda` (human redeable format). `usdcat binary.usa > output.usda
        - `usdview`. It allows to visualize `usd` files
 - There is a cmd util called `sdfconverter` which allows to convert a `urdf` or `usd` files in to `sdf`. it's simple to use: `sdfconverter -i <input> -o <ouput>`

## Checklist
- [ ] Added tests

### DIFF
--- a/src/usd/usd_parser/sensors.cc
+++ b/src/usd/usd_parser/sensors.cc
@@ -57,7 +57,8 @@ namespace usd
       camera.SetName(_prim.GetPath().GetName());
       camera.SetHorizontalFov(horizontalAperture);
       camera.SetLensFocalLength(focalLength);
-      camera.SetRawPose(pose);
+      ignition::math::Pose3d poseCamera(0, 0, 0, 1.57, 0, -1.57);
+      camera.SetRawPose(pose * -poseCamera);
       camera.SetNearClip(clippingRange[0]);
       camera.SetFarClip(clippingRange[1]);
       camera.SetImageWidth(640);
--- a/src/parser_usd.cc
+++ b/src/parser_usd.cc
@@ -45,7 +45,6 @@ inline namespace SDF_VERSION_NAMESPACE {
   typedef std::map<std::string, std::vector<SDFExtensionPtr> >
     StringSDFExtensionPtrMap;
 
-
   USD2SDF::USD2SDF()
   {
     g_fixedJointsTransformedInRevoluteJoints.clear();
@@ -308,14 +307,14 @@ inline namespace SDF_VERSION_NAMESPACE {
         _oldLinkName.empty())
     {
       // std::cerr << "/* visual name " << _oldLinkName.c_str() << '\n';
-      sdfVisual->SetAttribute("name", _oldLinkName.c_str());
+      sdfVisual->SetAttribute("name", ignition::common::basename(_oldLinkName).c_str());
     }
     else
     {
       // std::cerr << "/* visual name " << _link->name + kLumpPrefix + _oldLinkName << '\n';
 
       sdfVisual->SetAttribute("name",
-          (_link->name + kLumpPrefix + _oldLinkName).c_str());
+          ignition::common::basename(_link->name + kLumpPrefix + _oldLinkName).c_str());
     }
 
     // add the visualisation transfrom
@@ -702,11 +701,11 @@ inline namespace SDF_VERSION_NAMESPACE {
     if (_oldLinkName.compare(0, _link->name.size(), _link->name) == 0 ||
         _oldLinkName.empty())
     {
-      sdfCollision->SetAttribute("name", _oldLinkName.c_str());
+      sdfCollision->SetAttribute("name", ignition::common::basename(_oldLinkName).c_str());
     }
     else
     {
-      sdfCollision->SetAttribute("name", (_link->name
+      sdfCollision->SetAttribute("name", ignition::common::basename(_link->name
           + kLumpPrefix + _oldLinkName).c_str());
     }
 
@@ -902,7 +901,7 @@ inline namespace SDF_VERSION_NAMESPACE {
       {
         joint->SetAttribute("type", jtype.c_str());
       }
-      joint->SetAttribute("name", _link->parent_joint->Name().c_str());
+      joint->SetAttribute("name", ignition::common::basename(_link->parent_joint->Name()).c_str());
       // Add joint pose relative to parent link
       AddTransform(
           joint, _link->parent_joint->RawPose());
@@ -912,10 +911,10 @@ inline namespace SDF_VERSION_NAMESPACE {
       {
         relativeToAttr = "__model__";
       }
-      pose->SetAttribute("relative_to", relativeToAttr.c_str());
+      pose->SetAttribute("relative_to", ignition::common::basename(relativeToAttr).c_str());
 
-      AddKeyValue(joint, "parent", _link->getParent()->name);
-      AddKeyValue(joint, "child", _link->name);
+      AddKeyValue(joint, "parent", ignition::common::basename(_link->getParent()->name));
+      AddKeyValue(joint, "child", ignition::common::basename(_link->name));
 
       tinyxml2::XMLElement *jointAxis = doc->NewElement("axis");
       tinyxml2::XMLElement *jointAxisLimit = doc->NewElement("limit");
@@ -1024,7 +1023,7 @@ inline namespace SDF_VERSION_NAMESPACE {
     tinyxml2::XMLElement *elem = _root->GetDocument()->NewElement("link");
 
     // set body name
-    elem->SetAttribute("name", _link->name.c_str());
+    elem->SetAttribute("name", ignition::common::basename(_link->name).c_str());
 
     // compute global transform
     ignition::math::Pose3d localTransform;
@@ -1171,7 +1170,7 @@ inline namespace SDF_VERSION_NAMESPACE {
       if (sdfSensor->Type() == sdf::SensorType::CAMERA)
       {
         const sdf::Camera * camera = sdfSensor->CameraSensor();
-        sensorXML->SetAttribute("name", camera->Name().c_str());
+        sensorXML->SetAttribute("name", ignition::common::basename(camera->Name()).c_str());
 
         double pose_value[6];
         pose_value[0] = camera->RawPose().Pos().X();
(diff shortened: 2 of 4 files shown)


koonpeng:
> Another possibility is keeping USD support in a separate component within this codebase.

I'm very new to the ignition build system so I might be totally off, does a separate component means declaring it with `ign_add_component`? From what I understand, the function adds a "library or component" to a "core library", but I think this PR adds a new executable.

scpeters:
> > sdformat doesn't use ign-cmake2, so it would be some work to add a component to this package
> 
> If this is something that we think would be valuable, it would be a good excuse to update SDFormat to use `ign-cmake` (#181)

I've taken an initial step towards this in https://github.com/ignitionrobotics/ign-cmake/pull/190, which is awaiting review

scpeters:
> > Another possibility is keeping USD support in a separate component within this codebase.
> 
> I'm very new to the ignition build system so I might be totally off, does a separate component means declaring it with `ign_add_component`? From what I understand, the function adds a "library or component" to a "core library", but I think this PR adds a new executable.

There is an aspect that is more significant than whether the code produces a library or executable, which is to define a cmake component for packaging purposes. See for instance the discussion of the `COMPONENTS` keyword in the [find_package documentation](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature). A different component can be found installed and found independently, and is especially useful for isolating dependencies, so you can build or install the core components without needing all the dependencies used by other components.

adlarkin:
> Of all the parts of libsdformat, the parser_urdf code is probably the most in need of improvement. So building USD support off of this code is not the best foundation, though I understand why you would do it, since it is there and basically functional. If we could think of a better implementation for this type of conversion, I think it would be valuable, since this parser_urdf code is so hard to maintain.

On a somewhat related note, it looks like `parser_urdf` is deprecated and should have been removed by now (I think?): https://github.com/ignitionrobotics/sdformat/blob/913a7754c319190defa549bb73dc584e59f9734f/src/parser_urdf.hh#L35-L40

Do we need to wait until the next `sdformat` version to remove this class, or can we do it now since [it's not installed](https://github.com/ignitionrobotics/sdformat/commit/b4efba2b1f1da99002b64d93aeb34a3711a25294)?

scpeters:
> > sdformat doesn't use ign-cmake2, so it would be some work to add a component to this package
> 
> It looks like `ign-cmake2` is listed in `sdformat`'s `packages.apt` file:
> 
> https://github.com/ignitionrobotics/sdformat/blob/5385ce9500bd1662d52ba3d0c2c919d13f86c949/.github/ci/packages.apt#L1
> 
> Is this a mistake, or does `sdformat` actually depend on `ign-cmake2` somehow?

libsdformat depends on ignition-cmake2 but isn't an ignition-cmake project that uses the cmake macros like ign_configure_project and ign_configure_build. Until recently, it only depended on ignition-cmake2 indirectly via ignition-math6, although it now uses `IgnCodeCheck.cmake` as of #682. It will take a bit more work (see https://github.com/ignitionrobotics/sdformat/issues/181) including changes in ign-cmake to allow libsdformat to use the ign-cmake cmake helpers to add a component to this package

adlarkin:
> that comment is wrong; this code is still needed. `parser_urdf.hh` is a private header similar to `FrameSemantics.hh`

I made a PR that removes this comment in #740 

> It will take a bit more work (see #181) including changes in ign-cmake to allow libsdformat to use the ign-cmake cmake helpers to add a component to this package

Before we go ahead and make changes to `ign-cmake`, I think we should determine if we actually want to place the functionality of this PR into a separate component. If we decide on taking a different approach (for example, placing the USD converter in a separate repository), then it wouldn't make sense to put forth the effort of updating `ign-cmake` right now. Does everyone think we should place this PR in a separate component of `sdformat`, or should we consider other alternatives?

scpeters:
> > It will take a bit more work (see #181) including changes in ign-cmake to allow libsdformat to use the ign-cmake cmake helpers to add a component to this package
> 
> Before we go ahead and make changes to `ign-cmake`, I think we should determine if we actually want to place the functionality of this PR into a separate component. If we decide on taking a different approach (for example, placing the USD converter in a separate repository), then it wouldn't make sense to put forth the effort of updating `ign-cmake` right now. Does everyone think we should place this PR in a separate component of `sdformat`, or should we consider other alternatives?

If we want to perpetually support conversion between SDFormat and USD, I think there are advantages to keeping this code in the same repository. Since it requires additional dependencies, I think adding it in a separate component would be appropriate.

If we wanted to put this code in a separate repository, I think that would be a valid choice as well; it would just take a bit more effort to keep the code synchronized and manage version differences. It would also make it easier to limit our commitment to support of the USD format, for better or worse.

src/usd/usd_parser/physics.cc
```diff
@@ -0,0 +1,36 @@
...
+namespace usd
+{
+  void ParsePhysicsScene(const pxr::UsdPrim &_prim, WorldInterfaceSharedPtr &_world)
+  {
+    auto variant_physics_scene = pxr::UsdPhysicsScene(_prim);
+    pxr::GfVec3f gravity;
+    variant_physics_scene.GetGravityDirectionAttr().Get(&gravity);
+    _world->gravity[0] = gravity[0];
+    _world->gravity[1] = gravity[1];
+    _world->gravity[2] = gravity[2];
```
koonpeng, line 33: Do we need to account for up axis?

src/usd/usd_parser/sensors.cc
```diff
@@ -0,0 +1,111 @@
...
+    GetTransform(_prim, _usdData, pose, scale, _linkName);
+    sensor->SetRawPose(pose);
+    if(_prim.IsA<pxr::UsdGeomCamera>())
+    {
+      sensor->SetType(sdf::SensorType::CAMERA);
+
+      sdf::Camera camera;
+
+      auto variantCamera = pxr::UsdGeomCamera(_prim);
+      float horizontalAperture = 20.955;
```
koonpeng, line 49: What is this value based off from?

src/usd/usd_parser/utils.cc
```diff
@@ -0,0 +1,666 @@
...
+
+          std::vector<pxr::UsdShadeInput> inputs = variantshader.GetInputs();
+          for (auto &input : inputs)
+          {
+            std::cerr << "\tGetFullName " << input.GetFullName() << " " << input.GetBaseName() << '\n';
+            if (input.GetBaseName() == "diffuse_texture")
+            {
+              pxr::SdfAssetPath materialPath;
+              pxr::UsdShadeInput diffuseTextureShaderInput =
+                variantshader.GetInput(pxr::TfToken("diffuse_texture"));
```
koonpeng, line 133: Is there a doc explaining the available tokens?

ahcorde:
TODO
 - [ ] Review "Y axis" usd files
 - [ ] Generate triangles from a polygon (with more than 3 vertices per polygon)
     - Remove `CGAL` dependency, use instead `gts`
 - [ ] Cleanup
 - [ ] Document

adlarkin:
I'm going to add the `needs upstream release` label to this PR because the build fails without https://github.com/ignitionrobotics/ign-math/pull/366. Based on the discussion had in https://github.com/ignitionrobotics/ign-math/pull/366#issuecomment-1012356372, I'm wondering if this PR is actually needed - it seems like an unrelated change/addition in this PR is what's introducing the build error. I will follow-up on this if I find anything in this PR that could be related to the build error.

src/usd/usd_parser/utils.cc
```diff
@@ -0,0 +1,666 @@
...
+        std::string upAxis = data.second->_upAxis;
+        std::cerr << "upAxis " << upAxis << '\n';
+      //   if(upAxis == "Y")
+      //   {
+      //     ignition::math::Matrix4d m(
+      //       1, 0, 0, 0,
+      //       0, 0, -1, 0,
+      //       0, 1, 0, 0,
+      //       0, 0, 0, 1);
+      //     t.translate = m * t.translate;
```
adlarkin, line 330: For the scenario of `upAxis == "Y"`, is there a reason why the rotation matrix is 4x4? It should be 3x3, since we are in a 3d space. Reading the [USD upAxis docs](https://graphics.pixar.com/usd/release/api/group___usd_geom_up_axis__group.html#details), I found the following information that might be useful:
```
UsdGeom stipulates a right-handed coordinate system.
Therefore, when viewing a UsdStage with a "Y" up axis, the stage's Z axis will be pointing out of the screen, and when viewing a UsdStage with a "Z" up axis, the stage's Y axis will be pointing into the screen.
```

So, if we are looking at the screen, this means that X will always be pointing to the right (i.e., the X coordinate will never change regardless of the `upAxis`). So, when the `upAxis` is `Y`, I think we need to perform a 90 degree rotation in the YZ plane, from the Y axis to the Z axis (for a visual representation of this, look at `Ryz` [here](https://robotics.stackexchange.com/a/19616)). So, I think that the rotation matrix (`m`) should be this (it worked for a few points I tested on paper/by hand):
```
1     0       0              1 0 0
0  cos(90)  -sin(90)    =    0 0 -1
0  sin(90)   cos(90)         0 1  0
```
Note that this is the matrix `m` you currently have, but with the last row/column removed.

adlarkin (changes requested):
This PR is really large and doesn't have documentation, so it's hard to review all at once. Once https://github.com/ignitionrobotics/sdformat/pull/817 is merged, would you mind refactoring this PR to match the component structure? Then, we can break it up into smaller PRs and review the smaller PRs, like what we're doing for SDF -> USD conversion.

src/usd/usd_parser/utils.cc
adlarkin: It sounds like this is now resolved by https://github.com/ignitionrobotics/sdformat/pull/825

ahcorde:
Closing this PR in favour of this metaticket https://github.com/ignitionrobotics/sdformat/issues/866

PR #744, opened 2021-11-10
Support wide angle camera

iche033:
# 🎉 New feature

## Summary

Updates the Sensor DOM to support parsing wide angle cameras.

## Checklist
- [x] Added tests


ahcorde (dismissed):
can you add some tests?

iche033:
added integration test in 2800f2c

scpeters (approved):

src/Sensor.cc
```diff
@@ -64,7 +64,8 @@ const std::vector<std::string> sensorTypeStrs =
   "navsat",
   "segmentation_camera",
   "boundingbox_camera",
-  "custom"
+  "custom",
+  "wideanglecamera"
```
chapulina, line 68: We should document this string here: https://github.com/ignitionrobotics/sdformat/blob/f2a286cbf89aca7daa722a70a357e0de32224372/sdf/1.9/sensor.sdf#L10-L36

Most other cameras are using `camel_case` (at least partially), can we do it here too? i.e. `wideangle_camera` or `wide_angle_camera`, or all of them to be safe?

scpeters: the existing convention with gazebo classic is [wideanglecamera](https://github.com/osrf/gazebo/blob/gazebo11/worlds/lensflare_plugin.world#L97). I think we can choose to add alternates, but I would keep `wideanglecamera` for backwards compatibility

iche033: changed `wideanglecamera` to `wide_angle_camera` here. Also updated the Load function to check for both `wide_angle_came` and `wideanglecamera` when loading the sdf for backward compatibility. 
8aacb98

ahcorde (approved):

MrKeith99:
Is there any possibility to add this feature in `Gazebo Fortress` as well?
I am struggling to use `Humble` and `gz_ros2_control` within the `Gazebo Fortress` or `Gazebo Garden`.

iche033:
the PRs to add wide angle camera support in gz-rendering (https://github.com/gazebosim/gz-rendering/pull/490 and https://github.com/gazebosim/gz-rendering/pull/733) include ABI breaking changes so I think it'll be difficult to backport all wide angle camera related changes to Fortress.

PR #741, opened 2021-11-09
9 to 10 20211108

nkoenig:
# ➡️ Forward port

Port sdf9 to sdf10

Branch comparison: https://github.com/ignitionrobotics/sdformat/compare/sdf10...sdf9

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)

PR #745, opened 2021-11-11
Fix URDF fixed joint reduction of plugins

scpeters:
# 🦟 Bug fix

Fixes bug in URDF conversion for libsdformat10+

## Summary

While attempting to merge forward #500 from 9 -> 10 (follow-up to #741), I found that the test added in #500 was failing. The test passes on the `sdf6` and `sdf9` branches, which use `tinyxml`, but was failing in my first merge attempt to `sdf10`, which uses `tinyxml2`. To simplify matters, I ported a simplified form of the test from #500 directly to `sdf10` in https://github.com/ignitionrobotics/sdformat/commit/787abbc066061200b2711d63298a26766267f5e8 to illustrate the failure in `ReduceSDFExtensionPluginFrameReplace`. The test is fixed in https://github.com/ignitionrobotics/sdformat/commit/c02e63d212e1b7b72dcd87de76dd9ba3fe73dcf6 by accounting for the extra root element ~~in the tinyxml2 document API~~ in the `blobs` data structure added in #264.

Note that there are many other `ReduceSDFExtension*FrameReplace` functions in `parser_urdf.cc` that are untested and likely suffer this same error. I will open an issue to document this.

## Checklist
- [X] Added tests

### DIFF
--- a/src/parser_urdf.cc
+++ b/src/parser_urdf.cc
@@ -183,7 +183,7 @@ void ReduceSDFExtensionProjectorFrameReplace(
 /// reduced fixed joints:  apply appropriate frame updates in plugins
 ///   inside urdf extensions when doing fixed joint reduction
 void ReduceSDFExtensionPluginFrameReplace(
-      std::vector<XMLDocumentPtr>::iterator _blobIt,
+      tinyxml2::XMLElement *_blob,
       urdf::LinkSharedPtr _link, const std::string &_pluginName,
       const std::string &_elementName,
       ignition::math::Pose3d _reductionTransform);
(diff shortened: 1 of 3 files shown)


scpeters:
> Note that there are many other `ReduceSDFExtension*FrameReplace` functions in `parser_urdf.cc` that are untested and likely suffer this same error. I will open an issue to document this.

#746

src/parser_urdf.cc
```diff
@@ -3588,17 +3588,18 @@ void ReduceSDFExtensionPluginFrameReplace(
 {
   std::string linkName = _link->name;
   std::string parentLinkName = _link->getParent()->name;
-  if ((*_blobIt)->FirstChildElement()->Name() == _pluginName)
+  tinyxml2::XMLElement *rootElement = (*_blobIt)->FirstChildElement();
```
iche033, line 3591: add a comment here that there is an extra root element in tinyxml2?

iche033 (approved):
> by accounting for the extra root element in the tinyxml2 document API.

ouch :/

Changes look good to me.

src/parser_urdf.cc
scpeters: ok, I think it's not an issue with the tinyxml2 API but actually a change in how we used it from #264...

scpeters: we changed what we store in the SDFExtension data structure here: https://github.com/ignitionrobotics/sdformat/pull/264/commits/6dd6d07d38871c223ff278186041df5252c024a1#diff-2bc5ca23bcfc66fe173f513399bec8065b1cb4e607d741566517d7368e6bce0fL1315-L1320

I'll update the PR description accordingly

scpeters: I just refactored the fix in https://github.com/ignitionrobotics/sdformat/pull/745/commits/b98b827670614f413b3e6a8c1633c6978ea68b19: I change the function signature so it now accepts an `XMLElement` pointer instead of an `XMLDocument` iterator. Then we can call `FirstChildElement()` once and pass the resulting pointer, which simplifies the implementation in `ReduceSDFExtensionPluginFrameReplace`

I'll fix the other functions in follow-up pull requests if you think this approach is ok

iche033: yep the approach looks good to me.

osrf-triage:

PR #748, opened 2021-11-11
Suggested changes to #729

azeey:
These changes make it so that the default string used in the sdf description is used as-is if an element is added by default or it's value is empty. This avoids introducing changes in the output of `ToString` (eg. `<magnetic_field>5.5645e-6 22.8758e-6 -42.3884e-6</magnetic_field>` as opposed to `<magnetic_field>6e-06 2.3e-05 -4.2e-05</magnetic_field>` in #729)


aaronchongth (approved):
Yeah this makes a lot of sense. Thanks for finding a solution to this dilemma :bow: I'll go ahead and merge it in

PR #754, opened 2021-11-15
Suggestion to move dom conversion to sensor classes

nkoenig:
Suggestion to #753. This separates out all the conversion logic into each sensor object.


iche033 (approved):
looks good to me

PR #755, opened 2021-11-16
Populate light sdf::ElementPtr from Light DOM

iche033:
# 🎉 New feature

needed by https://github.com/ignitionrobotics/ign-gazebo/pull/1209

## Summary

similar to #753 

Adds a function in Light DOM class to populate an sdf::ElementPtr

## Checklist
- [x] Added tests


test/integration/sdf_dom_conversion.cc
```diff
@@ -402,3 +404,181 @@ TEST(SDFDomConversion, Sensors)
...
 }
+
+//////////////////////////////////////////////////
+TEST(SDFDomConversion, Lights)
+{
+  // this test loads the lights.sdf test file, then
+  // 1) converts Light DOM to Element
+  // 2) loads the Element back to DOM,
+  // 3) verify the values
+  const std::string testFile = sdf::testing::TestFile("sdf", "lights.sdf");
```
nkoenig, line 415: I think you forgot to include the `lights.sdf` file.

include/sdf/Light.hh
```diff
@@ -268,6 +268,11 @@ namespace sdf
     private: void SetPoseRelativeToGraph(
         sdf::ScopedGraph<PoseRelativeToGraph> _graph);
 
+    /// \brief Fill the provided _elem with data from this light.
+    /// \param[out] _elem SDF element point to populate
+    /// \return True if successful.
+    public: bool PopulateElement(sdf::ElementPtr _elem) const;
```
azeey, line 274: Can you add documentation that the user first needs to:
  * Allocate memory for `_elem`. i.e, just creating an `sdf::ElementPtr` object and calling this function with it would crash the program.
  * Initialize `_elem` with `sdf::initFile` using the correct description file. In this case, `light.sdf`.

To avoid confusion, would it be better for the function to do both and return an `sdf::ElementPtr` instead?

iche033: > To avoid confusion, would it be better for the function to do both and return an sdf::ElementPtr instead?

do you mean something like `sdf::ElementPtr Element() const` that creates, initializes, and returns the sdf::ElementPtr?

azeey: Yes, but named differently so it doesn't collide with the existing `Element()` functions in DOM objects. Maybe `BuildElement` or `CreateElement`?

iche033: ok sounds good. How about `ToElement`? same idea as the other `ToString()` functions in sdformat?

azeey: Yeah, `ToElement` sounds good!

iche033: done. 5b68f18

iche033: I'll create a separate PR to update the other PopulateElement APIs introduced in https://github.com/ignitionrobotics/sdformat/pull/753

src/Light.cc
```diff
@@ -448,3 +448,49 @@ void Light::SetType(const LightType _type)
 {
   this->dataPtr->type = _type;
 }
+
+/////////////////////////////////////////////////
+bool Light::PopulateElement(sdf::ElementPtr _elem) const
+{
+  std::string lightTypeStr = "point";
```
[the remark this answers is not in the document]
iche033: b68f18

test/integration/sdf_dom_conversion.cc
```diff
@@ -402,3 +404,163 @@ TEST(SDFDomConversion, Sensors)
...
+    EXPECT_EQ(sdf::LightType::DIRECTIONAL, dirLight->Type());
+    EXPECT_EQ(ignition::math::Pose3d(0, 1, 2, 0, 0, 0),
+        dirLight->RawPose());
+    EXPECT_TRUE(dirLight->CastShadows());
+    EXPECT_EQ(ignition::math::Color(0.0f, 1.0f, 0.2f, 1.0f),
+        dirLight->Diffuse());
+    EXPECT_EQ(ignition::math::Color(0.0f, 0.2f, 0.1f, 1.0f),
+        dirLight->Specular());
+    EXPECT_EQ(ignition::math::Vector3d(4.0, 5.0, 6.0), dirLight->Direction());
+  }
```
azeey, line 565: I think the tests are good, but they don't cover the use case where an `sdf::Light` is created by the user *without* calling `sdf::Light::Load`. But the main advantage of `ToElement` is to be used in that scenario or when a DOM object is modified after `Load` is called. Otherwise, we would just use the existing `Element()` function.

iche033: something like a9fedbd64ccdcfaf26a54d449092c84e1501e816? The new test checks the values produced by `ToElement` without calling `Load` and also checks the case when values are modified then `ToElement` is called.

azeey (approved):

PR #779, opened 2021-12-09
Added ign fuel to USD converter

ahcorde:

# 🎉 New feature

## Summary

This PR will allow to convert models with the fuel models.

This adds a new dependency -> `ignition fuel tools7`

## Test it

Some of the Ignition fuel models.

## Checklist
- [ ] Added tests


ahcorde:
> @ahcorde I think this is the relative paths issue you were telling me about. What workaround/fix did you use?

the workaround is to run the converter inside the model folder

``` 
cd cd ~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/<model>/<version>/
usdConverter model.sdf  model.usd
```

src/CMakeLists.txt
```diff
@@ -245,6 +245,7 @@ target_link_libraries(${sdf_target}
     ignition-common${IGN_COMMON_VER}::ignition-common${IGN_COMMON_VER}
     ignition-common${IGN_COMMON_VER}::graphics
     ignition-utils${IGN_UTILS_VER}::ignition-utils${IGN_UTILS_VER}
+		ignition-fuel_tools${IGN_FUEL_TOOLS_VER}::ignition-fuel_tools${IGN_FUEL_TOOLS_VER}
```
adlarkin, line 248: nit: fix indentation

src/cmd/usdConverter.cc
```diff
@@ -19,9 +19,31 @@
 
 #include <pxr/usd/usd/stage.h>
 
+#include <ignition/fuel_tools/FuelClient.hh>
+#include <ignition/fuel_tools/Interface.hh>
+#include <ignition/common/Util.hh>
```
adlarkin, line 24: nit: header include order, and all in same group

src/cmd/usdConverter.cc
```diff
@@ -19,9 +19,31 @@
...
+
 #include "sdf/sdf.hh"
 #include "sdf_usd_parser/world.hh"
 
+
+std::unique_ptr<ignition::fuel_tools::FuelClient> fuelClient =
```
adlarkin, line 30: nit: include `<memory>` header for unique_ptr

src/cmd/usdConverter.cc
```diff
@@ -40,6 +68,8 @@ int main(int argc, const char* argv[])
     return -2;
   }
 
+  // std::cerr << root.Element()->ToString("") << std::endl;
+
```
adlarkin, line 72: remove these lines

src/usd/sdf_usd_parser/material.cc
```diff
@@ -19,7 +19,14 @@
...
 
+#include <ignition/common/URI.hh>
+#include <ignition/common/Filesystem.hh>
+#include <ignition/common/Util.hh>
+
+#include <ignition/fuel_tools/ClientConfig.hh>
+
 #include <ignition/math/Color.hh>
+
 #include <pxr/base/tf/stringUtils.h>
```
adlarkin, line 30: nit: header includes

adlarkin (approved):
> the workaround is to run the converter inside the model folder

Works for me, thanks! We will probably need to address the relative paths issue, but we can do that later on

chapulina (changes requested):
> This adds a new dependency -> ignition fuel tools7

We should keep these projects separate. Fuel is currently integrated with SDF through the sdf::findFileCallback](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/src/Server.cc#L91-L93).

adlarkin:
> > This adds a new dependency -> ignition fuel tools7
> 
> We should keep these projects separate. Fuel is currently integrated with SDF through the [sdf::findFileCallback](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/src/Server.cc#L91-L93). 

This is all a proof of concept right now, so we can change things if needed (notice how this is targeting a custom branch). We are trying to test if the SDF -> USD converter works with things like materials and meshes, so fuel models are a good way to do this.

Regarding keeping the dependencies separate, I agree that no new dependency is ideal, but the issue is that with this new USD work, we should ultimately provide a way for users to convert SDF files to USD. This is currently being done in the SDF repository (see the `src/cmd/usdConverter.cc` file in this PR). If we want to support fuel models (which are used a lot in SDF files), then I feel like we have no other option but to add this dependency if we decide to keep the SDF -> USD conversion tool in the SDF repository. It should also be worth noting that we are planning to make all of this work be in a component, so the dependency would be "optional".

However, after saying all of that, it would be nice to not have to depend on fuel at all, even if users opt-in to the USD component. I think this means that the SDF -> USD converter script would need to go somewhere other than the SDF repository. @chapulina, any thoughts about where this functionality may be better placed if the SDF repository is not the best place for it? If we were to remove this code from the SDF repository, it would seem strange to me that users would have to configure another repository/library (for example, `ign-gazebo`) just to get the functionality of converting SDF to USD. I feel like the appropriate place for this code is in the SDF repository, but I understand that it introduces the issue of additional dependencies.

chapulina (commented):
It's important to keep the separation of concerns in mind. Each library should be responsible for one task:

* The SDF-USD converter will convert SDF files to USD and vice-versa. 
* Ignition Fuel Tools fetches resources from a Fuel server.

>  If we want to support fuel models (which are used a lot in SDF files), then I feel like we have no other option but to add this dependency if we decide to keep the SDF -> USD conversion tool in the SDF repository. 

Instead of fetching resources from specific places, libSDFormat offers an API so that downstream packages can tell it where to look for resources. Some packages leverage environment variables, others leverage callbacks... That's by design so SDF isn't tied to some specific way of fetching resources.

The same way, the converter should support all user-defined environmnet variables and callbacks without forcing some specific lookup mechanism on downstream consumers.

> it would seem strange to me that users would have to configure another repository/library (for example, ign-gazebo) just to get the functionality of converting SDF to USD

The functionality would still be there, the difference is that the converter wouldn't be aware of Fuel models unless the user explicitly sets it up. That's true for libSDFormat in general, not just for the USD component.

> We are trying to test if the SDF -> USD converter works with things like materials and meshes, so fuel models are a good way to do this.

Agreed that it's convenient to use Fuel models to demonstrate the functionality. If testing with `ign-gazebo` isn't desired, you could create an example standalone program in the `examples` folder that shows how to use the converter with Fuel as an example of how a user can setup the integration.

> This is all a proof of concept right now, so we can change things if needed (notice how this is targeting a custom branch).

I think it's fine to do it as a proof of concept, just wanted to make sure you folks were aware of the constraints you'll need to operate with when merging this into a supported branch.

adlarkin:
Thanks @chapulina for the comment! That was all very helpful. Your proposal of leaving configuration up to the user makes sense. Since we are still in "proof of concept mode" and are operating on a strict deadline, I think that we will leave this as-is for now. But, when it's time to actually merge this into libSDFormat, we can take the approach of leaving configuration up to the user and/or adding something to the `examples` folder. I like the `examples` idea a lot because users don't need to figure out how to configure the API for fuel at all - they can just build the example and run it.

PR #784, opened 2021-12-13
Added equality operator to Light DOM

nkoenig:

# 🎉 New feature

## Summary

In `Ignition Gazebo` the following line requires equality operators for a component, so I've added an equality operator.

```
std::vector<Entity> lights = _ecm.EntitiesByComponents(components::ParentEntity(_entity), components::Light());
```

## Test it

Run the tests.

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/Light.hh
+++ b/include/sdf/Light.hh
@@ -273,6 +273,20 @@ namespace sdf
     /// \return SDF element pointer with updated light values.
     public: sdf::ElementPtr ToElement() const;
 
+    /// Return true if this light object has the same property values as the
+    /// provided light object.
+    /// \param[in] _light Light object to compare against.
+    /// \return True if this light object has the same property values as
+    /// the provided _light object, false otherwise.
+    public: bool operator==(const Light &_light) const;
+
+    /// Return true if this light object has different property values from the
+    /// provided light object.
+    /// \param[in] _light Light object to compare against.
+    /// \return True if this light object has different property values from the
+    /// provided light object.
+    public: bool operator!=(const Light &_light) const;
+
     /// \brief Allow Link::SetPoseRelativeToGraph or World::Load to call
     /// SetXmlParentName and SetPoseRelativeToGraph,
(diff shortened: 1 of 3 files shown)


src/Light_TEST.cc
```diff
@@ -283,6 +289,15 @@ TEST(DOMLight, CopyAssignmentAfterMove)
 
   EXPECT_EQ("light2", light1.Name());
   EXPECT_EQ("light1", light2.Name());
+
+  // light1 and light2 have different names, they should be equal.
```
chapulina, line 293:
```suggestion
  // light1 and light2 have different names, they shouldn't be equal.
```

include/sdf/Light.hh
```diff
@@ -273,6 +273,20 @@ namespace sdf
     /// \return SDF element pointer with updated light values.
     public: sdf::ElementPtr ToElement() const;
 
+    /// Return true if this light object has the same property values as the
+    /// provided light object.
```
chapulina, line 277: Maybe it's worth explicitly mentioning that the `ElementPtr` is not checked.

chapulina (approved):

azeey:
When calling `_ecm.EntitiesByComponents(components::ParentEntity(_entity), components::Light());`, my expectation is to get all the lights under the parent entity, but what this will do is only give the lights that match the default constructed light. Is that what we want? Should `component::Light` not be just an empty component?

nkoenig:
> When calling `_ecm.EntitiesByComponents(components::ParentEntity(_entity), components::Light());`, my expectation is to get all the lights under the parent entity, but what this will do is only give the lights that match the default constructed light. Is that what we want? Should `component::Light` not be just an empty component?

Ahh, good point. I'll decline this PR and take another approach.

chapulina:
> Should component::Light not be just an empty component?

* Yeah see https://github.com/ignitionrobotics/ign-gazebo/issues/376

PR #809, opened 2021-12-30
Fix parsing 'type' attibutes in plugins

chapulina:
# 🦟 Bug fix

* Critical bug identified in https://github.com/ignitionrobotics/ign-gazebo/pull/1264
* Bug introduced in https://github.com/ignitionrobotics/sdformat/pull/689

## Summary

The logic added to check for the `type` attribute has broken use-cases inside `ign-gazebo`, such as [this](https://github.com/ignitionrobotics/ign-gazebo/blob/fe335f9cd47e7f7b3ff35159babe28d9fd55670d/test/worlds/triggered_publisher.sdf#L9-L14):

```
    <plugin filename="ignition-gazebo-triggered-publisher-system" name="ignition::gazebo::systems::TriggeredPublisher">
      <input type="ignition.msgs.Empty" topic="/in_1"/>
      <output type="ignition.msgs.Boolean" topic="/out_1">
          data: true
      </output>
    </plugin>
```

Which now causes an assertion:

```
173: Error [Param.cc:768] Unknown parameter type[ignition.msgs.Boolean]
173: Exception [Param.cc:81] SDF ASSERTION                     
173: Invalid parameter
173: In function       : Param
173: Assert expression : this->dataPtr->ValueFromStringImpl( this->dataPtr->typeName, _default, this->dataPtr->defaultValue)
```

My assumption from #689 is that we only want to parse the `/pose@type` attributes here, but don't affect the rest, so I added the extra check for "pose". I left a TODO there because I think we should remove the special type handling from that function to a pose-specific function. 

I suspect I may be missing some important detail, but tests pass. So I hope this is enough as a band-aid to fix the regression, and I'd like to check with @aaronchongth and @azeey when they're back on making a proper fix.

## Checklist
- [x] Added tests


src/Plugin_TEST.cc
```diff
@@ -186,7 +186,7 @@ TEST(DOMPlugin, LoadWithChildren)
   std::string pluginStr = R"(<plugin name='3D View' filename='MinimalScene'>
   <ignition-gui>
     <title>3D View</title>
-    <property type='bool' key='showTitleBar'>false</property>
```
scpeters, line 189: I think the problem is related to unrecognized `type` values, like `ignition.msgs.Boolean` in [triggered_publisher.sdf](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/test/worlds/triggered_publisher.sdf#L11)

chapulina: I was just fixing the test failure. With my change, the parser is now converting `false` into `0`.

I haven't touched the parser much before, so I'd appreciate some guidance on the proper fix. My assumption from #689 is that we only want to parse the `/pose@type` attributes here, but don't affect the rest, that's why I added the extra check for `"pose"` with a TODO to remove the special type handling from that function to a pose-specific function. But I suspect I may be missing some important detail.

chapulina:
All tests are passing here, so I'm opening this for review. I updated the PR description with more context.

chapulina:
* The fix was also confirmed downstream: https://github.com/ignitionrobotics/ign-gazebo/pull/1266#issuecomment-1003172481

scpeters:
I've added a line in https://github.com/ignitionrobotics/sdformat/pull/809/commits/5fc1571415fdd1fe1ffbaf8ffc74a3b54a0c0b1d that reproduces the failure seen in ign-gazebo, caused by an unrecognized `type` attribute

scpeters:
I'd like to see what @aaronchongth thinks about this change

chapulina:
> I've added a line in 5fc1571 that reproduces the failure seen in ign-gazebo, caused by an unrecognized type attribute

It failed similar to the other line:

```
-    <property type='Ignition.Msgs.Boolean'>false</property>
+    <property type='Ignition.Msgs.Boolean'>0</property>
```

> I'd like to see what @aaronchongth thinks about this change

My intention here is to revert the regression without reverting #689 completely. I don't think this is the final fix, and I hope the original authors come back with a cleaner fix later. I'd just like to fix the broken use cases on `ign-gazebo` for from-source Fortress users and source/nightly Garden users.

Are you concerned we may be breaking something else with this, @scpeters ?

aaronchongth:
Hi @chapulina, @scpeters! Thanks for raising this, I am checking out what a good fix for this is at the moment.

The changes to `parser.cc` in https://github.com/ignitionrobotics/sdformat/pull/689, especially checking for the `type` attribute, was added specifically for this scenario in `Plugin_TEST.cc`. My bad for assuming that all the types will still be consistent with what `sdformat12` is used to.

I have a fix ready https://github.com/ignitionrobotics/sdformat/pull/811, let me know what you think. I haven't managed to run this fix against `ign-gazebo6` yet as it builds pretty slowly on my available laptop, but I should get some results in a few hours when it's done.

scpeters (approved):

aaronchongth (approved):

osrf-triage:

PR #811, opened 2021-12-31
Fix for custom element with attribute 'type'

aaronchongth:
Signed-off-by: Aaron Chong <aaronchongth@gmail.com>

# 🦟 Bug fix

Related to https://github.com/ignitionrobotics/sdformat/pull/809.

## Summary

Reverted changes to parser for custom element with attribute 'type', added fix in ParamPrivate::ValueFromStringImpl instead.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -1942,27 +1942,17 @@ void copyChildren(ElementPtr _sdf,
       ElementPtr element(new Element);
       element->SetParent(_sdf);
       element->SetName(elemName);
-      std::optional<std::string> typeName = std::nullopt;
+      if (elemXml->GetText() != nullptr)
+      {
+        element->AddValue("string", elemXml->GetText(), true);
+      }
+
       for (const tinyxml2::XMLAttribute *attribute = elemXml->FirstAttribute();
            attribute; attribute = attribute->Next())
       {
-        const std::string attributeName(attribute->Name());
-        // TODO(chapulina) This isn't a good place to parse pose-specific
-        // attributes
-        if (elemName == "pose" && attributeName == "type")
-          typeName = attribute->Value();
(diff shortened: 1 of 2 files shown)


src/parser.cc
```diff
@@ -1942,27 +1942,17 @@ void copyChildren(ElementPtr _sdf,
...
+      for (const tinyxml2::XMLAttribute *attribute = elemXml->FirstAttribute();
+           attribute; attribute = attribute->Next())
       {
-        if (typeName.has_value())
-          element->AddValue(typeName.value(), elemXml->GetText(), true);
-        else
-          element->AddValue("string", elemXml->GetText(), true);
+        element->AddAttribute(attribute->Name(), "string", "", 1, "");
+        element->GetAttribute(attribute->Name())->SetFromString(
+            attribute->Value());
```
scpeters, line 1955: it looks like you've reversed the order of some logic in parser.cc, calling `element->AddValue` before / after setting the attribute values. Does this order matter? If not, can you reverse it so that the diff will be easier to read?

scpeters (commented):
this does look like a better fix; I just have one question about the order of some changes

src/parser.cc
aaronchongth: This order does not matter as in this case the value is always added as type `string`.

Only with specific types like `pose`, where the attributes like `degrees` or `rotation_format` are read, will it matter. But it still makes sense to have attributes handled before the value, regardless.

Reversed in https://github.com/ignitionrobotics/sdformat/pull/811/commits/c630d3093775b04418eae131e2fbb9230a14dbd7.

scpeters (approved):
thank you!

PR #819, opened 2022-01-12
USD -> SDF: polygon triangulation

koonpeng:
# 🎉 New feature

## Summary

This implements the `PolygonToTriangles` function using the fan-triangulation algorithm. Initially tried to implement delaunay triangulation based on ignition::common's implementation, but the implementation is a wrapper around gts's implementation, which does not support 3D.

While researching a custom implementation, found out that there is a much simpler way using the fan-triangulation, it is not as robust as delaunay but it looks like a better fit for the use case of triangulating convex polygons. It seems like tools like blender also uses a form of fan-triangulation of these use cases and delaunay is more commonly used for more complex scenarios like generating terrains/heightmaps etc.

## Test it
Currently tested with `GearJoint.usd` in isaac sim's examples (should be able to find it with a file system search). It seems to work somewhat, but I'm not sure if the current output is due to bad triangulation or other bugs.

usd:

image

sdf:

image

## Checklist
- [ ] Added tests

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -96,7 +96,6 @@ else()
 endif()
 
 find_package(pxr REQUIRED)
-find_package(CGAL)
 
 include_directories(${PXR_INCLUDE_DIRS})
 set(sources ${sources}
@@ -265,7 +264,6 @@ target_link_libraries(${sdf_target}
     ignition-common${IGN_COMMON_VER}::graphics
     ${PYTHON_LIBRARY}
     ${PXR_LIBRARIES}
-    CGAL::CGAL
   PRIVATE
     ${TinyXML2_LIBRARIES})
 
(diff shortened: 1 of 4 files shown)


koonpeng:
It appears that gts already supports triangulating meshes, `gts_surface_strip` should do the trick. I will start that functionality to ignition common, after which, this PR will need the new version.

ahcorde (changes requested):
I think something is wrong here. 

 - I modified the `gearjoint.usd` inside `IssacSim` and I removed everything expect one of the cylinders and I saved this world
 - Then I converted this saved world with the `sdfconverter` and I get the following shape:

cylinder

If we forget for a moment the transformations, the mesh is not rendered properly.

koonpeng:
Opening the dae in blender at least looks like a cylinder

image

Will try to find out more about what is going on, maybe the way ign performs smoothing causes the distortion.

PR #831, opened 2022-01-26
sdf -> usd: Materials

ahcorde:
# 🎉 New feature

## Summary

Suppor for materials in the sdf to usd converter:
 - Conversor between sdf::Material and ignition::common::Materials
    - Added tests 
 - Added material parser
    - Added tests

## Checklist
- [x] Added tests


koonpeng (approved):
Can you put the link to the docs containing the TfTokens used?

ahcorde:

usd/src/Conversions.cc
```diff
@@ -0,0 +1,154 @@
...
+  std::shared_ptr<ignition::common::Material> convert(const sdf::Material &_in)
+  {
+    auto out = std::make_shared<ignition::common::Material>();
+    out->SetEmissive(_in.Emissive());
+    out->SetDiffuse(_in.Diffuse());
+    out->SetSpecular(_in.Specular());
+    out->SetAmbient(_in.Ambient());
+    out->SetRenderOrder(_in.RenderOrder());
+    out->SetLighting(_in.Lighting());
+    out->SetAlphaFromTexture(false, false, _in.DoubleSided());
```
adlarkin, line 100: According to the API, `SetAlphaFromTexture` takes a `double` for the second input parameter: https://github.com/ignitionrobotics/ign-common/blob/ignition-common4_4.5.0/graphics/include/ignition/common/Material.hh#L161-L166

Here, the second input parameter is a `bool`. The code still compiles because `false` implicitly converts to an int with a value of 0, which is then implicitly converted to a double of `0.0`. I don't think we want to rely on this implicit conversion, though. What if we used the default value for the second parameter (`0.5`) instead? That would make the call look like this:
```suggestion
    out->SetAlphaFromTexture(false, 0.5, _in.DoubleSided());
```

usd/src/Conversions_TEST.cc
```diff
@@ -0,0 +1,163 @@
...
+{
+  std::shared_ptr<ignition::common::Material> materialCommon =
+    std::make_shared<ignition::common::Material>();
+  materialCommon->SetEmissive(ignition::math::Color(1, 0.2, 0.2, 0.7));
+  materialCommon->SetDiffuse(ignition::math::Color(0.1, 0.3, 0.4, 0.5));
+  materialCommon->SetSpecular(ignition::math::Color(0.11, 0.22, 0.23, 0.77));
+  materialCommon->SetAmbient(ignition::math::Color(0.25, 0.21, 0.28, 0.5));
+  materialCommon->SetRenderOrder(5);
+  materialCommon->SetLighting(true);
+  materialCommon->SetAlphaFromTexture(false, false, true);
```
adlarkin, line 107: same comment

usd/src/sdf_parser/Material_Sdf2Usd_TEST.cc
```diff
@@ -0,0 +1,231 @@
...
+    std::string materialshaderPath = materialPath + "/Shader";
+    auto materialShaderPrim = this->stage->GetPrimAtPath(
+      pxr::SdfPath(materialshaderPath));
+    EXPECT_TRUE(materialShaderPrim);
+
+    CheckMaterial(materialShaderPrim,
+      pxr::GfVec3f(0.2, 0.5, 0.1),
+      pxr::GfVec3f(0, 0, 0),
+      "materials/textures/albedo_map.png",
+      "materials/textures/normal_map.png",
```
adlarkin, line 205: When I run this test, I see error messages for these files:
```
[Err] [SystemPaths.cc:467] Could not resolve file [albedo_map.png]
[Err] [SystemPaths.cc:467] Could not resolve file [normal_map.png]
```

Does this need to be addressed?

adlarkin (commented):
I made a few updates in https://github.com/ignitionrobotics/sdformat/pull/831/commits/7130c1fe863fcf0aacf6ecee0754385247ddd737, but I have not reviewed all of the files in this PR yet. Here are some questions that I have so far.

usd/src/sdf_parser/Material_Sdf2Usd_TEST.cc
ahcorde: These failures are expected we are not including these files.

usd/src/sdf_parser/Material.cc
```diff
@@ -0,0 +1,462 @@
...
+    const std::map<pxr::TfToken, pxr::VtValue> &_customData,
+    const pxr::TfToken &_displayName,
+    const pxr::TfToken &_displayGroup,
+    const std::string &_doc,
+    const pxr::TfToken &_colorSpace = pxr::TfToken(""))
+  {
+    auto shader = pxr::UsdShadeShader(_prim);
+    if (!shader)
+    {
+      std::cerr << "Not able to convert the prim to a UsdShadeShader\n";
```
adlarkin, line 109: Instead of printing to `sdf::cerr`, can we create a `UsdError` object and return that? We will probably need to modify the `CreateMaterialInput` API to return a `UsdErrors` object instead of returning `void`.

usd/include/sdf/usd/sdf_parser/Material.hh
```diff
@@ -0,0 +1,54 @@
...
+  namespace usd
+  {
+    /// \brief Parse an SDF material into a USD stage.
+    /// \param[in] _material The SDF material to parse.
+    /// \param[in] _stage The stage that should contain the USD representation
+    /// of _material.
+    /// \return The parsed usd material (invalid material if parsing failed)
+    pxr::UsdShadeMaterial IGNITION_SDFORMAT_USD_VISIBLE
+      ParseSdfMaterial(const sdf::Material *_material,
+        pxr::UsdStageRefPtr &_stage);
```
adlarkin, line 49: Related to my other comment, since this method makes calls to `CreateMaterialInput` which should return `UsdErrors`, can this method also be updated to return `UsdErrors`? This will also require updating code elsewhere that calls this method. I propose that the method signature be modified to this:

```suggestion
    /// \brief Parse an SDF material into a USD stage.
    /// \param[in] _materialSdf The SDF material to parse.
    /// \param[in] _stage The stage that should contain the USD representation
    /// of _material.
    /// \param[out] _materialUsd The parsed USD material representation of _materialSdf
    /// (invalid material if parsing failed)
    /// \return UsdErrors, which is a list of UsdError objects. This list is empty if no errors
    /// occurred when parsing _materialSdf to _materialUsd
    UsdErrors IGNITION_SDFORMAT_USD_VISIBLE
      ParseSdfMaterial(const sdf::Material *_materialSdf,
        pxr::UsdStageRefPtr &_stage, pxr::UsdShameMaterial &_materialUsd);
```

usd/src/sdf_parser/Material.cc
```diff
@@ -0,0 +1,462 @@
...
+      material = pxr::UsdShadeMaterial(usdMaterialPrim);
+    }
+
+    auto usdShader = pxr::UsdShadeShader::Define(
+      _stage,
+      pxr::SdfPath(mtl_path + "/Shader"));
+    auto shaderPrim = _stage->GetPrimAtPath(pxr::SdfPath(mtl_path + "/Shader"));
+    if (!shaderPrim)
+    {
+      std::cerr << "Not able to cast the UsdShadeShader to a Prim" << '\n';
```
adlarkin, line 182: Same comment applies here about changing `sdf::cerr` to `UsdError` (see https://github.com/ignitionrobotics/sdformat/pull/831#discussion_r816438465)

usd/src/sdf_parser/Geometry.cc
```diff
@@ -380,6 +383,24 @@ namespace usd
         pxr::GfVec3f(meshMax.X(), meshMax.Y(), meshMax.Z()));
       usdMesh.CreateExtentAttr().Set(extentBounds);
 
+      int materialIndex = subMesh->MaterialIndex();
+      if (materialIndex != -1)
```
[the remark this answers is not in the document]
ahcorde: uhmm, when there is no material the value is setted to -1 https://github.com/ignitionrobotics/ign-common/blob/ignition-common4_4.5.0/graphics/src/SubMesh.cc#L55

usd/src/cmd/sdf2usd.cc
```diff
@@ -51,8 +55,168 @@ struct Options
...
+    {
+      std::string current(*file);
+      if (ignition::common::isDirectory(current))
+      {
+        auto systemPaths = ignition::common::systemPaths();
+        systemPaths->AddFilePaths(current);
+        findFileByExtension(current, "", true);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
ahcorde: This part is required to insert all subdirectories inside a model folder.

usd/include/sdf/usd/sdf_parser/Material.hh
```diff
@@ -0,0 +1,60 @@
...
+  // Inline bracket to help doxygen filtering.
+  inline namespace SDF_VERSION_NAMESPACE {
+  //
+  namespace usd
+  {
+    /// \brief Parse an SDF material into a USD stage.
+    /// \param[in] _materialSdf The SDF material to parse.
+    /// \param[in] _stage The stage that should contain the USD representation
+    /// of _material.
+    /// \param[out] _materialUsd The parsed USD material representation of
```
scpeters, line 47: nit: doc-string variable name doesn't match `_materialUsd` vs `_materialPath`; let's pick one

usd/src/sdf_parser/Geometry.cc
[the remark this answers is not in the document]
scpeters: that looks like an inconsistency in the ign-common4 API. We should consider changing that for ign-common5: https://github.com/ignitionrobotics/ign-common/issues/317

usd/src/sdf_parser/Geometry.cc
```diff
@@ -380,6 +383,43 @@ namespace usd
...
+        const auto material = ignMesh->MaterialByIndex(materialIndex);
+        const sdf::Material materialSdf = sdf::usd::convert(material);
+        std::string materialPath;
+        UsdErrors materialErrors = ParseSdfMaterial(
+          &materialSdf, _stage, materialPath);
+        if (!materialErrors.empty())
+        {
+          errors.push_back(UsdError(
+            sdf::usd::UsdErrorCode::SDF_TO_USD_PARSING_ERROR,
+            "Unable to parsematerial"));
```
scpeters, line 398: nit: missing a space between words, capitalization optional

"Unable to parse Material" / "Unable to parse material"

usd/include/sdf/usd/Conversions.hh
```diff
@@ -0,0 +1,56 @@
...
+  //
+  namespace usd
+  {
+    /// \brief Specialized conversion from an Ignition Common Material
+    /// to a SDF material
+    /// \param[in] _in Ignition Common Material.
+    /// \return SDF material.
+    IGNITION_SDFORMAT_USD_VISIBLE
+    sdf::Material convert(
+        const std::shared_ptr<ignition::common::Material> &_in);
```
scpeters, line 43: is a `const std::shared_ptr<Material>` what we want to pass in here? this will still allow mutation of the underlying data. Can we pass in a `const` raw pointer or `std::shared_ptr<const Material>` instead?

usd/include/sdf/usd/Conversions.hh
```diff
@@ -0,0 +1,56 @@
...
+
+    /// \brief Specialized conversion from an SDF material to a Ignition Common
+    /// material.
+    /// \param[in] _in SDF material.
+    /// \return Ignition Common Material.
+    IGNITION_SDFORMAT_USD_VISIBLE
+    std::shared_ptr<ignition::common::Material> convert(
```
scpeters, line 50: why return a `shared_ptr` here? Is it important that the resulting `Material` be allocated on the heap? we return an `sdf::Material` object from the other function

ahcorde: not really sure about how to deal with this, because I can't return a `ignition::common::Material` because is a non-copyable class (it has a `unique_ptr` inside). I decided touse the same type in both methods. 

@scpeters what do you think is right type ?

scpeters: ok, that's reasonable. I'll open an issue in ign-common suggesting that we improve the Material class to make it copyable

usd/include/sdf/usd/Conversions.hh
scpeters: I do think the `const`-ness of this pointer could still be adjusted

usd/include/sdf/usd/Conversions.hh
azeey: If it has a move ctor, then you should be might be able to return `ignition::common::Material` and the compiler will use copy elision to construct the object at the call location.

ahcorde: it doesn't have a move ctor https://github.com/ignitionrobotics/ign-common/blob/ign-common4/graphics/include/ignition/common/Material.hh#L90-L97

@azeey, does it make sense to include it ? or can we keep the `shared_ptr` ?

azeey: Unfortunately, since it has a user declared destructor, the compiler will not generate a move ctor. Simply removing the destructor would fix this, but I think that will break ABI. So, returning a smart pointer seems to be our best bet, but if we have to do that, my preference would be to use a `unique_ptr` unless there is a particular reason to use a `shared_ptr`.

ahcorde: `SubMesh::MaterialByIndex` returns a `shared_ptr` which mean we need to convert this `shared_ptr` to `unique_ptr` to finally get the `sdf::Material`. I will bet for the `shared_ptr`to simplify things.

scpeters (approved):
I didn't go deeply into all the details here, but the API seems ok

usd/include/sdf/usd/Conversions.hh
adlarkin: I updated this API to use a `const ignition::common::Material *` in https://github.com/ignitionrobotics/sdformat/pull/831/commits/8b97a1c31c3714bd0f487d808058cb501ce4bae5, so a `shared_ptr` is no longer used.

usd/include/sdf/usd/Conversions.hh
adlarkin: I have an idea for getting around the `std::shared_ptr` issue - see https://github.com/ignitionrobotics/sdformat/pull/831/commits/3095a50c29269fa54a93db9d6355372a608fc4a2. If you guys think this isn't a good solution, feel free to revert this commit.

usd/src/sdf_parser/Geometry.cc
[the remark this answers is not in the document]
adlarkin: I added a `TODO` note in https://github.com/ignitionrobotics/sdformat/pull/831/commits/cdc0257c33be6aca984f3d98233a2c578d83a7c7 about fixing this when we merge forward to `sdf13`: https://github.com/ignitionrobotics/sdformat/blob/cdc0257c33be6aca984f3d98233a2c578d83a7c7/usd/src/sdf_parser/Geometry.cc#L386-L388 For now, the casting approach that's being done here is fine.

adlarkin (commented):
I made some changes in https://github.com/ignitionrobotics/sdformat/pull/831/commits/cdc0257c33be6aca984f3d98233a2c578d83a7c7 and https://github.com/ignitionrobotics/sdformat/pull/831/commits/8b97a1c31c3714bd0f487d808058cb501ce4bae5. I also tried to address the `std::shared_ptr` issue discussed in https://github.com/ignitionrobotics/sdformat/pull/831#discussion_r817076486 with https://github.com/ignitionrobotics/sdformat/pull/831/commits/3095a50c29269fa54a93db9d6355372a608fc4a2, so once someone looks at my workaround, I think this is good to go.

osrf-triage:

PR #837, opened 2022-01-28
sdf -> usd converter: add joints

adlarkin:
# 🎉 New feature

## Summary
Adds support for converting joints to their USD representation when parsing a SDF model to USD. Currently, the following joint types are supported:
* revolute
* ball
* fixed
* prismatic

## Test it
You can run the unit test that has been added (`sdformat/usd/src/sdf_parser/Joint_Sdf2Usd_TEST.cc`)

## Checklist
- [x] Added tests


scpeters:
can you resolve the conflicts?

ahcorde:

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+          parentToJoint.Rot().Z());
+
+    auto childRotation = pxr::GfQuatf(
+          childToJoint.Rot().W(),
+          childToJoint.Rot().X(),
+          childToJoint.Rot().Y(),
+          childToJoint.Rot().Z());
+
+    const auto axis = _joint.Axis();
+    if (axis && (axis->Xyz() == ignition::math::Vector3d::UnitY))
```
scpeters, line 130: this entire logic block seems strange to me. I'm sure there's a good reason, but it's not obvious. Can you add documentation to explain what this code is doing and why?

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+    }
+    else if (axis->Xyz() == ignition::math::Vector3d::UnitZ ||
+        axis->Xyz() == -ignition::math::Vector3d::UnitZ)
+    {
+      usdJoint.CreateAxisAttr().Set(pxr::TfToken("Z"));
+    }
+    else
+    {
+      errors.push_back(UsdError(sdf::Error(sdf::ErrorCode::ELEMENT_INVALID,
+            "Revolute joint [" + _joint.Name() + "] has an invalid axis.")));
```
scpeters, line 217: I wouldn't say that the axis is invalid, since SDFormat supports expressing an axis unit vector in any direction. I would say "Revolute joint [name] has specified an axis [x y z], but USD only supports integer values of [-1, 0, 1] when specifying joint axis unit vectors."

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+            "] as a UsdPhysicsDriveAPI."));
+      return errors;
+    }
+
+    // TODO(ahcorde) Review damping and stiffness values
+    // I added these values as a proof of concept.
+    double damping = axis->Damping();
+    if (ignition::math::equal(damping, 0.0))
+    {
+      damping = 35;
```
scpeters, line 245: I would recommend adding nonzero values of damping and stiffness to the [test/sdf/double_pendulum.sdf](https://github.com/ignitionrobotics/sdformat/blob/sdf12/test/sdf/double_pendulum.sdf) example rather than using hard-coded values here

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+    }
+    else if (axis->Xyz() == ignition::math::Vector3d::UnitZ ||
+        axis->Xyz() == -ignition::math::Vector3d::UnitZ)
+    {
+      usdJoint.CreateAxisAttr().Set(pxr::TfToken("Z"));
+    }
+    else
+    {
+      errors.push_back(UsdError(sdf::Error(sdf::ErrorCode::ELEMENT_INVALID,
+            "Prismatic joint [" + _joint.Name() + "] has an invalid axis.")));
```
scpeters, line 297: update this error message as I suggested in the other comment about revolute joints

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+    switch (_joint.Type())
+    {
+      case sdf::JointType::REVOLUTE:
+        parsingErrors = ParseSdfRevoluteJoint(_joint, _stage, _path);
+        break;
+      case sdf::JointType::BALL:
+        // While USD allows for cone limits that can restrict motion in a given
+        // range, SDF does not have limits for a ball joint. So, there's
+        // nothing to do after creating a UsdPhysicsSphericalJoint, since this
+        // joint by default has no limits (i.e., allows for circular motion)
```
scpeters, line 353: I think we could open an issue to add these cone limits to the spec

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+      return errors;
+    }
+
+    // define the joint's parent/child links
+    jointPrim.CreateBody0Rel().AddTarget(parentLinkPath);
+    jointPrim.CreateBody1Rel().AddTarget(childLinkPath);
+
+    const auto poseErrors =
+      SetUSDJointPose(jointPrim, _joint, _parentModel);
+    if (!poseErrors.empty())
```
scpeters, line 395: I think you don't need this `if` statement. You can just unconditionally call `errors.insert(...)` and it will be a no-op if `poseErrors` is empty

usd/src/sdf_parser/Joint.cc
scpeters: https://github.com/ignitionrobotics/sdformat/issues/860

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+
+    ignition::math::Pose3d parentToJoint;
+    if (_joint.ParentLinkName() == "world")
+    {
+      ignition::math::Pose3d modelToJoint;
+      auto poseErrors = usd::PoseWrtParent(_joint, modelToJoint);
+      if (!poseErrors.empty())
+        return poseErrors;
+
+      // it is assumed the _parentModel's parent is the world
```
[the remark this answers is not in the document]
adlarkin: The SDF -> USD converter does not support nested models, which is mentioned in #845. `sdf::usd::ParseSdfModel` also exits early if a nested model is attempted to be parsed: https://github.com/ignitionrobotics/sdformat/blob/5059bb8b8dd3e946bf84c0654ed073d32278e931/usd/src/sdf_parser/Model.cc#L53-L59

However, I guess that downstream users could still try to call the `sdf::usd::ParseSdfJoint` method directly and have unexpected behavior if they're using a nested model. What about I update the documentation of `sdf::usd::ParseSdfJoint` to mention that the model that `_parentModel`'s parent must be the world, and cannot be a nested model?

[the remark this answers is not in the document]
scpeters: I think you just update this comment to say: "it is assumed the _parentModel's parent is the world because nested models are not yet supported (see issue #845)"

ahcorde (commented):
I pushed some code to address the feedback

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+      auto poseResolutionErrors =
+        _joint.SemanticPose().Resolve(parentToJoint, _joint.ParentLinkName());
+      if (!poseResolutionErrors.empty())
+      {
+        errors.push_back(UsdError(
+              sdf::Error(sdf::ErrorCode::POSE_RELATIVE_TO_INVALID,
+              "Unable to get the pose of joint [" + _joint.Name() +
+              "] w.r.t. its parent link [" + _joint.ParentLinkName() + "].")));
+        for (const auto &e : poseResolutionErrors)
+          errors.push_back(UsdError(e));
```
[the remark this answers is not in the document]
adlarkin: This case is a little different than other error cases, because `poseResolutionErrors` holds a list of `sdf::Error` objects, but `errors` is of type `UsdError`. So, as you can see in the loop, we have to construct a `UsdError` object from the existing `sdf::Error` object. Since we have to create a different type of error object before adding it to the error list, I think we will need to keep this loop.

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,402 @@
...
+    auto poseResolutionErrors = _joint.SemanticPose().Resolve(childToJoint,
+        _joint.ChildLinkName());
+    if (!poseResolutionErrors.empty())
+    {
+      errors.push_back(UsdError(
+          sdf::Error(sdf::ErrorCode::POSE_RELATIVE_TO_INVALID,
+            "Unable to get the pose of joint [" + _joint.Name() +
+            "] w.r.t. its child [" + _joint.ChildLinkName() + "].")));
+      for (const auto &e : poseResolutionErrors)
+        errors.push_back(UsdError(e));
```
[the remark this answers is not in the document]
adlarkin: We will need to keep this loop since this case is different from other error cases; see https://github.com/ignitionrobotics/sdformat/pull/837#discussion_r814082594 for an explanation.

usd/src/sdf_parser/Joint.cc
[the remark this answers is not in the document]
scpeters: got it

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,413 @@
...
+    else if (axis->Xyz() == ignition::math::Vector3d::UnitZ ||
+        axis->Xyz() == -ignition::math::Vector3d::UnitZ)
+    {
+      usdJoint.CreateAxisAttr().Set(pxr::TfToken("Z"));
+    }
+    else
+    {
+      errors.push_back(UsdError(sdf::Error(sdf::ErrorCode::ELEMENT_INVALID,
+        "Revolute joint [" + _joint.Name() + "] has specified an axis "
+        "[x y z], but USD only supports integer values of [0, 1] when"
```
scpeters, line 234: I meant to suggest printing `axis->Xyz()` in place of `x y z` in this message

usd/src/sdf_parser/Joint.cc
```diff
@@ -0,0 +1,413 @@
...
+    else if (axis->Xyz() == ignition::math::Vector3d::UnitZ ||
+        axis->Xyz() == -ignition::math::Vector3d::UnitZ)
+    {
+      usdJoint.CreateAxisAttr().Set(pxr::TfToken("Z"));
+    }
+    else
+    {
+      errors.push_back(UsdError(sdf::Error(sdf::ErrorCode::ELEMENT_INVALID,
+        "Prismatic joint [" + _joint.Name() + "] has specified an axis "
+        "[x y z], but USD only supports integer values of [0, 1] when"
```
scpeters, line 307: repeat: I meant to suggest printing `axis->Xyz()` in place of `x y z` in this message

scpeters (commented):
only two minor comments remaining

I still don't get the Y-axis revolute joint issue, but there are todo comments there, so we can come back to it at some point

usd/src/sdf_parser/Joint.cc
adlarkin: https://github.com/ignitionrobotics/sdformat/pull/837/commits/c03c363adaa6e469a66baab012c1ad6a0c981dbc

usd/src/sdf_parser/Joint.cc
adlarkin: https://github.com/ignitionrobotics/sdformat/pull/837/commits/c03c363adaa6e469a66baab012c1ad6a0c981dbc

usd/src/sdf_parser/Joint.cc
adlarkin: I noticed that there's a `TODO` note here now. When doing original testing on a prototype a few months ago, @ahcorde found that joints along the y axis had to be handled differently in USD due to the USD axis conventions. I do not recall the specifics of the fix, but since there's a TODO note, I will mark this as resolved and we can revisit it later.

adlarkin:
> I still don't get the Y-axis revolute joint issue, but there are todo comments there, so we can come back to it at some point

Sounds good, @ahcorde may have some more info about this - I left a comment with some more background/context: https://github.com/ignitionrobotics/sdformat/pull/837#discussion_r814323859

All other feedback should be addressed now, so this is ready for final review.

scpeters (approved):

osrf-triage:

PR #842, opened 2022-02-02
Root to element

nkoenig:
# 🎉 New feature

## Summary

Adds `Root::ToElement`, similar to the other `ToElement` functions. This does change the API of `World::ToElement`, which I think is okay because sdf12 has not been released with the original `World::ToElement` function.

## Test it

Run the tests.

## Checklist
- [x] Added tests


azeey:
Can you resolve the conflicts?

nkoenig:
> Can you resolve the conflicts?

Resolved.

include/sdf/Root.hh
```diff
@@ -183,6 +198,13 @@ namespace sdf
     /// an error code and message. An empty vector indicates no error.
     public: Errors UpdateGraphs();
 
+    /// \brief Create and return an SDF element filled with data from this
+    /// root.
+    /// \param[in] _useIncludeTag This will pass the _useIncludeTag to
+    /// sdf::Model::ToElement.
+    /// \return SDF element pointer with updated root values.
+    public: sdf::ElementPtr ToElement(bool _useIncludeTag = true) const;
```
azeey, line 206: I have a feeling this will need more parameters at some point. Something like the `sdf::PrintConfig` class or an enum would work, but we can leave it for future work if you don't have time for it now.

src/Model.cc
```diff
@@ -961,7 +961,11 @@ sdf::ElementPtr Model::ToElement(bool _useIncludeTag) const
           "relative_to")->Set<std::string>(this->dataPtr->poseRelativeTo);
     }
     includeElem->GetElement("static")->Set(this->Static());
-    includeElem->GetElement("placement_frame")->Set(this->PlacementFrameName());
+    if (!this->dataPtr->placementFrameName.empty())
+    {
+      includeElem->GetElement("placement_frame")->Set(
+          this->PlacementFrameName());
+    }
```
azeey, line 968: How do we tackle plugins? If the plugin was originally added using the include tag, adding it here makes sense, but if the plugin was part of the model, then adding it here would be problematic.

azeey (changes requested):
Looks great! My main question is regarding plugins

src/Model.cc
nkoenig: Line 1015 outputs all of the plugins contained by this Model. It does the following:

```
  // Add in the plugins
  for (const Plugin &plugin : this->dataPtr->plugins)
    elem->InsertElement(plugin.ToElement(), true);
```

Do you see a problem with that logic?

include/sdf/Root.hh
nkoenig: A separate class might be needed, but I'd rather add the functionality when I have concrete use case.

src/Model.cc
azeey: The case I'm thinking of is this:
Say you have model "robot"
```xml
<model name="robot">
   ...
   <plugin name='"plugin_1" ... />
</model>
```

And that model is included in a world

```xml
<world name="default">
  <include>
     <uri>path/to/robot</uri>
     <plugin name="plugin_2" .../>
  </include>
</world>
```

Currently, if `_useIncludeTag = true` when we call `ToElement()`, the resulting `ElementPtr` will have `plugin_1`, but not `plugin_2`.

nkoenig: I see, thanks. I'll look into it.

nkoenig: Handled this case in [bfbad45](https://github.com/ignitionrobotics/sdformat/pull/842/commits/bfbad45b7678d172cdbddacbb498d06c3bdcb59f)

nkoenig:
> Since using `ToElement(true)` won't capture the effects of the [param passing](http://sdformat.org/tutorials?tut=param_passing_tutorial&cat=specification&) functionality, it would be good to mentiona that in `ToElement`'s documentation.

I've add documentation to the `ToElement` functions.

> 
> One overall observation is that we have recently made efforts to avoid printing default elements and attributes (see #575) when calling `sdf::Element::ToString`. However, this currently doesn't work if the `ElementPtr` is obtained using `ToElement`. Should I ticket an issue?

Yes please.

azeey:
Ticketed https://github.com/ignitionrobotics/sdformat/issues/867

osrf-triage:

PR #844, opened 2022-02-07
Change Root to non-unique impl pointer

nkoenig:

# 🎉 New feature

## Summary

Change Root from a unique impl pointer to a non-unique impl pointer. This will allow the Root object to be copyable.

## Checklist
- [x] Added tests


include/sdf/Root.hh
```diff
@@ -152,7 +152,7 @@ namespace sdf
     public: sdf::ElementPtr Element() const;
 
     /// \brief Private data pointer
-    IGN_UTILS_UNIQUE_IMPL_PTR(dataPtr)
+    IGN_UTILS_IMPL_PTR(dataPtr)
```
azeey, line 155: We'll need to define the copy ctor and copy assignment operator to properly handle frame graphs.

azeey (changes requested):

include/sdf/Root.hh
scpeters: to verify that frame graphs work, we would need a test that loads an `sdf::Root` from an SDFormat string or file and then copies it and tries to resolve a pose or attached-to frame to confirm that the graphs are properly copied

nkoenig: I think I'm missing something here. I added tests to capture frame copying, but the tests just passed without requiring the definition of a copy constructor and copy assignment operator. See [6b64370](https://github.com/gazebosim/sdformat/pull/844/commits/6b6437074cd8522543cc20152dc914e0743d1ce9).

src/Frame.cc
```diff
@@ -217,3 +218,18 @@ sdf::ElementPtr Frame::Element() const
...
+{
+  sdf::ElementPtr elem(new sdf::Element);
+  sdf::initFile("frame.sdf", elem);
+
+  elem->GetAttribute("name")->Set(this->dataPtr->name);
+
+  if (!this->dataPtr->attachedTo.empty())
+    elem->GetAttribute("attached_to")->Set(this->dataPtr->attachedTo);
+
+  elem->GetElement("pose")->Set(this->dataPtr->pose);
```
scpeters, line 233: I recommend copying the code from `Link::ToElement` that handles the relative_to attribute:

https://github.com/gazebosim/sdformat/blob/main/src/Link.cc#L655-L662

include/sdf/Root.hh
scpeters: you're right; I commented this before I saw your latest commit; this is fine

src/Frame.cc
nkoenig: Done in 76d3794

scpeters (approved):

PR #818, opened 2022-01-08
Add sdf -> usd light functionality to usd component

adlarkin:
Signed-off-by: Ashton Larkin <42042756+adlarkin@users.noreply.github.com>

# 🎉 New feature

## Summary
This builds on top of #817 to add light converting functionality from SDF to USD.

There are a few things that may need to be addressed, but they aren't blocking the core functionality (I left some TODO notes in the code to highlight things we might want to address). I think we can start to review this PR now, and then re-visit the TODOs later when time permits to add enhanced functionality. Perhaps we should create an issue that lists what enhanced functionality is missing from the USD -> SDF creator?

I also think we can close #796 because the only test in that PR is for lights, which has been moved to this PR (we will need to address https://github.com/ignitionrobotics/sdformat/pull/796#pullrequestreview-842491424, which is also a potential issue with the tests that are written in #817).

## Test it
You can run the unit test that has been added in this PR (`Light_Sdf2Usd_TEST`), or you can try converting https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/examples/worlds/lights.sdf to a USD file with the `sdf2usd` executable that's generated in #817 (this assumes you have PXR installed and have built the `usd` component of sdformat).

If you want to see how the light intensities look in isaac sim, you'll need to add an object to the `lights.sdf` file above in order to have the lights reflect off of something. You can add the following code to `usd/src/World.cc`, which adds a ground plane to the world:
```diff
diff --git a/usd/src/World.cc b/usd/src/World.cc
index a3aaa138..90bb81f7 100644
--- a/usd/src/World.cc
+++ b/usd/src/World.cc
@@ -30,6 +30,16 @@
 #include "sdf/World.hh"
 #include "sdf/usd/Light.hh"
 
+// includes needed for creating a plane
+#include <pxr/usd/usdGeom/xform.h>
+#include <ignition/math/Vector3.hh>
+#include <ignition/math/Pose3.hh>
+#include <pxr/usd/usdGeom/cube.h>
+#include <pxr/base/gf/vec3f.h>
+#include <pxr/base/vt/array.h>
+#include <pxr/usd/usdGeom/xformCommonAPI.h>
+#include "sdf/usd/Utils.hh"
+
 namespace usd
 {
   bool ParseSdfWorld(const sdf::World &_world, pxr::UsdStageRefPtr &_stage,
@@ -69,6 +79,23 @@ namespace usd
               << "the moment. Models that are children of the world "
               << "are currently being ignored.\n";
 
[...]

PR #850, opened 2022-02-12
Hide the USD component's Util.hh file from public API

adlarkin:
Signed-off-by: Ashton Larkin <42042756+adlarkin@users.noreply.github.com>

# 🦟 Bug fix

## Summary
When I originally worked on #818, I created a set of helper methods for achieving generic parsing tasks between SDF and USD. These methods were placed in `usd/include/sdf/usd/sdf_parser/Utils.hh`. After #818 was merged, I realized that the `Utils.hh` is meant for internal use only, and should not be exposed to downstream users. So, this PR moves `Utils.hh` to `usd/src`, and also updates the methods in this file to make use of the `UsdError` class that was introduced in #836.

_It should be noted that the ABI checker may fail since I am removing methods from the public API. However, I talked with others offline, and they said that for a scenario like this, it's okay because the public API I am changing in the USD component has not been released yet. So, we will have to ignore the ABI checker for this PR._

## Checklist
- [ ] Added tests


scpeters (commented):
I wouldn't expect the ABI checker to fail since that CI job does not have the `pxr` dependency installed or built from source

also, it looks like this isn't simply a move of the header file; it looks like you made some changes to some of the methods in Utils.hh as well. This could be fine, but you should explain them in the pull request description

adlarkin:
> I wouldn't expect the ABI checker to fail since that CI job does not have the `pxr` dependency installed or built from source

You're right, ABI checker didn't fail :+1: 

> also, it looks like this isn't simply a move of the header file; it looks like you made some changes to some of the methods in Utils.hh as well. This could be fine, but you should explain them in the pull request description

Yes, you are right, there are some slight changes. They're in the PR description if you look at the end of the first paragraph:
```
also updates the methods in this file to make use of the UsdError class that was introduced in https://github.com/ignitionrobotics/sdformat/pull/836.
```

Let me know if you think any other descriptions/documentation needs to be added.

adlarkin:
@scpeters @ahcorde I renamed the file from `Utils.hh` to `UsdUtils.hh` and added tests for the methods in this file in https://github.com/ignitionrobotics/sdformat/pull/850/commits/33beba4e38809343423965fa7d26a2381b8f5c3a. This should be ready for review if you wouldn't mind taking a look.

ahcorde:

ahcorde (approved):

usd/src/UsdUtils.hh
```diff
@@ -45,38 +47,62 @@ namespace sdf
   //
   namespace usd
   {
-    /// \brief Get an object's pose w.r.t. its parent.
+    /// \brief Set an object's pose w.r.t. its parent.
```
scpeters, line 50: is this really `Set`? It seems more like it's getting the pose, since `_pose` is an output of this function

scpeters (approved):

osrf-triage:

PR #847, opened 2022-02-09
Updated tests for ign-math's ostream fix

jennuine:
# 🦟 Bug fix

- Requires https://github.com/ignitionrobotics/ign-math/pull/376

## Summary
With changes from the above ign-math PR, the output stream does not change the value. This change affected how [`magneticField`](https://github.com/ignitionrobotics/sdformat/blob/1df5df20817f6f05ade84369020be3a334257c44/src/World.cc#L74-L75) value is printed:

```c++
math::Vector3d magneticField(5.5645e-6, 22.8758e-6, -42.3884e-6);
cout << magneticField;

// before printed:   6e-06 2.3e-05 -4.2e-05
// after ign-math PR: 5.5645e-06 2.28758e-05 -4.23884e-05
```

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #854, opened 2022-02-18
Switch to utils version of env functions

mjcarroll:
Since we depend on utils, we can use the functionality from there.


ahcorde (approved):

test/env.hh
```diff
@@ -1,139 +0,0 @@
...
-             bool _allowEmpty)
-    {
-      std::string v;
-      bool valid = false;
-#ifdef _WIN32
-      // Unused on Windows, suppress warning
-      (void) _allowEmpty;
-      const DWORD buffSize = 32767;
-      static char buffer[buffSize];
-      if (GetEnvironmentVariable(_name.c_str(), buffer, buffSize))
```
azeey, line 57: The implementation of `env` here is different from `ign-utils`. This might be what's causing the test failures.

azeey (commented):
Some of the CI failures should be fixed by #847, but the Console_TEST failures come from this PR. My guess is it's the difference between `GetEnvironmentVariable` and `getenv_s` in Windows.

PR #863, opened 2022-02-25
USD -> SDF: Parse Physics and generated SDF file

ahcorde:
# 🎉 New feature

## Summary

Parse Physics and generated SDF file

## Test it

If you have the isaac-sim-assets-2021.1.1 assets you can try for example:

```bash
usd2sdf isaac-sim-assets-2021.1.1//Robots/UR10/ur10.usd ur10.sdf
```

The expected output must look like:

```xml
<sdf version="1.7">
    <world name="ur10_world">
          <gravity>0 0 -9.8</gravity>
    </world>
</sdf>
```

## Checklist
- [ ] Added tests


scpeters:
> ```shell
> usd2sdf isaac-sim-assets-2021.1.1//Robots/UR10/ur10.usd ur10.sdf
> ```
> 
> The expected output must look like:
> 
> ```
> <sdf version="1.7">
>     <world name="ur10_world">
>         <physics>
>             <gravity>0 0 -9.8</gravity>
>         </physics>
>     </world>
> </sdf>
> ```

this example doesn't look right to me. `<gravity>` should be directly under the `<world>`; it is not under `<physics>` in version 1.6+. In 1.5 and earlier, gravity was in the physics block, but not in newer versions.

* https://github.com/ignitionrobotics/sdformat/blob/sdf12/sdf/1.6/1_5.convert

ahcorde:
> > ```shell
> > usd2sdf isaac-sim-assets-2021.1.1//Robots/UR10/ur10.usd ur10.sdf
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The expected output must look like:
> > ```
> > <sdf version="1.7">
> >     <world name="ur10_world">
> >         <physics>
> >             <gravity>0 0 -9.8</gravity>
> >         </physics>
> >     </world>
> > </sdf>
> > ```
> 
> this example doesn't look right to me. `<gravity>` should be directly under the `<world>`; it is not under `<physics>` in version 1.6+. In 1.5 and earlier, gravity was in the physics block, but not in newer versions.
> 
> * https://github.com/ignitionrobotics/sdformat/blob/sdf12/sdf/1.6/1_5.convert

Thank you @scpeters , I updated the PR

usd/src/usd_parser/USDWorld.cc
```diff
@@ -0,0 +1,76 @@
...
+        if (!data.second)
+        {
+          errors.push_back(UsdError(UsdErrorCode::INVALID_PRIM_PATH,
+                "Unable to retrieve the pxr::UsdPhysicsScene named ["
+                + prim.GetPath().GetName() + "]"));
+          return errors;
+        }
+
+        ParseUSDPhysicsScene(prim, _world, data.second->MetersPerUnit());
+        continue;
```
adlarkin, line 69: I believe this should be `break` instead of `continue`, since we have found and parsed the USD physics scene

usd/src/usd_parser/USDWorld.cc
```diff
@@ -0,0 +1,76 @@
...
+    }
+    _world->worldName = reference->GetDefaultPrim().GetName().GetText();
+
+    auto range = pxr::UsdPrimRange::Stage(reference);
+    for (auto const &prim : range)
+    {
+      if (prim.IsA<pxr::UsdPhysicsScene>())
+      {
+        std::pair<std::string, std::shared_ptr<USDStage>> data =
+          usdData.FindStage(prim.GetPath().GetName());
```
adlarkin, line 59: Can `prim.GetPath().GetName()` can be simplified to `prim.GetName()`?

usd/src/usd_parser/USDWorld.cc
```diff
@@ -0,0 +1,76 @@
...
+        if (!data.second)
+        {
+          errors.push_back(UsdError(UsdErrorCode::INVALID_PRIM_PATH,
+                "Unable to retrieve the pxr::UsdPhysicsScene named ["
+                + prim.GetPath().GetName() + "]"));
```
adlarkin, line 64: Same question here about simplifying to `prim.GetName()`

usd/src/usd_parser/usd_model/WorldInterface.hh
```diff
@@ -0,0 +1,61 @@
...
+
+namespace sdf
+{
+  // Inline bracke to help doxygen filtering.
+  inline namespace SDF_VERSION_NAMESPACE {
+  //
+  namespace usd
+  {
+    /// \brief This class store data about the world
+    class WorldInterface {
```
adlarkin, line 38: Is there a particular reason why the file that defines this class is in a `usd_model` subdirectory? What if we put this file in the `usd/src/usd_parser/` directory with the other source files?

adlarkin (commented):
I still have a few more files to review, but here are a few questions that I have so far.

usd/src/usd_parser/USDWorld.cc
ahcorde: In this loop we will parse the rest of the elements: Links, sensors, plugins, etc. I perfer to keep `continue`

usd/src/usd_parser/usd_model/WorldInterface.hh
ahcorde: I just want to differentiate that these classes will create a model of the USD, they are not part of the parser.

adlarkin (commented):
I made final iterations in https://github.com/ignitionrobotics/sdformat/pull/863/commits/318eff97b9c1459683b304cdeef87c68f1014336. Overall, LGTM, but I don't know much about `tinyxml2`, so @azeey or @scpeters, would you mind taking one final look before we merge this?

usd/src/cmd/usd2sdf.cc
```diff
@@ -15,11 +15,11 @@
  *
  */
 
-#include <iostream>
```
azeey, line 18: Don't we still need `iostream`?

usd/src/usd_parser/USD2SDF.hh
```diff
@@ -0,0 +1,80 @@
...
+  {
+    /// \brief USD to SDF converter
+    /// This class helps generate the SDF file
+    class USD2SDF
+    {
+      /// \brief constructor
+      public: USD2SDF();
+
+      /// \brief destructor
+      public: ~USD2SDF();
```
azeey, line 43: Destructor can be removed

usd/src/usd_parser/USD2SDF.hh
```diff
@@ -0,0 +1,80 @@
...
+// Inline bracket to help doxygen filtering.
+inline namespace SDF_VERSION_NAMESPACE {
+  namespace usd
+  {
+    /// \brief USD to SDF converter
+    /// This class helps generate the SDF file
+    class USD2SDF
+    {
+      /// \brief constructor
+      public: USD2SDF();
```
azeey, line 40: Constructor can be removed, or use `USD2SDF() = default`.

usd/src/usd_parser/Parser.cc
```diff
@@ -0,0 +1,44 @@
...
+namespace usd
+{
+  UsdErrors parseUSDFile(
+      const std::string &_inputFilenameUsd,
+      const std::string &_outputFilenameSdf)
+  {
+    UsdErrors errors;
+    USD2SDF usd2sdf;
+    auto doc = tinyxml2::XMLDocument(true, tinyxml2::COLLAPSE_WHITESPACE);
+    auto readErrors = usd2sdf.Read(_inputFilenameUsd, &doc);
```
azeey, line 33: nit: This can just be `errors` and you won't need to copy it over below.

usd/src/usd_parser/USD2SDF.cc
```diff
@@ -0,0 +1,144 @@
...
+  {
+    errors.emplace_back(UsdError(
+      UsdErrorCode::SDF_TO_USD_PARSING_ERROR,
+      "Error parsing usd file [" + _fileName + "]"));
+    return errors;
+  }
+
+  tinyxml2::XMLElement *sdf = nullptr;
+  sdf = _sdfXmlOut->NewElement("sdf");
+  sdf->SetAttribute("version", "1.7");
```
azeey, line 58: Any reason to hardcode the SDFormat version? We usually use `SDF_PROTOCOL_VERSION`

usd/src/usd_parser/USD2SDF.cc
```diff
@@ -0,0 +1,144 @@
...
+  const auto errorsParseUSD = parseUSDWorld(_fileName, worldInterface);
+  if (!errorsParseUSD.empty())
+  {
+    errors.emplace_back(UsdError(
+      UsdErrorCode::SDF_TO_USD_PARSING_ERROR,
+      "Error parsing usd file [" + _fileName + "]"));
+    return errors;
+  }
+
+  tinyxml2::XMLElement *sdf = nullptr;
```
azeey, line 56: I see that the general approach is to create a tinyxml2 document and manually populate the elements in this function. I suspect this will be harder to scale than first converting to an `sdf::ElementPtr` and calling `ToString`. Even better might be to convert to `sdf::Root`, then call `sdf::Root::ToElement()` then `ToString`. This is much less error prone, handles all the formatting. It's also a lot more future proof. With the current approach, if we update the spec, we have to remember to update this function.

usd/src/usd_parser/USDPhysics.cc
```diff
@@ -0,0 +1,62 @@
...
+{
+inline namespace SDF_VERSION_NAMESPACE {
+namespace usd
+{
+  void ParseUSDPhysicsScene(
+    const pxr::UsdPhysicsScene &_scene,
+    std::shared_ptr<WorldInterface> &_world,
+    double _metersPerUnit)
+  {
+    if (const auto gravityAttr = _scene.GetGravityDirectionAttr())
```
azeey, line 36: Does `UsdAttribute` implicitly convert to `bool`?

usd/src/usd_parser/USDPhysics.hh
```diff
@@ -0,0 +1,50 @@
...
+  //
+  namespace usd
+  {
+    /// \brief Parse the physics attributes of a USD file
+    /// \param[in] _scene USD physics scene to extract attributes from
+    /// \param[out] _world World interface where the data is placed
+    /// \param[in] _metersPerUnit meters per unit in the USD
+    void IGNITION_SDFORMAT_USD_VISIBLE ParseUSDPhysicsScene(
+      const pxr::UsdPhysicsScene &_scene,
+      std::shared_ptr<WorldInterface> &_world,
```
azeey, line 44: By making this a `shared_ptr`, we are constraining the type of input. i.e, if the user has a `unique_ptr`, they wouldn't be able to use this function. If we know that users will always have a `shared_ptr` when they use this function, then this is okay, otherwise, I suggest using `WorldInterface &` as the type instead. Then the user can just dereference whichever smart pointer they're using when they call this function.

usd/src/usd_parser/usd_model/WorldInterface.hh
```diff
@@ -0,0 +1,60 @@
...
+namespace sdf
+{
+  // Inline bracket to help doxygen filtering.
+  inline namespace SDF_VERSION_NAMESPACE {
+  //
+  namespace usd
+  {
+    /// \brief This class stores data about the world
+    class WorldInterface {
+      public:
```
azeey, line 38: nit: Can you update the formatting for this class. `public: ` on each member. If all the members are going to be public, then maybe a `struct` is better?.

azeey (changes requested):

usd/src/usd_parser/USDPhysics.cc
ahcorde: yes

adlarkin (commented):
Thank you for the feedback and iterations, @ahcorde and @azeey! The new changes look good, and I think it's much easier to follow with `sdf::Root`. I just have one small comment, and will let @azeey take one last look before we merge.

usd/src/usd_parser/USDPhysics.hh
```diff
@@ -0,0 +1,49 @@
...
+{
+  // Inline bracket to help doxygen filtering.
+  inline namespace SDF_VERSION_NAMESPACE {
+  //
+  namespace usd
+  {
+    /// \brief Parse the physics attributes of a USD file
+    /// \param[in] _scene USD physics scene to extract attributes from
+    /// \param[out] _world World interface where the data is placed
+    /// \param[in] _metersPerUnit meters per unit in the USD
```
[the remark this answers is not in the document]
ahcorde: without `IGNITION_SDFORMAT_USD_VISIBLE` the test is not compiling

usd/src/usd_parser/USDPhysics.hh
```diff
@@ -0,0 +1,49 @@
...
+    void IGNITION_SDFORMAT_USD_VISIBLE ParseUSDPhysicsScene(
```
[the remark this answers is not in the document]
ahcorde: This function can't return any error.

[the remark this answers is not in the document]
azeey: You'll need to compile the appropriate `.cc` file along with the test like https://github.com/ignitionrobotics/sdformat/blob/ed04483dae428166ea548cc25d0f7c80d5e28995/src/CMakeLists.txt#L61-L88

usd/src/usd_parser/USDPhysics.hh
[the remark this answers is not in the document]
azeey: You'll need to compile the appropriate `.cc` file along with the test like https://github.com/ignitionrobotics/sdformat/blob/ed04483dae428166ea548cc25d0f7c80d5e28995/src/CMakeLists.txt#L61-L88

usd/src/usd_parser/USDPhysics.hh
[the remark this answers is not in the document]
azeey: Okay, I guess since this is private API, we can change our minds later if we need to.

usd/src/usd_parser/USDPhysics.hh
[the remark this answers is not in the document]
ahcorde: https://github.com/ignitionrobotics/sdformat/pull/863/commits/eae0182a29fd39e441c5f42f6d781f172dcf0223

usd/src/usd_parser/USD2SDF.cc
```diff
@@ -0,0 +1,63 @@
...
+
+  const auto addWorldErrors = _root.AddWorld(sdfWorld);
+  if (!addWorldErrors.empty())
+  {
+    for (const auto & error: addWorldErrors)
+    {
+      errors.emplace_back(error);
+    }
+    errors.emplace_back(UsdError(
+      UsdErrorCode::SDF_ERROR,
```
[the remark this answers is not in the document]
ahcorde: https://github.com/ignitionrobotics/sdformat/pull/863/commits/eae0182a29fd39e441c5f42f6d781f172dcf0223

osrf-triage:

PR #871, opened 2022-03-08
usd -> sdf: Read transforms

ahcorde:

# 🎉 New feature

## Summary

These functions allow to read the transform of a prim. It allows to read the tranforms from a prim to its parents and it should stop when the name given to the functions is equal to the name of the parent.

## Checklist
- [ ] Added tests


usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,261 @@
...
+      // to make them match the scaled values.
+      if (!_tfs.empty()) {
+        auto& child = _tfs.back();
+        child.Pos().Set(
+          child.Pos().X() * t.scale[0],
+          child.Pos().Y() * t.scale[1],
+          child.Pos().Z() * t.scale[2]);
+      }
+
+      if (!t.isRotationZYX)
```
koonpeng, line 73: Should we handle other rotation order?

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,261 @@
...
+      {
+        if (t.isRotation)
+        {
+          pose.Rot() = t.q[0];
```
koonpeng, line 77: Assuming that is means rotationXYZ, do we need to decompose the order of rotation to 3 tfs (similar to ZYX case)?

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,261 @@
...
+          scale[0] = scaleTmp[0];
+          scale[1] = scaleTmp[1];
+          scale[2] = scaleTmp[2];
+        }
+        t.scale = ignition::math::Vector3d(scale[0], scale[1], scale[2]);
+      }
+      else if (op == "xformOp:rotateZYX" || op == "xformOp:rotateXYZ")
+      {
+        pxr::GfVec3f rotationEuler(0, 0, 0);
+        auto attribute = _prim.GetAttribute(pxr::TfToken("xformOp:rotateZYX"));
```
koonpeng, line 161: This if block handles `rotateZYX` and `rotateXYZ`, but this only consider `rotateZYX`.

usd/src/usd_parser/USDTransforms_TEST.cc
```diff
@@ -0,0 +1,194 @@
...
+  for (unsigned int i = 0; i < _rotation.size(); ++i)
+  {
+    EXPECT_EQ(_rotation[i], usdTransforms.q[i]);
+  }
+
+  EXPECT_EQ(!_rotation.empty(), usdTransforms.isRotation);
+}
+
+/////////////////////////////////////////////////
+TEST(Utils, GetTransform)
```
koonpeng, line 66: Notice that there is no tests on rotation, I'm suspecting some test cases like `rotateXYZ` will fail.

koonpeng (dismissed):

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,268 @@
...
+        else if (attribute.GetTypeName().GetCPPTypeName() == "GfVec3d")
+        {
+          pxr::GfVec3d rotationEulerTmp(0, 0, 0);
+          attribute.Get(&rotationEulerTmp);
+          rotationEuler[0] = rotationEulerTmp[0];
+          rotationEuler[1] = rotationEulerTmp[1];
+          rotationEuler[2] = rotationEulerTmp[2];
+        }
+        ignition::math::Quaterniond qX, qY, qZ;
+        ignition::math::Angle angleX(rotationEuler[0] * IGN_PI / 180.0);
```
azeey, line 185: nit:
```suggestion
        ignition::math::Angle angleX(IGN_DTOR(rotationEuler[0]));
```

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,119 @@
...
+    /// \brief This function will parse all the parents transforms of a prim
+    /// This will stop when the name of the parent is the same as _schemaToStop
+    /// \param[in] _prim Initial prim to read the transform
+    /// \param[in] _usdData USDData structure to get info about the prim, for
+    /// example: metersperunit
+    /// \param[out] _tfs A vector with all the transforms
+    /// \param[out] _scale The scale of the prims
+    /// \param[in] _schemaToStop Name of the prim where the loop will stop
+    /// reading transforms
+    void IGNITION_SDFORMAT_USD_VISIBLE GetAllTransforms(
```
azeey, line 87: `GetAllTransforms` and `GetTransform` look like they're helper functions. Do they need to be `VISIBLE`?

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,119 @@
...
+  // Inline bracke to help doxygen filtering.
+  inline namespace SDF_VERSION_NAMESPACE {
+  //
+  namespace usd
+  {
+    /// \brief This class stores the transforms of a schema
+    /// This might contains scale, translate or rotation operations
+    /// The booleans are used to check if there is a transform defined
+    /// in the schema
+    /// Rotation is splitted in a vector because this might be defined
```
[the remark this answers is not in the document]
ahcorde: I prefer to keep this as 3 rotations II think it's more clear.

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#ifndef SDF_USD_USD_PARSER_UTILS_HH_
+#define SDF_USD_USD_PARSER_UTILS_HH_
```
adlarkin, line 19:
```suggestion
#ifndef SDF_USD_USD_PARSER_TRANSFORMS_HH_
#define SDF_USD_USD_PARSER_TRANSFORMS_HH_
```

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+#undef __DEPRECATED
+#include <pxr/usd/usdGeom/gprim.h>
+#pragma pop_macro ("__DEPRECATED")
+
+#include "sdf/sdf_config.h"
+#include "sdf/system_util.hh"
+#include "sdf/usd/Export.hh"
+#include "sdf/usd/UsdError.hh"
+
+#include "USDData.hh"
```
adlarkin, line 44:
```suggestion
#include "sdf/usd/UsdError.hh"
#include "USDData.hh"
```
nit: remove whitespace

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+
+namespace sdf
+{
+  // Inline bracke to help doxygen filtering.
```
adlarkin, line 48:
```suggestion
  // Inline bracket to help doxygen filtering.
```

usd/include/sdf/usd/usd_parser/USDTransforms.hh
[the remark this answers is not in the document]
adlarkin: As a follow-up to this, it looks like the `RotationXYZ` and `RotationZYX` methods of this class can be used to determine the order of stored rotations. Is this a correct assumption, @ahcorde?

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+      public: UDSTransforms();
+
+      /// \brief Translate
+      public: ignition::math::Vector3d Translate();
+
+      /// \brief Scale
+      public: ignition::math::Vector3d Scale();
+
+      /// \brief Rotation
+      public: std::vector<ignition::math::Quaterniond> Rotations();
```
adlarkin, line 71: Can more detail be added to the documentation of these methods? If they are getters, it would be nice to imply that these methods get a value. Also, if they are getters, then all of these methods can be marked `const`.

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+
+      /// \brief Set translate
+      public: void SetTranslate(const ignition::math::Vector3d &_translate);
+
+      /// \brief Set scale
+      public: void SetScale(const ignition::math::Vector3d &_scale);
```
adlarkin, line 77: nit: I believe that our style guide says to avoid using the `Set` prefix for setters

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+
+      /// \brief Add rotation
+      public: void AddRotation(const ignition::math::Quaterniond &_q);
```
adlarkin, line 80: nit: add documentation for the input parameter, `_q`

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+
+      /// \brief True if there is a rotation ZYX defined or false otherwise
+      bool RotationZYX();
+
+      /// \brief True if there is a rotation XYZ defined or false otherwise
+      bool RotationXYZ();
+
+      /// \brief True if there is a rotation (as a quaterion) defined
+      /// or false otherwise
+      bool Rotation();
```
adlarkin, line 90: These methods can be marked as `const`, since they are getters

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+
+      /// \brief Set if there is any rotation ZYX defined
+      void SetRotationZYX(bool _rotationZYX);
+
+      /// \brief Set if there is any rotation XYZ defined
+      void SetRotationXYZ(bool _rotationXYZ);
+
+      /// \brief Set if there is any rotation defined
+      void SetRotation(bool _rotation);
```
adlarkin, line 99: A few things about these methods:
* I believe that our style guide says to avoid using the `Set` prefix for setters
* can documentation be added for the input parameters?
* It would be nice to add documentation that outlines the difference between `Rotation` and `RotationZYX`, and the difference between `Rotation` and `RotationXYZ`. For example, is could `Rotation` be `false` and `RotationXYZ`/`RotationZYX` be `true`? Or is `Rotation` always true if `RotationXYZ` or `RotationZYX` is true?

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+
+      /// \brief Private data pointer.
+      IGN_UTILS_IMPL_PTR(dataPtr)
+    };
+
+    /// \brief This function will parse all the parents transforms of a prim
```
adlarkin, line 105:
```suggestion
    /// \brief This function will parse all the parent transforms of a prim.
```

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+    /// This will stop when the name of the parent is the same as _schemaToStop
+    /// \param[in] _prim Initial prim to read the transform
+    /// \param[in] _usdData USDData structure to get info about the prim, for
+    /// example: metersperunit
+    /// \param[out] _tfs A vector with all the transforms
```
adlarkin, line 110: How are the transforms in this vector ordered? Does it start from `_prim` and work up the tree towards `_schemaToStop`, or does it start at `_schemaToStop` and work its way down to `_prim`?

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+    /// \param[out] _scale The scale of the prims
+    /// \param[in] _schemaToStop Name of the prim where the loop will stop
+    /// reading transforms
+    void GetAllTransforms(
+      const pxr::UsdPrim &_prim,
+      USDData &_usdData,
```
adlarkin, line 116: This parameter is documented as `param[in]`. Can it be marked as `const`? If it needs to be mutable, can the documentation be updated to mark this parameter as `param[out]`?

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+      std::vector<ignition::math::Pose3d> &_tfs,
+      ignition::math::Vector3d &_scale,
+      const std::string &_schemaToStop);
+
+    /// \brief This function get the transform from a prim to the specified
+    /// schemaToStop variable
```
adlarkin, line 122:
```suggestion
    /// \brief This function gets the transform from a prim to the specified
    /// _schemaToStop variable
```

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+    /// This will stop when the name of the parent is the same as _schemaToStop
```
adlarkin, line 123: nit: this line of documentation isn't needed, I think you can delete it

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+    /// \param[in] _prim Initial prim to read the transform
+    /// \param[in] _usdData USDData structure to get info about the prim, for
+    /// example: metersperunit
+    /// \param[out] _pose Pose of the prim
```
adlarkin, line 127: Can you document what this pose is relative to? Is the pose expressed from `_schemaToStop` to `_prim`, or from `_prim` to `_schemaToStop`?

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+    /// \param[out] _scale The scale of the prim
+    /// \param[in] _schemaToStop Name of the prim where the loop will stop
+    /// reading transforms
+    void IGNITION_SDFORMAT_USD_VISIBLE GetTransform(
+      const pxr::UsdPrim &_prim,
+      USDData &_usdData,
```
adlarkin, line 133: This parameter is documented as `param[in]`. Can it be marked as `const`? If it needs to be mutable, can the documentation be updated to mark this parameter as `param[out]`?

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,146 @@
...
+      ignition::math::Pose3d &_pose,
+      ignition::math::Vector3d &_scale,
+      const std::string &_schemaToStop);
+
+    /// \brief Read the usd prim transforms. Scale, rotation or transform might
+    /// be defined as float or doubles
+    /// \param[in] _prim Prim where the transforms are read
+    UDSTransforms IGNITION_SDFORMAT_USD_VISIBLE ParseUSDTransform(
```
adlarkin, line 141: Can documentation be added for the return variable?

adlarkin (commented):
I have some initial comments/questions about the public header file.

usd/src/usd_parser/USDTransforms_TEST.cc
```diff
@@ -0,0 +1,195 @@
...
+  pxr::UsdPrim prim = _stage->GetPrimAtPath(pxr::SdfPath(_primName));
+  ASSERT_TRUE(prim);
+
+  sdf::usd::UDSTransforms usdTransforms =
+    sdf::usd::ParseUSDTransform(prim);
+
+  EXPECT_EQ(_translation, usdTransforms.Translate());
+  EXPECT_EQ(_scale, usdTransforms.Scale());
+  EXPECT_EQ(_rotation.size(), usdTransforms.Rotations().size());
+  for (unsigned int i = 0; i < _rotation.size(); ++i)
```
adlarkin, line 54: Can you add a check that the two vectors being compared here are of the same size to avoid out of bounds access errors?
```suggestion
  ASSERT_EQ(_rotation.size(), usdTransforms.Rotations().size());
  for (unsigned int i = 0; i < _rotation.size(); ++i)
```

usd/src/usd_parser/USDTransforms_TEST.cc
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/////////////////////////////////////////////////
+TEST(Utils, GetAllTransform)
+{
+  std::string filename = sdf::testing::TestFile("usd", "upAxisZ.usda");
+  auto stage = pxr::UsdStage::Open(filename);
+  ASSERT_TRUE(stage);
+
+  sdf::usd::USDData usdData(sdf::testing::TestFile("usd", "upAxisZ.usda"));
```
adlarkin, line 177:
```suggestion
  sdf::usd::USDData usdData(fileName);
```

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,395 @@
...
+  {
+    ignition::math::Pose3d poseUpAxis = ignition::math::Pose3d(
+      ignition::math::Vector3d(0, 0, 0),
+      ignition::math::Quaterniond(IGN_PI_2, 0, 0));
+    _tfs.push_back(poseUpAxis);
+  }
+}
+
+//////////////////////////////////////////////////
+void GetTransform(
```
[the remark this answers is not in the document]
ahcorde: This method is used in the follow up PRs

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,395 @@
...
+        t.Translate() * metersPerUnit,
+        ignition::math::Quaterniond(1, 0, 0, 0));
+
+      _tfs.push_back(poseZ);
+      _tfs.push_back(poseY);
+      _tfs.push_back(poseX);
+      _tfs.push_back(poseT);
+    }
+    parent = parent.GetParent();
+  }
```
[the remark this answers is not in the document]
ahcorde: I know it's weird, but I tried world with these rotations and it works like that. Maybe add a TODO or NOTE just in case someone see something weird ?

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,395 @@
...
+    pose.Pos() = t.Translate() * metersPerUnit;
+    // scaling is lost when we convert to pose, so we pre-scale the
+    // translation to make them match the scaled values.
+    if (!_tfs.empty()) {
+      auto& child = _tfs.back();
+      child.Pos().Set(
+        child.Pos().X() * t.Scale()[0],
+        child.Pos().Y() * t.Scale()[1],
+        child.Pos().Z() * t.Scale()[2]);
+    }
```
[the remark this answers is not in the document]
ahcorde: This part was done by @koonpeng maybe he can add something here

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,268 @@
...
+      {
+        if (t.isRotation)
+        {
+          pose.Rot() = t.q[0];
+        }
+        _tfs.push_back(pose);
+      }
+      else
+      {
+        ignition::math::Pose3d poseZ = ignition::math::Pose3d(
```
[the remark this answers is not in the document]
azeey: I see this is marked as resolved, but I don't see any changes to this effect in the code. Could you elaborate?

usd/src/usd_parser/USDTransforms.cc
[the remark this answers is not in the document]
azeey: This might be related to my previous comment. This seems really strange. Is there a test showing that this works regardless of the rotation order?

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,395 @@
...
+    parent = parent.GetParent();
+  }
+
+  if (upAxis == "Y")
+  {
+    ignition::math::Pose3d poseUpAxis = ignition::math::Pose3d(
+      ignition::math::Vector3d(0, 0, 0),
+      ignition::math::Quaterniond(IGN_PI_2, 0, 0));
+    _tfs.push_back(poseUpAxis);
+  }
```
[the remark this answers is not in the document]
azeey: How was this resolved?

usd/src/usd_parser/USDTransforms_TEST.cc
```diff
@@ -0,0 +1,200 @@
...
+    "/shapes/sun",
+    stage,
+    ignition::math::Vector3d(0, 0, 10),
+    {
+      ignition::math::Quaterniond(1, 0, 0, 0),
+      ignition::math::Quaterniond(1, 0, 0, 0),
+      ignition::math::Quaterniond(1, 0, 0, 0)
+    },
+    ignition::math::Vector3d(1, 1, 1)
+  );
```
azeey, line 172: I would like to see a test for objects with non-trivial rotations (with non-zero values for roll pitch and yaw).

usd/include/sdf/usd/usd_parser/USDTransforms.hh
```diff
@@ -0,0 +1,150 @@
...
+    /// Rotation is splitted in a vector because this might be defined
+    /// as a rotation of 3 angles (ZYX, XYZ, etc).
+    class IGNITION_SDFORMAT_USD_VISIBLE UDSTransforms
+    {
+      /// \brief Default constructor
+      public: UDSTransforms();
+
+      /// \brief Translate
+      /// \return A 3D vector with the translation
+      public: const ignition::math::Vector3d Translate() const;
```
azeey, line 65: I think it would be better to call this `Translation`. It would be consistent with `Rotation`.

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,409 @@
...
+
+  /// \brief True if there is a rotation ZYX defined or false otherwise
+  public: bool isRotationZYX = false;
+
+  /// \brief True if there is a rotation XYZ defined or false otherwise
+  public: bool isRotationXYZ = false;
+
+  /// \brief True if there is a rotation (as a quaterion) defined
+  /// or false otherwise
+  public: bool isRotation = false;
```
azeey, line 65: Instead of having an extra variable that the user has to set, could we just check whether the vector `q` is empty? Having to keep this variable in sync with the contents of `q` is, IMO, bugprone, especially since the user can set this variable via a member function without actually adding anything to `q`.

azeey (changes requested):

usd/src/usd_parser/USDTransforms.cc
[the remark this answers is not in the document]
ahcorde: I added this comment 

`// Add additional rotation to match with Z up Axis`

usd/include/sdf/usd/usd_parser/USDTransforms.hh
ahcorde: I reverted this change because I need to set `FindStage(...) const` this will change the API, but this was not already released, maybe I can change the signature of this method. What do you think @azeey ?

ahcorde:
@azeey and @adlarkin I made the requested changes even the rotation one, maybe I will need to open fix in the future or the sdf2usd converter is wrong. Not really sure about this

The only change that is still pending is the one related with the `const USDData _usdData`. I need to change the signature of `FindStage`, but this will break ABI, this module was not released yet, maybe I can still change it

usd/src/usd_parser/USDTransforms.cc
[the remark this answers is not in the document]
koonpeng: afaik, this transform is to convert the up axis from Y to Z. If `upAxis` is already Z there is no need to add the extra transform. Technically there could be a `upAxis = X` but I don't think any "normal" renderer uses that.

ahcorde:

usd/src/usd_parser/USDTransforms.cc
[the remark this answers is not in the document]
azeey: Okay, thanks for the explanation.

usd/src/usd_parser/USDTransforms_TEST.cc
azeey: Thanks for adding the test with non-trivial rotation. However, it still doesn't test how `RotationXYZ` and `RotationZYX` are different. If you converted this `.usda` file to SDFormat, what would be the `<pose>` of the `ellipsoid`? How about if you had `rotateZYX ` instead of `rotateXYZ` in the `.usda` file? Euler angle conventions are always confusing so it's best to iron it out before going forward.

adlarkin:
@koonpeng would you mind taking a look at https://github.com/ignitionrobotics/sdformat/pull/871#discussion_r833865606 and giving some feedback?

usd/src/usd_parser/USDTransforms.cc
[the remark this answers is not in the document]
koonpeng: iirc, the last pose's scale is being used as the value for the model's scale.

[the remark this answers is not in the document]
adlarkin: Thanks for the explanation. It sounds like we are assuming that the scale is defined in a parent prim of whatever we are transforming, but is this a valid assumption? For example, what if in the example you just gave, `TransformChild` has `translate[10,0,0]` and also has `scale[10,0,0]`? In this case, I believe the result should be `[100,0,0]`, but we would instead have this saved as `[10,0,0]` because we looked for the scale in the parent instead of the child.

usd/src/usd_parser/USDTransforms.cc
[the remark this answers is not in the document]
adlarkin: stored rotation order is now dependent on `xyz` and `zyx`, so I'll mark this as resolved: https://github.com/ignitionrobotics/sdformat/blob/0198bfa913496bcf184f48375294a3025df33bfb/usd/src/usd_parser/USDTransforms.cc#L211-L222

usd/src/usd_parser/USDTransforms.cc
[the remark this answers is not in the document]
adlarkin: I left a `TODO` note about handling `upAxis == "X"` in https://github.com/ignitionrobotics/sdformat/pull/871/commits/e2acde3f93cc5be5bb136851cc8f813e6753d430 in case we decide to revisit this later, but it sounds like handling `Y` and `Z` are sufficient for now. I will mark this as resolved.

usd/src/usd_parser/USDTransforms.cc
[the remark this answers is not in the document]
koonpeng: Checking the code again, the pose and scaling is computed by multiplying each children transform. So given, `TransformChild` with `translate[10,0,0], scale[2,0,0]` and `TransformParent` with `translate[20,0,0],scale[10,0,0]`. The final pose will be

ParentPose * ChildPose = translate[30,0,0] (or is it Child * Parent? Can't remeber the order of operations in pose multipliciations)
ChildScale * ParentScale = scale[20,0,0]

Given a point [0,0,0], that will end up as [30,0,0] which is wrong. What should have happened is TransformChild moved it to [10,0,0], then TransformParent moves it to [120,0,0]. The scaling applied on the intermediate point is lost because we separated the calculation of the pose and scaling.

adlarkin:
I have reviewed the changes in this PR and have a few proposals, which I listed in a new PR: https://github.com/ignitionrobotics/sdformat/pull/924

cc @ahcorde @azeey

ahcorde:
> I have reviewed the changes in this PR and have a few change proposals, which I listed in a new PR: #924
> 
> cc @ahcorde @azeey

As I mentioned in this comment https://github.com/ignitionrobotics/sdformat/pull/924#pullrequestreview-927188541 The suggestions to this PR https://github.com/ignitionrobotics/sdformat/pull/924 are breaking the rotations

The gripper has a wrong rotation

Selection_061

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,339 @@
...
+      ignition::math::Angle angleZ(IGN_DTOR(rotationEuler[2]));
+      qX = ignition::math::Quaterniond(angleX.Normalized().Radian(), 0, 0);
+      qY = ignition::math::Quaterniond(0, angleY.Normalized().Radian(), 0);
+      qZ = ignition::math::Quaterniond(0, 0, angleZ.Normalized().Radian());
+
+      if (op == kXFormOpRotateZYX)
+      {
+        std::swap(angleX, angleZ);
+      }
+      t.SetRotation((qX * qY) * qZ);
```
azeey, line 261: Since ign-math already does Euler angles, my recommendation would be to use that instead of creating 3 quaternions here. I think this can be simplified as:
```
// SDFormat/ign-math uses extrinsic XYZ rotations, so it's the same as `rotateXYZ`
ignition::math::Quaterniond rot(IGN_DTOR(rotationEuler[0]), IGN_DTOR(rotationEuler[1]), IGN_DTOR(rotationEuler[2]))
  if (op == kxFormOpRotateZYX){
    rot.Invert();
  }
  t.SetRotation(rot);

```

See https://graphics.pixar.com/usd/release/api/class_usd_geom_xformable.html#details for USD rotation order

See http://sdformat.org/tutorials?tut=specify_pose&cat=specification& for SDFormat rotation order.

usd/src/usd_parser/USDTransforms.cc
```diff
@@ -0,0 +1,339 @@
...
+      ignition::math::Quaterniond qX, qY, qZ;
+      ignition::math::Angle angleX(IGN_DTOR(rotationEuler[0]));
+      ignition::math::Angle angleY(IGN_DTOR(rotationEuler[1]));
+      ignition::math::Angle angleZ(IGN_DTOR(rotationEuler[2]));
+      qX = ignition::math::Quaterniond(angleX.Normalized().Radian(), 0, 0);
+      qY = ignition::math::Quaterniond(0, angleY.Normalized().Radian(), 0);
+      qZ = ignition::math::Quaterniond(0, 0, angleZ.Normalized().Radian());
+
+      if (op == kXFormOpRotateZYX)
+      {
```
azeey, line 258: The angles are swapped here, but the angles are not used after this line, so it doesn't affect the end result. See my recommendation above.

usd/src/usd_parser/USDTransforms_TEST.cc
```diff
@@ -0,0 +1,221 @@
...
+    ASSERT_TRUE(usdTransforms.Rotation());
+    EXPECT_TRUE(ignition::math::equal(
+      _rotation.value().Roll(),
+       usdTransforms.Rotation().value().Roll(), 0.1));
+   EXPECT_TRUE(ignition::math::equal(
+     _rotation.value().Pitch(),
+      usdTransforms.Rotation().value().Pitch(), 0.1));
+    EXPECT_TRUE(ignition::math::equal(
+      _rotation.value().Yaw(),
+       usdTransforms.Rotation().value().Yaw(), 0.1));
```
azeey, line 66: This is fine. I would probably lower the tolerance a bit. If you want, you could just do
`EXPECT_TRUE(_rotation->Equal(*usdTransforms.Rotation(), tol));`

azeey (approved):

PR #875, opened 2022-03-09
USD -> SDF: Added lights attached to the world

ahcorde:

# 🎉 New feature

## Summary

USD -> SDF: Added lights attached to the world

## Checklist
- [ ] Added tests


test/usd/upAxisZ.usda
```diff
@@ -13,6 +13,23 @@ def "shapes"
...
+        kind = "model"
+    )
+    {
+        float angle = 1
+        float intensity = 3000
+        float shaping:cone:angle = 180
+        float shaping:cone:softness
+        float shaping:focus
+        color3f shaping:focusTint
+        asset shaping:ies:file
```
koonpeng, line 27: Is it intended that some of these attributes are undefined?

usd/src/usd_parser/USD2SDF.cc
```diff
@@ -75,6 +81,96 @@ UsdErrors USD2SDF::Read(const std::string &_filename,
   return errors;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+void USD2SDF::AddLights(
```
koonpeng, line 85: I'm confused about the role of this function.
* It looks like it is modifying the xml directly, shouldn't we use sdf apis instead?
* It looks like usd is already parsed into sdf by this point, why do we need this function?

usd/src/usd_parser/USDLight_TEST.cc
```diff
@@ -0,0 +1,59 @@
...
+
+  auto stage = pxr::UsdStage::Open(filename);
+  ASSERT_TRUE(stage);
+
+  pxr::UsdPrim prim = stage->GetPrimAtPath(pxr::SdfPath("/shapes/defaultLight"));
+  ASSERT_TRUE(prim);
+
+  auto light = sdf::usd::ParseUSDLights(
+    prim, usdData, "/World/defaultLight");
+  ASSERT_TRUE(light);
```
koonpeng, line 58: Add some checks on the output?

usd/src/usd_parser/USDWorld.cc
```diff
@@ -50,10 +62,65 @@ namespace usd
...
     {
+      // Skip materials, the data is already available in the USDData class
+      if (prim.IsA<pxr::UsdShadeMaterial>() || prim.IsA<pxr::UsdShadeShader>())
+      {
+        continue;
+      }
+
       std::string primName = prim.GetName();
+      std::string primPath = pxr::TfStringify(prim.GetPath());
+      std::string primType = pxr::TfStringify(prim.GetPath());
```
koonpeng, line 78: It seems strange to me that `primPath` and `primType` are identical, maybe there is a mistake here?

usd/src/usd_parser/USDWorld.cc
```diff
@@ -50,10 +62,65 @@ namespace usd
...
+
+      // In general USD models used in Issac Sim define the model path
+      // under a root path for example:
+      //  -> /robot_name/robot_name_link0
+      // But sometimes for enviroments it uses just a simple path:
+      //  -> /ground_plan
+      //  -> /wall_0
+      // the shortName variable defines if this is the first case when it's
+      // False or when it's true then it's the second case.
+      if (primPathTokens.size() >= 2)
```
koonpeng, line 91: The usd input may not come from isaac sim and so it may not be following isaac sim convention. We should probably note somewhere that the USD to SDF converter only works for isaac sim USDs.

koonpeng (changes requested):

koonpeng (commented):
code LGTM, but I got different results when converting sdf -> usd -> sdf

original:
```
    <light type="directional" name="sun">
      <cast_shadows>true</cast_shadows>
      <pose>0 0 10 0 0 0</pose>
      <diffuse>0.8 0.8 0.8 1</diffuse>
      <specular>0.8 0.8 0.8 1</specular>
      <attenuation>
        <range>1000</range>
        <constant>0.9</constant>
        <linear>0.01</linear>
        <quadratic>0.001</quadratic>
      </attenuation>
      <direction>-0.5 0.1 -0.9</direction>
    </light>
```

usda
```
    def DistantLight "sun"
    {
        float inputs:intensity = 100
        float intensity = 100
        float3 xformOp:rotateXYZ = (0, -0, 0)
        double3 xformOp:translate = (0, 0, 10)
        uniform token[] xformOpOrder = ["xformOp:translate", "xformOp:rotateXYZ"]
    }
```

converted
```
    <light name='sun' type='directional'>
      <pose>0 0 10 0 -0 0</pose>
      <cast_shadows>true</cast_shadows>
      <intensity>0.01</intensity>
      <direction>0 0 -1</direction>
      <diffuse>1 1 1 1</diffuse>
      <specular>1 1 1 1</specular>
      <attenuation>
        <range>10</range>
        <linear>1</linear>
        <constant>1</constant>
        <quadratic>0</quadratic>
      </attenuation>
      <spot>
        <inner_angle>0</inner_angle>
        <outer_angle>0</outer_angle>
        <falloff>0</falloff>
      </spot>
    </light>
```

koonpeng (approved):
> There is some data that is not common between these two formats, I would say that some of the data might be "lost in translation"

I assume that these differences are due to the different formats and can't be addressed easily?

usd/src/usd_parser/USDWorld.cc
```diff
@@ -67,10 +76,66 @@ namespace usd
...
+              linkName = "/" + primPathTokens[0];
+              shortName = true;
+            }
+          }
+        }
+        if (!shortName)
+        {
+          linkName = "/" + primPathTokens[0] + "/" + primPathTokens[1];
+        }
+      }
```
adlarkin, line 124: Is this code mainly used for finding the parent prim of the light? I find it a little odd that we are manually parsing through the first 2 levels of the prim path to find the parent. I imagine this could be problematic if the light's parent isn't in the first two levels of the stage. Is there a better way to find the parent of a light that offers more flexibility/is less error prone?

adlarkin (commented):
I have a few questions before merging. Otherwise, I think this looks good.

usd/src/usd_parser/USDWorld.cc
ahcorde: I didn't find a better way to do it, I'm open to suggestions

adlarkin: I can't think of a better way at the moment - since we are pressed for time, I added a TODO note in https://github.com/ignitionrobotics/sdformat/pull/875/commits/847271a1821043f64cefab13c0c7219a43d6cb6e and created an issue in #927.

PR #876, opened 2022-03-10
USD -> SDF: Added model tag

ahcorde:

# 🎉 New feature

## Summary

USD -> SDF: Added model tag

## Checklist
- [ ] Added tests


usd/src/usd_parser/USDWorld.cc
```diff
@@ -80,6 +87,25 @@ namespace usd
       std::vector<std::string> primPathTokens =
         ignition::common::split(primPath, "/");
 
+      if (primPathTokens.size() == 1 && !prim.IsA<pxr::UsdGeomCamera>())
```
koonpeng, line 90: This is making an assumption on the scene graph of the usd, it wouldn't hold if the usd does not come from isaac sim. We should probably note somewhere that the converter only works for isaac sim usds.

koonpeng (changes requested):

adlarkin:
@ahcorde I made some changes and added tests, would you mind taking a look? Here are the changes I made: https://github.com/ignitionrobotics/sdformat/pull/876/files/af2ce4c6caf2cf1a7cf7be2214de94d4a1d03800..553b93b03bbd482dd0602a2c066547aaee125a0f

usd/src/usd_parser/USDLight_TEST.cc
```diff
@@ -54,7 +54,7 @@ TEST(USDLightsTest, DistanceLight)
   ASSERT_TRUE(stage);
 
   pxr::UsdPrim prim = stage->GetPrimAtPath(
-    pxr::SdfPath("/shapes/defaultLight"));
+    pxr::SdfPath("/defaultLight"));
```
adlarkin, line 57: Since we no longer have the `shapes` top level prim in the testing file, does the `schemaToStop` in the corresponding `ParseUSDLights` call need to be updated? It's currently `/shapes`, which doesn't exist: https://github.com/ignitionrobotics/sdformat/blob/a8249fbc63ce23b01a73df2344f0bcc63a4c9fab/usd/src/usd_parser/USDLight_TEST.cc#L60-L61

usd/src/usd_parser/USDLight_TEST.cc
```diff
@@ -70,7 +70,7 @@ TEST(USDLightsTest, DistanceLight)
   EXPECT_EQ(ignition::math::Pose3d(0, 0, 10, 0, IGN_DTOR(45), 0),
       light->RawPose());
 
-  prim = stage->GetPrimAtPath(pxr::SdfPath("/shapes/diskLight"));
+  prim = stage->GetPrimAtPath(pxr::SdfPath("/diskLight"));
```
adlarkin, line 73: Since we no longer have the `shapes` top level prim in the testing file, does the `schemaToStop` in the corresponding `ParseUSDLights` call need to be updated? It's currently `/shapes`, which doesn't exist: https://github.com/ignitionrobotics/sdformat/blob/a8249fbc63ce23b01a73df2344f0bcc63a4c9fab/usd/src/usd_parser/USDLight_TEST.cc#L76-L77

usd/src/usd_parser/USDWorld.cc
```diff
@@ -94,6 +103,35 @@ namespace usd
...
+          prim,
+          usdData,
+          pose,
+          scale,
+          model.Name());
+        modelPtr->SetRawPose(pose);
+
+        modelPtr->SetStatic(!prim.HasAPI<pxr::UsdPhysicsRigidBodyAPI>());
+
+        _world.AddModel(model);
```
adlarkin, line 132: I don't think there's any need for `modelPtr` in this `if` block. We should be able to just use the `sdf::Model` object, so that the code would look something like this (note how the `sdf::Model` object is added to the world after all of its attributes are set):
```suggestion
        sdf::Model model = sdf::Model();
        model.SetName(primPathTokens[0]);

        ignition::math::Pose3d pose;
        ignition::math::Vector3d scale{1, 1, 1};
        GetTransform(
          prim,
          usdData,
          pose,
          scale,
          model.Name());
        model.SetRawPose(pose);

        model.SetStatic(!prim.HasAPI<pxr::UsdPhysicsRigidBodyAPI>());

        _world.AddModel(model);
```

usd/src/usd_parser/USDWorld.cc
```diff
@@ -59,6 +66,8 @@ namespace usd
     if (!errors.empty())
       return errors;
 
+    sdf::Model * modelPtr;
```
adlarkin, line 69: As I mentioned in my previous comment, I don't think that we need this.

adlarkin (approved):

PR #884, opened 2022-03-17
Port embedSdf script from Ruby to Python3 and provide unittests

Bi0T1N:
# 🎉 New feature
Removes the Ruby dependency and thus is related to/partially solves #274 (for complete Ruby dependency removal #643 should be merged as well).

## Summary
This PR converts the script that creates the `.cc` file for mapping of SDF filename and SDF content from Ruby to Python. Additionally a few unittests are added for the function that uses the mapping and which wasn't tested before.
For details see also the commit messages.

**Note**: Since I wasn't sure if this change counts as an API/ABI change or not, I targeted it to *main*.

## Checklist
- [x] Added tests


scpeters:

sdf/embedSdf.rb
```diff
@@ -1,45 +0,0 @@
...
-end
-
-# Embed the supported *.sdf files.
-supportedSdfVersions.each do |version|
-  Dir.glob("#{version}/*.sdf").sort.each { |file| embed(file) }
-end
-
-# Embed the supported *.convert files.
-supportedSdfConversions.each do |version|
-  Dir.glob("#{version}/*.convert").sort.each { |file| embed(file) }
```
scpeters, line 36: I've noticed some windows test failures. I've inspected the generated Embedded.cc file in our CI machine and believe it is related to `/` vs `\` characters for path separators. On this line (and line 31), in can be seen the ruby script uses `/` characters, but I see `\` in the generated file that has windows test failures. Here is an excerpt:

~~~
const std::map<std::string, std::string> &GetEmbeddedSdf()
{
  static const std::map<std::string, std::string> result {
// NOLINT
{
"1.9\actor.sdf",
R"__sdf_literal__(
<!-- Actor -->
~~~

scpeters: nice work! there's still one failure, but all the others are fixed now

src/SDF_TEST.cc
```diff
@@ -538,6 +538,47 @@ TEST(SDF, Version)
...
+
+TEST(SDF, EmbeddedSpecNonExistent)
+{
+  std::string result;
+  std::string output;
+
+  testing::internal::CaptureStderr();
+  result = sdf::SDF::EmbeddedSpec("unavailable.sdf", false);
+  output = testing::internal::GetCapturedStderr();
+  EXPECT_NE(output.find("Unable to find SDF filename"), std::string::npos);
```
scpeters, line 570: if you replace the `;` with ` << output;`, it should print the entire output if the expectation fails. This may help with debugging the failure on windows

Bi0T1N: Looks like there is an issue with the CI now: https://build.osrfoundation.org/job/sdformat-ci-pr_any-windows7-amd64/3327/

Bi0T1N: Your suggestion didn't helped much - nothing was printed, see https://build.osrfoundation.org/job/sdformat-ci-pr_any-windows7-amd64/3328/
However, I found a similar approach in parser_TEST.cc that doesn't depend on GTest and which hopefully will also work here. Otherwise I'll comment this test out for Windows and someone else has to fix it later with direct access to a Windows machine...

sdf/embedSdf.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    Generates a string pair with the folder and filename as well as the content
+    of the file in ascending order
+
+    :param arg_pathlist: Foldername and all filenames inside it
+    :returns: mapping pairs for the std::map
+    """
+    map_str = ''
+    files = []
+    for path in pathlist:
```
azeey, line 96: This is picking up `pathlist` from the `if __main__` block, not the argument to 
this function. I suggest extracting the contents of `if __main__` into its own 
function `main()` to avoid this.

```suggestion
    for path in arg_pathlist:
```

src/SDF_TEST.cc
```diff
@@ -538,6 +538,60 @@ TEST(SDF, Version)
...
+  EXPECT_NE(result.find("name=\"version\" type=\"string\""), std::string::npos);
+}
+
+TEST(SDF, EmbeddedSpecNonExistent)
+{
+  std::string result;
+
+  // Capture sdferr output
+  std::stringstream stderr_buffer;
+  auto old = std::cerr.rdbuf(stderr_buffer.rdbuf());
```
azeey, line 569: Could you use the `sdf::testing::RedirectConsoleStream` utility class for this? 
Here's an example: https://github.com/gazebosim/sdformat/blob/431acb2e1b542d4804a3505a61505e5a7aacb94f/test/integration/includes.cc#L681-L711

azeey (commented):
I left a few minor comments, but overall, this looks great! Thank you for your 
contribution!

`embedSdf.rb` has since been updated to include a new SDF version (1.10). Do 
you mind updating the python code to reflect that?

sdf/embedSdf.py
```diff
@@ -0,0 +1,149 @@
...
+    res.append(' * Unless required by applicable law or agreed to in writing, software')
+    res.append(' * distributed under the License is distributed on an "AS IS" BASIS,')
+    res.append(' * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.')
+    res.append(' * See the License for the specific language governing permissions and')
+    res.append(' * limitations under the License.')
+    res.append(' *')
+    res.append('*/')
+    res.append('')
+    return '\n'.join(res)
+
```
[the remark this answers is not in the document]
Bi0T1N: I had that in the beginning but it destroyed the formatting and I wasn't aware that there is a special function to remove the indentation.

src/SDF_TEST.cc
Bi0T1N: Done

Bi0T1N:
I fixed the merge conflict that was due to the `embedSdf.rb` already being deleted while changes for `SDF 1.10` have been added by targeting brach sdf13...

azeey (approved):
This looks great. Thanks @Bi0T1N for the contribution.

Made a small style change in https://github.com/gazebosim/sdformat/pull/884/commits/e16e985170b079899cb3f7f3342feef5a4091534.

j-rivero:
> I'm not sure if it will break anyone to make python required for build, but it may be worth checking with @j-rivero

Since this is changing ruby by python it does not look like a no-no-go to me but python can be a problem in some resource-constraint environments like embedded. Can we make it that python is an optional dependency and if it is not found it does not run the necessary tests but the rest is still working as before?

scpeters:
> Since this is changing ruby by python it does not look like a no-no-go to me but python can be a problem in some resource-constraint environments like embedded. Can we make it that python is an optional dependency and if it is not found it does not run the necessary tests but the rest is still working as before?

This pull request addresses a hard build requirement of generating the `EmbeddedSdf.cc` c++ code file, so I don't think it could be an optional dependency

j-rivero:
> This pull request addresses a hard build requirement of generating the `EmbeddedSdf.cc` c++ code file, so I don't think it could be an optional dependency

umm I see. As said, not the biggest no-go change but these kind of modifications implies downstream work and other possible hidden consequences. See: we have python already listed in [-release ](https://github.com/gazebo-release/sdformat13-release/blob/main/ubuntu/debian/control#L14) and [brew](https://github.com/osrf/homebrew-simulation/blob/master/Formula/sdformat13.rb#L24) so no big deal but @traversaro carefully don't have it in the [libsdformat13 build of conda](https://github.com/conda-forge/libsdformat-feedstock/blob/main/recipe/meta.yaml#L32-L38).

I leave this up to you, if the change is going to highly improve the maintenance or features in sdf13 go for it. If it is just a transition to get rid of ruby, I would target `main` for the change.

azeey:
I'm in favor of targeting `main` since this is not critical for `sdf13`.

scpeters:
> I'm in favor of targeting `main` since this is not critical for `sdf13`.

ok, let's retarget and update the cmake to require python

Bi0T1N:
Apparently I misinterpreted the comments and made the changes to the wrong project. Anyway, rebased my changes on main and added Python interpreter as a build dependency with `gz_build_error`.

mjcarroll (approved):

scpeters (approved):

PR #895, opened 2022-03-22
USD to SDF: material fixed

ahcorde:

# 🦟 Bug fix

## Summary

 - Set the workflow type
 - remove the texture file before copy it if this already exists
 - Avoid find resources in the local path

## Checklist
- [ ] Added tests

### DIFF
--- a/usd/src/usd_parser/USDData_TEST.cc
+++ b/usd/src/usd_parser/USDData_TEST.cc
@@ -44,10 +44,6 @@ TEST(USDData, Constructor)
   systemPaths->AddFilePaths(sdf::testing::TestFile("usd"));
 
   {
-    // Remove the material folder if there is already one created
-    ignition::common::removeAll(
-      ignition::common::joinPaths(ignition::common::cwd(), "materials"));
-
     sdf::testing::ScopeExit removeCopiedMaterials(
         []
         {
--- a/usd/src/usd_parser/USDData_TEST.cc
+++ b/usd/src/usd_parser/USDData_TEST.cc
@@ -43,8 +43,11 @@ TEST(USDData, Constructor)
   auto systemPaths = ignition::common::systemPaths();
   systemPaths->AddFilePaths(sdf::testing::TestFile("usd"));
 
-  // Open a valid USD file
   {
+    // Remove the material folder if there is already one created
+    ignition::common::removeAll(
+      ignition::common::joinPaths(ignition::common::cwd(), "materials"));
+
     sdf::testing::ScopeExit removeCopiedMaterials(
         []
         {
@@ -52,6 +55,7 @@ TEST(USDData, Constructor)
             ignition::common::joinPaths(ignition::common::cwd(), "materials"));
         });
 
+    // Open a valid USD file
     std::string filename = sdf::testing::TestFile("usd", "upAxisZ.usda");
     sdf::usd::USDData usdData(filename);
     EXPECT_EQ(0u, usdData.Init().size());
--- a/usd/src/usd_parser/USDMaterial.cc
+++ b/usd/src/usd_parser/USDMaterial.cc
@@ -39,17 +39,40 @@ namespace sdf
   /// \param[in] _dest The destination for the copy of _ori
   /// \return A list of UsdErrors. An empty list means no errors occurred when
   /// copying _ori to _dest
-  UsdErrors copyFile(const std::string &_ori, const std::string &_dest)
+  UsdErrors copyFile(const std::string &_ori, std::string &_dest)
   {
     UsdErrors errors;
     if (ignition::common::exists(_ori))
     {
-      // If the file exists then we should remove it and we will copy the
-      // file below
-      if (ignition::common::exists(_dest))
+      // If the file exists then we append a number suffix to the destination
+      // file
+      // For example:
+      // /bar/foo.extension
+      // /bar/foo_X.extension
+
+      std::string tmpDest = _dest;
+      std::string parentPath = ignition::common::parentPath(_dest);
+      std::string::size_type fileExtensionIndex = tmpDest.rfind(".");
+
+      std::string fileExtension = "." + tmpDest.substr(fileExtensionIndex + 1);
+      std::string fileNameWithoutExtension = ignition::common::basename(_dest);
+      size_t pos = fileNameWithoutExtension.find(fileExtension);
+      if (pos != std::string::npos)
       {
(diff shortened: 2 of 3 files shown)


ahcorde:

ahcorde:

usd/src/usd_parser/USDMaterial.cc
```diff
@@ -44,6 +44,13 @@ namespace sdf
     UsdErrors errors;
     if (ignition::common::exists(_ori))
     {
+      // If the file exists then we should remove it and we will copy the
+      // file below
+      if (ignition::common::exists(_dest))
+      {
+        ignition::common::removeFile(_dest);
+      }
```
adlarkin, line 52: Is there a chance that this code removes a file that is still needed at some point later on?

ahcorde: When you try to use `ignition::common::copyFile` and the file exists this will return false and it generates an error. I prefer to remove the existing file and void the error.

This may happen when you try to run several times the same command with the same file in the same folder.

ahcorde: done here https://github.com/ignitionrobotics/sdformat/pull/895/commits/d529bf4b3692569cd9b5db801e1ce13b65c6cc09
I don't know if there is an easy way to do it

usd/src/usd_parser/USDMaterial.cc
```diff
@@ -39,18 +39,41 @@ namespace sdf
...
+      if (pos != std::string::npos)
+      {
+        // If found then erase it from string
+        fileNameWithoutExtension.erase(pos, fileExtension.length());
+      }
+      int index = 0;
+      while (ignition::common::exists(tmpDest))
       {
-        ignition::common::removeFile(_dest);
+        tmpDest = ignition::common::joinPaths(
```
adlarkin, line 68: I actually don't think the `tmpDest` variable is needed, since the original `_dest` variable is never updated until after the while loop. You could make the loop look like this, and then remove the assignment of `tmpDest` to `_dest` after the while loop since `_dest` is now being updated in the loop as needed (this would allow removal of the `tmpDest` variable completely; we can use `_dest` wherever `tmpDest` is currently being used):
```c++
while (ignition::common::exists(_dest))
{
  _dest = ignition::common::jointPaths(...);
  index++;
}
```

usd/src/usd_parser/USDMaterial.cc
```diff
@@ -39,18 +39,41 @@ namespace sdf
...
+      // file
+      // For example:
+      // /bar/foo.extension
+      // /bar/foo_X.extension
+
+      std::string tmpDest = _dest;
+      std::string parentPath = ignition::common::parentPath(_dest);
+      std::string::size_type fileExtensionIndex = tmpDest.rfind(".");
+
+      std::string fileExtension = "." + tmpDest.substr(fileExtensionIndex + 1);
```
adlarkin, line 57: I think that we should check the value of `fileExtensionIndex`. If the destination string being searched for does not have a `.` character, `fileExtensionIndex` would be `std::npos`, which means that the `substr` call after `rfind` would throw an exception.

usd/src/usd_parser/USDMaterial.cc
```diff
@@ -39,18 +39,41 @@ namespace sdf
...
     if (ignition::common::exists(_ori))
     {
-      // If the file exists then we should remove it and we will copy the
-      // file below
-      if (ignition::common::exists(_dest))
+      // If the file exists then we append a number suffix to the destination
+      // file
+      // For example:
+      // /bar/foo.extension
+      // /bar/foo_X.extension
```
adlarkin, line 51: It would be nice to put all of the code for appending a numeric suffex to the destination file in an `if (ignition::common::exists(_dest))` block. That way, we don't do the preliminary work of saving the parent path, file extension, and file name without extension unless we absolutely have to.

usd/src/usd_parser/USDMaterial.cc
```diff
@@ -39,18 +39,41 @@ namespace sdf
...
+
+      std::string tmpDest = _dest;
+      std::string parentPath = ignition::common::parentPath(_dest);
+      std::string::size_type fileExtensionIndex = tmpDest.rfind(".");
+
+      std::string fileExtension = "." + tmpDest.substr(fileExtensionIndex + 1);
+      std::string fileNameWithoutExtension = ignition::common::basename(_dest);
```
adlarkin, line 58: nit: can `parentPath`, `fileExtension`, and `fileNameWithoutExtension` be marked `const`?

usd/src/usd_parser/USDMaterial.cc
adlarkin: I think the approach you took in https://github.com/ignitionrobotics/sdformat/commit/d529bf4b3692569cd9b5db801e1ce13b65c6cc09 is pretty good. I left a few review comments about it below.

usd/src/usd_parser/USDData_TEST.cc
```diff
@@ -43,15 +43,19 @@ TEST(USDData, Constructor)
   auto systemPaths = ignition::common::systemPaths();
   systemPaths->AddFilePaths(sdf::testing::TestFile("usd"));
 
-  // Open a valid USD file
   {
+    // Remove the material folder if there is already one created
+    ignition::common::removeAll(
+      ignition::common::joinPaths(ignition::common::cwd(), "materials"));
```
adlarkin, line 49: Is this necessary? I imagine that if the folder already exists, this test will simply create files in that folder and skip the folder creation step

adlarkin (commented):
LGTM, just one final question.

usd/src/usd_parser/USDData_TEST.cc
ahcorde: This was failing in the buildfarm, I decided to remove first the folder to avoid potencial issues

adlarkin: I believe I found the issue. It looks like a different test created a `materials` directory and didn't delete it when the test was finished, which caused this test to add numeric suffixes to files that already exist. I made a fix in https://github.com/ignitionrobotics/sdformat/pull/895/commits/c60ebc9ccaf761217aca14d4e1b1973f829de453, I believe CI should come back green now :crossed_fingers:

ahcorde: Ci is green! :green_circle:

osrf-triage:

PR #902, opened 2022-03-22
USD to SDF: Added links

ahcorde:
# 🎉 New feature

## Checklist
- [ ] Added tests


usd/src/usd_parser/USDLinks.cc
```diff
@@ -0,0 +1,163 @@
...
+    _link = new sdf::Link();
+    _link->SetName(_nameLink);
+
+    pxr::UsdPrim tmpPrim = _prim;
+    ignition::math::Pose3d pose;
+    ignition::math::Vector3d scale(1, 1, 1);
+    GetTransform(tmpPrim, _usdData, pose, scale, "");
+    size_t nSlash = std::count(_nameLink.begin(), _nameLink.end(), '/');
+    if (nSlash > 1)
+      _link->SetRawPose(pose);
```
adlarkin: Why is `nSlash` needed? In the `GetTransform` call, if the prim is the link prim itself, the pose should be `(0, 0, 0, 0, 0, 0)`, right? So, I imagine that we should always be able to set the link's raw pose without any issues.

usd/src/usd_parser/USDLinks.cc
```diff
@@ -0,0 +1,163 @@
...
+    // _link->SetScale(scale);
```
adlarkin, line 124: Can this line be deleted?

usd/src/usd_parser/USDLinks.cc
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  // If the schema is a rigid body use this name instead.
+  if (_prim.HasAPI<pxr::UsdPhysicsRigidBodyAPI>())
+  {
+    _link->SetName(primPathStr);
+  }
```
adlarkin: Why is this needed? Also, should the link's name be set to the basename of the prim path here?

usd/src/usd_parser/USDLinks.cc
```diff
@@ -0,0 +1,148 @@
...
+    if (nSlash > 1)
+      _link->SetRawPose(pose);
+    // _link->SetScale(scale);
+  }
+
+  // If the schema is a rigid body use this name instead.
+  if (_prim.HasAPI<pxr::UsdPhysicsRigidBodyAPI>())
+  {
+    _link->SetName(pxr::TfStringify(_prim.GetPath()));
+  }
```
adlarkin, line 131: Why is this needed? Also, should we be setting the link's name to the basename of the prim path string?

adlarkin (commented):
I pushed a few commits with feedback, but I still have a few comments/questions which I left below. There are also a few other things:
1. Would you mind adding tests that show adding links to models works as expected?
2. I see the following warning when I build the code:
```
/sdf_to_usd/src/sdformat/usd/src/usd_parser/USDWorld.cc: In function 'sdf::v12::usd::UsdErrors sdf::v12::usd::parseUSDWorld(const string&, sdf::v12::World&)':
/sdf_to_usd/src/sdformat/usd/src/usd_parser/USDWorld.cc:227:30: warning: 'modelPtr' may be used uninitialized in this function [-Wmaybe-uninitialized]
             modelPtr->AddLink(*link);
```

usd/src/usd_parser/USDLinks.cc
```diff
@@ -0,0 +1,163 @@
...
+        diagonalInertia = pxr::GfVec3f(0.0001, 0.0001, 0.0001);
+      }
+
+      if (mass < 0.0001)
+      {
+        mass = 0.1;
+      }
+
+      massMatrix.SetMass(mass);
+      massMatrix.SetDiagonalMoments(
```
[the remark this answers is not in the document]
ahcorde: I'm not really sure how to set this values with the current API, any thoughts @adlarkin @scpeters ?

usd/src/usd_parser/USDLinks.cc
```diff
@@ -0,0 +1,163 @@
...
+  std::string primPathStr = pxr::TfStringify(_prim.GetPath());
+  std::string primType = _prim.GetPrimTypeInfo().GetTypeName().GetText();
+
+  std::pair<std::string, std::shared_ptr<USDStage>> data =
+    _usdData.FindStage(primNameStr);
+
+  // Is this a new link ?
+  if (_link == nullptr)
+  {
+    _link = new sdf::Link();
```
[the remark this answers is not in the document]
ahcorde: we are getting a raw pointer from `sdf::Link` I was trying to keep it simple

usd/src/usd_parser/USDLinks.cc
ahcorde: if a path is a rigid body I used this name as the name of the link

adlarkin:
@ahcorde it looks like there are merge conflicts with `sdf12`, would you mind resolving them?

ahcorde:
Done @adlarkin

usd/src/usd_parser/USDLinks.cc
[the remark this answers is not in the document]
adlarkin: I left a `TODO` note in the code about adding these things in so that we can revisit it later

usd/src/usd_parser/USDLinks.cc
```diff
@@ -0,0 +1,162 @@
...
+  const std::string &_nameLink,
+  sdf::Link *_link,
+  USDData &_usdData)
+{
+  std::string primNameStr = _prim.GetPath().GetName();
+  std::string primPathStr = pxr::TfStringify(_prim.GetPath());
+  std::string primType = _prim.GetPrimTypeInfo().GetTypeName().GetText();
+
+  std::pair<std::string, std::shared_ptr<USDStage>> data =
+    _usdData.FindStage(primNameStr);
```
adlarkin, line 131: I don't see the `data` variable being used anywhere else in this method. Can this call to `FindStage` be removed? I'm not sure that it's needed.

usd/src/usd_parser/USDLinks.cc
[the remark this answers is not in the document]
adlarkin: I changed the API slightly to use `std::optional<sdf::Link>` instead of raw pointers. I think it's a safer approach since it avoids the need for manually allocating/de-allocating memory

usd/src/usd_parser/USDLinks.cc
```diff
@@ -0,0 +1,162 @@
...
+  {
+    _link = new sdf::Link();
+    _link->SetName(ignition::common::basename(_nameLink));
+
+    ignition::math::Pose3d pose;
+    ignition::math::Vector3d scale(1, 1, 1);
+    GetTransform(_prim, _usdData, pose, scale, "");
+    size_t nSlash = std::count(_nameLink.begin(), _nameLink.end(), '/');
+    if (nSlash > 1)
+      _link->SetRawPose(pose);
```
adlarkin, line 144: @ahcorde I left a [previous comment](https://github.com/ignitionrobotics/sdformat/pull/902#discussion_r840922946) about this, but I still don't understand the purpose of this code. Can you explain why we need to count the slashes and make sure it's `> 1` before setting the pose? I think it would be nice to leave a comment in the code in case this has to be re-visited later.

adlarkin (commented):
I pushed a few changes, with the main one being using `std::optional<sdf::Link>` instead of a raw pointer to `sdf::Link`. I have a few final questions below.

usd/src/usd_parser/USDLinks.cc
ahcorde: Added a explanation and added TODO

usd/src/usd_parser/USDLinks.cc
ahcorde: removed

ahcorde:
@adlarkin ready for a final review

usd/src/usd_parser/USDLinks.cc
```diff
@@ -139,6 +136,10 @@ void ParseUSDLinks(
     ignition::math::Pose3d pose;
     ignition::math::Vector3d scale(1, 1, 1);
     GetTransform(_prim, _usdData, pose, scale, "");
+    // This is a special case when a geometry is defined in the higher level
+    // of the path. we should only set the position is the path at least has
```
adlarkin, line 140:
```suggestion
    // of the path. we should only set the position if the path at least has
```

adlarkin (approved):

PR #904, opened 2022-03-22
USD to SDF: Added diff drive plugin

ahcorde:
# 🎉 New feature

## Summary

USD to SDF: Added diff drive plugin

## Checklist
- [ ] Added tests


usd/src/usd_parser/USDWorld.cc
```diff
@@ -163,6 +163,56 @@ namespace usd
...
+          pxr::TfToken("leftWheelJointName"));
+        auto rightWheelAttr = prim.GetAttribute(
+          pxr::TfToken("rightWheelJointName"));
+        auto wheelBaseAttr = prim.GetAttribute(
+          pxr::TfToken("wheelBase"));
+        auto wheelRadiusAttr = prim.GetAttribute(
+          pxr::TfToken("wheelRadius"));
+
+        sdf::Plugin diffDrivePlugin;
+        diffDrivePlugin.SetName("ignition::gazebo::systems::DiffDrive");
```
azeey, line 178: I know this is a Draft, but I just noticed this. In general, we're trying to make SDFormat less Gazebo-centric. Adding a specific plugin like this seems to go against that. Could this be handled somewhere outside of the SDFormat library?

adlarkin: I think you make a good point here, but unfortunately, I'm not sure of another workaround/solution at the moment. We are also already doing this elsewhere in the usd -> sdf parser, which probably needs to be updated :grimacing:: https://github.com/ignitionrobotics/sdformat/blob/1637dda60744e041fa49ebec2c7d0dfaa135521a/usd/src/usd_parser/USDWorld.cc#L322-L343

What if for now, we updated the [`usd2sdf` command](https://github.com/ignitionrobotics/sdformat/blob/sdf12/usd/src/cmd/usd2sdf.cc) to have a command line argument that specifies whether users want `ign-gazebo` plugins or not? I think that would resolve the permanent dependency on these plugins for now. If the long-term goal is to move the gazebo plugin code out of the SDFormat codebase completely, I can create an issue about this so that it can be re-visited later.

scpeters: I agree with Addisu's point, and your proposed fix (command-line flag and issue)

adlarkin: If I were to take the command-line flag approach, it looks like the API of this public method would have to be broken (I'm thinking I'd pass along a boolean that indicates whether gazebo plugins should be included or not): https://github.com/ignitionrobotics/sdformat/blob/d939ca45eb8fc955a1d8ff037fcaad87809dc23a/usd/include/sdf/usd/usd_parser/Parser.hh#L40-L42

Here's where this method is called from the `usd2sdf` executable: https://github.com/ignitionrobotics/sdformat/blob/b2364c1caca15e7f9a1d75bb8dd698bb1cf33b73/usd/src/cmd/usd2sdf.cc#L51

Would this be okay since we are still only building the USD component from source and haven't released binaries for it yet?

adlarkin: Another alternative to modifying the API would be to set an environment variable, but I assume this approach has its own set of drawbacks as well.

azeey: > Would this be okay since we are still only building the USD component from source and haven't released binaries for it yet?

If we haven't released binaries, breaking ABI should be okay, IMO, but not API.

adlarkin: I implemented allowing/disallowing gazebo plugins from the command line in https://github.com/ignitionrobotics/sdformat/pull/904/commits/4d4292c512ee446890a4ff27cac87e466f11072a. Let me know if this looks good to you. If so, I will go ahead and create an issue that links to this conversation so that we can re-visit it later.

adlarkin (approved):

usd/src/usd_parser/USDWorld.cc
azeey: Thanks. Looks good!

adlarkin: I created an issue that references this conversation: https://github.com/ignitionrobotics/sdformat/issues/987

I'll mark this as resolved since we can re-visit it later via the issue.

PR #911, opened 2022-03-25
Added convenience constructor to plugin

nkoenig:

# 🎉 New feature

A non-trivial amount of code is required to programmatically construct a plugin that has content. For example, this plugin
```
<plugin  filename="ignition-gazebo-sensors-system"  name="ignition::gazebo::systems::Sensors">
  <render_engine>ogre2</render_engine>
</plugin>
```

Requires this code:

```
sdf::Plugin sensorPlugin;
sensorPlugin.SetFilename("ignition-gazebo-sensors-system");
sensorPlugin.SetName("ignition::gazebo::systems::Sensors");
sdf::ElementPtr sensorElem(new sdf::Element);
sensorElem->SetName("render_engine");
sensorElem->AddValue("string", "ogre2", true);
sensorElem->Set<std::string>("ogre2");
sensorPlugin.InsertContent(sensorElem);
```

This pull request adds an `InsertContent` function and an initialization constructor to the `Plugin` class. Now you'll be able to build the same plugin using:
```
sdf::Plugin sensorPlugin("ignition-gazebo-sensors-system",  "ignition::gazebo::systems::Sensors", "render_engine>ogre2</render_engine>");
```

I copied the `copyChildren` function from `parser.cc` to `Utils.cc` so that the Plugin class can use it. This shouldn't effect API/ABI.

## Test it

New tests have been added to `Plugin_TEST.cc`

## Checklist
- [x] Added tests


src/CMakeLists.txt
```diff
@@ -118,6 +120,7 @@ target_link_libraries(${PROJECT_LIBRARY_TARGET_NAME}
     ignition-math${IGN_MATH_VER}::ignition-math${IGN_MATH_VER}
     ignition-utils${IGN_UTILS_VER}::ignition-utils${IGN_UTILS_VER}
   PRIVATE
+    ignition-common${IGN_COMMON_VER}::ignition-common${IGN_COMMON_VER}
```
ahcorde, line 123: we added `ignition::common` for USD I'm not really sure that you can use this dependency here

src/Plugin.cc
```diff
@@ -42,6 +43,18 @@ Plugin::Plugin()
...
 }
 
+/////////////////////////////////////////////////
+Plugin::Plugin(const std::string &_filename, const std::string &_name,
+               const std::string &_xmlContent)
+  : dataPtr(std::make_unique<sdf::PluginPrivate>())
+{
+  this->SetFilename(_filename);
+  this->SetName(_name);
+  std::string trimmed = ignition::common::trimmed(_xmlContent);
```
ahcorde, line 53: you are only using ignition::common here, can you remove this dependency here ?

ahcorde (changes requested):

src/CMakeLists.txt
nkoenig: The dependency doesn't seem bad to me. Here is the code I'd duplicate, which I could trim (haha) down:

```
void ignition::common::ltrim(std::string &_s)
{
  _s.erase(_s.begin(), std::find_if(_s.begin(), _s.end(),
        [](int c) {return !std::isspace(c);}));
}

void ignition::common::rtrim(std::string &_s)
{
  _s.erase(std::find_if(_s.rbegin(), _s.rend(),
        [](int c) {return !std::isspace(c);}).base(),
      _s.end());
}

void ignition::common::trim(std::string &_s)
{
  ignition::common::ltrim(_s);
  ignition::common::rtrim(_s);
}

std::string ignition::common::ltrimmed(std::string _s)
{
  ignition::common::ltrim(_s);
  return _s;
}   

std::string ignition::common::rtrimmed(std::string _s)
{
  ignition::common::rtrim(_s);
  return _s;
}

std::string ignition::common::trimmed(std::string _s)
{
  ignition::common::trim(_s);
  return _s;
}
```

Utilizing existing code rather than duplicating it seems like something we should do internally. Otherwise, we should be telling users to copy our code, and not use the libraries.

azeey: I don't think we should add a dependency in a stable version. We already have `sdf::trim` that can be used instead.

nkoenig: Good point, this is a stable version.

src/Plugin.cc
nkoenig: Done in [e6fc820](https://github.com/ignitionrobotics/sdformat/pull/911/commits/e6fc820d074fc8f07a654210142e147a69c909d9)

src/CMakeLists.txt
nkoenig: Done in [e6fc820](https://github.com/ignitionrobotics/sdformat/pull/911/commits/e6fc820d074fc8f07a654210142e147a69c909d9)

chapulina: Adding `ign-common` as a dependency to the core of SDF is something we should think carefully about. That brings a lot of dependencies. An alternative to consider would be to move some of this functionality to `ign-utils`.

PR #914, opened 2022-03-28
SDF to USD: set UsdPhysicsArticulationRootAPI

ahcorde:

# 🦟 Bug fix

We should set the `UsdPhysicsArticulationRootAPI` to allow articulation arms to move in IsaacSim

## Checklist
- [ ] Added tests

### DIFF
--- a/usd/src/sdf_parser/Model.cc
+++ b/usd/src/sdf_parser/Model.cc
@@ -181,8 +181,7 @@ namespace usd
       {
         errors.push_back(UsdError(sdf::usd::UsdErrorCode::FAILED_PRIM_API_APPLY,
               "Internal error: unable to mark model at path [" +
-              modelPrim.GetPath().GetString() + "] as a rigid body, "
-              "so mass properties won't be attached"));
+              modelPrim.GetPath().GetString() + "] as a rigid body."));
         return errors;
       }
     }
--- a/usd/src/sdf_parser/Model.cc
+++ b/usd/src/sdf_parser/Model.cc
@@ -32,6 +32,7 @@
 #include <pxr/usd/sdf/path.h>
 #include <pxr/usd/usd/stage.h>
 #include <pxr/usd/usdGeom/xform.h>
+#include <pxr/usd/usdPhysics/articulationRootAPI.h>
 #pragma pop_macro ("__DEPRECATED")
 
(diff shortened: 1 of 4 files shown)


adlarkin (approved):

ahcorde:
@adlarkin I made more changes, it was not working in same cases.

usd/src/sdf_parser/Model.cc
```diff
@@ -145,9 +157,35 @@ namespace usd
...
+        {
+          markedArticulationRoot = true;
+        }
+      }
     }
 
-    // TODO(adlarkin) finish parsing model
+    if (!markedArticulationRoot && !_model.Static())
+    {
+      if (!pxr::UsdPhysicsRigidBodyAPI::Apply(modelPrim))
```
adlarkin, line 180: I have a few questions about this:
1. If we are marking the model as a rigid body API, then why do we need to also mark the link as a rigid body API? I believe that marking only the model as a rigid body API should be sufficient, since the link is a child prim of the model prim.
2. In the first `if` block, why do we check if the model is not static **and** if the model has not been marked as an articulation root? As I understand it, `pxr::UsdPhysicsRigidBodyAPI` only implies static/non-static, which means that articulation root API should not have any influence on whether a prim is marked as a rigid body API or not. The only thing that should influence whether a prim is a rigid body API is whether it is static or not, so I believe the first `if` statement here should only check if the model is not static instead of checking of the model isn't static and also isn't marked as an articulation root API.

ahcorde: When I was converting the panda robot to USD I was having some issues. Here it's what I found testing the model in Isaac Sim

 - A `UsdPhysicsRigidBodyAPI` can't be marked as `rigidbody` otherwise I was having issue with the joints
 - If the parent is marked as `UsdPhysicsRigidBodyAPI` then I need to set all children to `rigidbody`.

I don't have a clear explanation about why is this happening but I fix this based on the behaviour in Issac Sim

adlarkin: Okay, thanks for the feedback :+1: so if a model prim is marked as an articulation root API, but not as a rigid body API, does it still move without issues? From the initial reading I did on USD, if a prim isn't marked as a rigid body, then I thought this prim would be interpreted as static.

PR #923, opened 2022-03-30
Use helper functions from ignition common and math

nkoenig:

# 🎉 New feature

This deprecates duplicate functions found in libsdformat in favor of the functions found in ignition common.

## Test it

The tests should all pass.

## Checklist
- [ ] Added tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,7 +111,8 @@ if (BUILD_SDF)
 
   ########################################
   # Find ignition common
-  ign_find_package(ignition-common5 COMPONENTS graphics REQUIRED)
+  ign_find_package(ignition-common5 COMPONENTS graphics REQUIRED_BY usd)
+  ign_find_package(ignition-common5 REQUIRED)
   set(IGN_COMMON_VER ${ignition-common5_VERSION_MAJOR})
 
   ########################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,7 +111,7 @@ if (BUILD_SDF)
 
   ########################################
   # Find ignition common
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ if (BUILD_SDF)
 
   ########################################
   # Find ignition common
-  ign_find_package(ignition-common5 COMPONENTS graphics REQUIRED_BY usd)
+  ign_find_package(ignition-common5 COMPONENTS graphics REQUIRED)
```
azeey, line 114: Is there a way to require only the core of ign-common and not the graphics component? I think the graphics component brings too many dependencies that we wouldn't want to impose on SDFormat users.

Even the core of ign-common brings a dependency on `libdl`, which I think is an unnecessary constraint for SDFormat. As @chapulina suggested in https://github.com/ignitionrobotics/sdformat/pull/911#discussion_r838737967, perhaps we should move some of the functions used in this PR to `ign-utils`.

chapulina (changes requested):
It's not worth it bringing `ign-common` as a dependency just to use a few short helper functions. There's some discussion in https://github.com/ignitionrobotics/sdformat/issues/351 about why bringing that dependency is problematic.

nkoenig:
> It's not worth it bringing `ign-common` as a dependency just to use a few short helper functions. There's some discussion in #351 about why bringing that dependency is problematic.

I've opened two other pull requests that replaces Console and Filesystem with the equivalents in ign-common.

CMakeLists.txt
nkoenig: I've added a find package for the just the main part of common. There are a couple other pull requests that also make use of ign-common's Filesystem and Console.

PR #939, opened 2022-04-06
Adding sdf::Error logging to Param

marcoag:
# 🎉 New feature

Work towards https://github.com/ignitionrobotics/sdformat/issues/820.

## Summary
Adding `sdf::Errors &_errors` as an argument option to any function that does any logging. The user can now choose to call functions with the `sdf::Errors` argument to retrieve the error messages or call them without it and get the same old behavior.

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/Error.hh
+++ b/include/sdf/Error.hh
@@ -147,6 +147,14 @@ namespace sdf
     /// \brief Merge include is unspported for the type of entity being
     /// included, or the custom parser does not support merge includes.
     MERGE_INCLUDE_UNSUPPORTED,
+
+    /// \brief Generic error type for parameters (values of SDFormat elements
+    /// or attributes).
+    PARAMETER_ERROR,
+
+    /// \brief The specified parameter (values of SDFormat elements
+    /// or attributes) type is unknown.
+    UNKNOWN_PARAMETER_TYPE,
   };
 
   class SDFORMAT_VISIBLE Error
--- a/include/sdf/Param.hh
+++ b/include/sdf/Param.hh
@@ -136,6 +136,19 @@ namespace sdf
                   const std::string &_default, bool _required,
                   const std::string &_description = "");
 
+    /// \brief Constructor.
+    /// \param[in] _key Key for the parameter.
+    /// \param[in] _typeName String name for the value type (double,
+    /// int,...).
+    /// \param[in] _default Default value.
+    /// \param[in] _required True if the parameter is required to be set.
+    /// \param[out] _errors Vector of errors.
+    /// \param[in] _description Description of the parameter.
+    public: Param(const std::string &_key, const std::string &_typeName,
+                  const std::string &_default, bool _required,
+                  sdf::Errors &_errors,
+                  const std::string &_description = "");
+
     /// \brief Constructor with min and max values.
     /// \param[in] _key Key for the parameter.
     /// \param[in] _typeName String name for the value type (double,
@@ -149,6 +162,22 @@ namespace sdf
     public: Param(const std::string &_key, const std::string &_typeName,
                   const std::string &_default, bool _required,
                   const std::string &_minValue, const std::string &_maxValue,
+                  const std::string &_description = "");
+
+    /// \brief Constructor with min and max values.
+    /// \param[in] _key Key for the parameter.
+    /// \param[in] _typeName String name for the value type (double,
+    /// int,...).
+    /// \param[in] _default Default value.
+    /// \param[in] _required True if the parameter is required to be set.
+    /// \param[in] _minValue Minimum allowed value for the parameter.
+    /// \param[in] _maxValue Maximum allowed value for the parameter.
+    /// \param[out] _errors Vector of errors.
+    /// \param[in] _description Description of the parameter.
+    public: Param(const std::string &_key, const std::string &_typeName,
+                  const std::string &_default, bool _required,
+                  const std::string &_minValue, const std::string &_maxValue,
+                  sdf::Errors &_errors,
                   const std::string &_description = "");
 
     /// \brief Copy constructor
@@ -180,13 +209,29 @@ namespace sdf
     public: std::string GetAsString(
         const PrintConfig &_config = PrintConfig()) const;
 
+    /// \brief Get the value as a string.
+    /// \param[out] _errors Vector of errors.
+    /// \param[in] _config Configuration for conversion to string.
+    /// \return String containing the value of the parameter.
+    public: std::string GetAsString(
+        sdf::Errors &_errors,
+        const PrintConfig &_config = PrintConfig()) const;
+
     /// \brief Get the default value as a string.
     /// \param[in] _config Configuration for conversion to string.
     /// \return String containing the default value of the parameter.
     public: std::string GetDefaultAsString(
         const PrintConfig &_config = PrintConfig()) const;
 
-    /// \brief Get the minimum allowed value as a string
+    /// \brief Get the default value as a string.
+    /// \param[out] _errors Vector of errors.
+    /// \param[in] _config Configuration for conversion to string.
+    /// \return String containing the default value of the parameter.
+    public: std::string GetDefaultAsString(
+        sdf::Errors &_errors,
+        const PrintConfig &_config = PrintConfig()) const;
+
+    /// \brief Get the minimum allowed value as a string.
     /// \param[in] _config Configuration for conversion to string.
     /// \return Returns a string containing the minimum allowed value of the
     /// parameter if the minimum value is specified in the SDFormat description
@@ -194,12 +239,32 @@ namespace sdf
     public: std::optional<std::string> GetMinValueAsString(
         const PrintConfig &_config = PrintConfig()) const;
 
-    /// \brief Get the maximum allowed value as a string
+    /// \brief Get the minimum allowed value as a string.
+    /// \param[out] _errors Vector of errors.
+    /// \param[in] _config Configuration for conversion to string.
+    /// \return Returns a string containing the minimum allowed value of the
+    /// parameter if the minimum value is specified in the SDFormat description
+    /// of the parameter. nullopt otherwise.
+    public: std::optional<std::string> GetMinValueAsString(
+        sdf::Errors &_errors,
+        const PrintConfig &_config = PrintConfig()) const;
+
+    /// \brief Get the maximum allowed value as a string.
     /// \param[in] _config Configuration for conversion to string.
     /// \return Returns a string containing the maximum allowed value of the
     /// parameter if the maximum value is specified in the SDFormat description
     /// of the parameter. nullopt otherwise.
     public: std::optional<std::string> GetMaxValueAsString(
+        const PrintConfig &_config = PrintConfig()) const;
+
+    /// \brief Get the maximum allowed value as a string.
+    /// \param[in] _config Configuration for conversion to string.
+    /// \param[out] _errors Vector of errors.
+    /// \return Returns a string containing the maximum allowed value of the
+    /// parameter if the maximum value is specified in the SDFormat description
+    /// of the parameter. nullopt otherwise.
+    public: std::optional<std::string> GetMaxValueAsString(
+        sdf::Errors &_errors,
         const PrintConfig &_config = PrintConfig()) const;
 
     /// \brief Set the parameter value from a string.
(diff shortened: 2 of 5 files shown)


include/sdf/Error.hh
```diff
@@ -147,6 +147,9 @@ namespace sdf
     /// \brief Merge include is unspported for the type of entity being
     /// included, or the custom parser does not support merge includes.
     MERGE_INCLUDE_UNSUPPORTED,
+
+    /// \brief Generic error type for parameters.
```
azeey, line 151: Since the word "parameter" is so overloaded, can you add to the comment that this pertains to the values of SDFormat elements or attributes?

include/sdf/Param.hh
```diff
@@ -401,6 +577,18 @@ namespace sdf
                                     const std::string &_valueStr,
                                     ParamVariant &_valueToSet) const;
 
+    /// \brief Method used to set the Param from a passed-in string
+    /// \param[in] _typeName The data type of the value to set
+    /// \param[in] _valueStr The value as a string
+    /// \param[out] _valueToSet The value to set
+    /// \param[out] _errors Vector of errors.
+    /// \return True if the value was successfully set, false otherwise
+    public: bool SDFORMAT_VISIBLE ValueFromStringImpl(
```
azeey, line 586: This is an internal function (even though the symbol is visible), so we can just change the original instead of creating an overload.

include/sdf/Param.hh
```diff
@@ -414,6 +602,21 @@ namespace sdf
...
 
+    /// \brief Method used to get the string representation from a ParamVariant,
+    /// or the string that was used to set it.
+    /// \param[in] _config Print configuration for the string output
+    /// \param[in] _typeName The data type of the value
+    /// \param[in] _value The value
+    /// \param[out] _valueStr The output string.
+    /// \param[out] _errors Vector of errors.
+    /// \return True if the string was successfully retrieved, false otherwise.
+    public: bool StringFromValueImpl(
```
azeey, line 613: This is an internal function, so we can just change the original instead of creating an overload.

include/sdf/Param.hh
```diff
@@ -430,6 +633,24 @@ namespace sdf
...
+    /// or the string that was used to set it.
+    /// \param[in] _config Print configuration for the string output
+    /// \param[in] _typeName The data type of the value
+    /// \param[in] _value The value
+    /// \param[in] _orignalStr The original string that was used to set the
+    /// value. A nullopt can be passed in if it is not available.
+    /// \param[out] _valueStr The output string.
+    /// \param[out] _errors Vector of errors.
+    /// \return True if the string was successfully retrieved, false otherwise.
+    public: bool StringFromValueImpl(
```
azeey, line 646: This is an internal function, so we can just change the original instead of creating an overload.

include/sdf/Param.hh
```diff
@@ -487,26 +708,53 @@ namespace sdf
   ///////////////////////////////////////////////
   template<typename T>
   bool Param::Set(const T &_value)
+  {
+    sdf::Errors errors;
+    bool result = this->Set<T>(_value, errors);
+    for (const auto &e : errors)
+    {
+      sdferr << e.Message() + "\n";
+    }
```
azeey, line 717: This for-loop can be shortened to `sdferr << errors << "\n";`

include/sdf/Param.hh
```diff
@@ -487,26 +708,53 @@ namespace sdf
...
   ///////////////////////////////////////////////
   template<typename T>
   bool Param::Get(T &_value) const
+  {
+    sdf::Errors errors;
+    bool result = this->Get<T>(_value, errors);
+    for (const auto &e : errors)
+    {
+      sdferr << e.Message() + "\n";
+    }
```
azeey, line 751: This for-loop can be shortened to `sdferr << errors << "\n";`

include/sdf/Param.hh
```diff
@@ -518,13 +766,15 @@ namespace sdf
       std::string typeStr = this->dataPtr->TypeToString<T>();
       if (typeStr.empty())
       {
-        sdferr << "Unknown parameter type[" << typeid(T).name() << "]\n";
+        _errors.push_back({ErrorCode::PARAMETER_ERROR,
```
azeey, line 769: Would a new error code, `UNKNOWN_PARAMETER_TYPE` make more sense?

include/sdf/Param.hh
```diff
@@ -556,6 +806,19 @@ namespace sdf
   ///////////////////////////////////////////////
   template<typename T>
   bool Param::GetDefault(T &_value) const
+  {
+    sdf::Errors errors;
+    bool result = this>GetDefault<T>(_value, errors);
+    for (const auto &e : errors)
+    {
+      sdferr << e.Message() + "\n";
+    }
```
azeey, line 815: This for-loop can be shortened to `sdferr << errors << "\n";`

include/sdf/Error.hh
```diff
@@ -147,6 +147,9 @@ namespace sdf
...
+    PARAMETER_ERROR
```
scpeters, line 152: nit: use a trailing comma

include/sdf/Param.hh
```diff
@@ -171,28 +200,63 @@ namespace sdf
...
+    /// \brief Get the minimum allowed value as a string.
+    /// \param[out] _errors Vector of errors.
+    /// \param[in] _config Configuration for conversion to string.
+    /// \return Returns a string containing the minimum allowed value of the
+    /// parameter if the minimum value is specified in the SDFormat description
+    /// of the parameter. nullopt otherwise.
+    public: std::optional<std::string> GetMinValueAsString(
+        sdf::Errors &_errors,
+        const PrintConfig &_config = PrintConfig()) const;
+    /// \brief Get the maximum allowed value as a string.
```
scpeters, line 242: nit: add blank line above this

src/Param.cc
```diff
@@ -65,54 +65,66 @@ Param::Param(const std::string &_key, const std::string &_typeName,
...
-          this->dataPtr->defaultValue),
-      "Invalid parameter");
-  this->dataPtr->value = this->dataPtr->defaultValue;
-  this->dataPtr->strValue = std::nullopt;
+  if(!errors.empty())
+  {
+    for (unsigned int i = 0; i < errors.size() - 1; ++i)
+    {
+      sdferr << errors[i].Message() + "\n";
+    }
```
azeey, line 77: This for-loop can be shortened to `sdferr << errors << "\n";`

src/Param.cc
```diff
@@ -65,54 +65,66 @@ Param::Param(const std::string &_key, const std::string &_typeName,
...
+
+    SDF_ASSERT(false, errors[errors.size() - 1].Message());
```
azeey, line 79: nit: You can use `errors.back()` instead.

src/Param.cc
```diff
@@ -1209,21 +1576,29 @@ bool Param::ValidateValue() const
...
-                     << *this->GetMinValueAsString() << "] for key ["
-                     << this->GetKey() << "]\n";
+              std::ostringstream oss;
+              oss << _val;
+              std::string val_str = oss.str();
+              _errors.push_back({ErrorCode::PARAMETER_ERROR,
+                  "The value [" + val_str
+                  + "] is less than the minimum allowed value of ["
+                  + *this->GetMinValueAsString() + "] for key ["
+                  + this->GetKey() + "]"});
```
azeey, line 1586: nit: Since we're creating an `ostringstream`, we could just use it to construct the entire string and then call `err-rs.push_back({ErrorCode::PARAMETER_ERROR, oss.str()})`

azeey (commented):
I have some minor comments below, but looks great in general! I think it would be good to add tests to ensure that console messages are not accidentally being printed. You can use `sdf::testing::RedirectConsoleStream` to capture the console output.

include/sdf/Error.hh
marcoag: Good idea, done.

include/sdf/Param.hh
marcoag: Oh ok, I was worried since it's `public` I could break something. I will remove the version without the `sdf::Errors` argument then.

include/sdf/Param.hh
marcoag: True, done.

include/sdf/Param.hh
marcoag: Removing the overload and changing the original then.

include/sdf/Param.hh
marcoag: Removing the overload and changing the original then.

src/Param.cc
marcoag: True, will do for both cases.

src/Param.cc
marcoag: Also changed here: https://github.com/ignitionrobotics/sdformat/blob/a7ee2929e5e51bb578ef9a0f85ad18df76d9fc1d/src/Param.cc#L112.

include/sdf/Param.hh
marcoag: Definitely, I  added the new error type and I also used it here: https://github.com/ignitionrobotics/sdformat/blob/a7ee2929e5e51bb578ef9a0f85ad18df76d9fc1d/src/Param.cc#L995

include/sdf/Param.hh
azeey: I couldn't see the changes. Has it been pushed?

src/Param.cc
```diff
@@ -65,54 +65,66 @@ Param::Param(const std::string &_key, const std::string &_typeName,
...
-  this->dataPtr->value = this->dataPtr->defaultValue;
-  this->dataPtr->strValue = std::nullopt;
+  if(!errors.empty())
+  {
+    for (unsigned int i = 0; i < errors.size() - 1; ++i)
+    {
+      sdferr << errors[i].Message() + "\n";
+    }
+
+    SDF_ASSERT(false, errors.back().Message());
```
scpeters, line 79: what is the purpose of this assert?

marcoag: The idea was to keep the same behavior as before (see https://github.com/ignitionrobotics/sdformat/blob/e363110e9c5a3671acb371537a6eaef5617f1d28/src/Param.cc#L106 ). Basically, f there are any errors in the vector use the assert to throw the exception with the last added message as it was done previously.

scpeters: thanks, I hadn't followed the changes closely enough, but I get it now!

src/Param.cc
azeey: I think you skipped this one.

azeey (commented):
I believe the only thing left is to add some tests to ensure error messages are not sent to the console.

src/Param.cc
marcoag: I wanted to keep previous behavior and print all messages except the last one. Then use the last one in the `SDF_ASSERT`. I guess I could create a "subvector" and pass that to `sdferr` and then call `SDF_ASSERT` with the last message, do think that's worth it?

azeey: Ah, okay. I'm good with that. The nice thing about doing `sdferr << errors` is that it prints other parts of the error, such as the error code and the line number (if there is one associated with the error). I won't block on this.

test/integration/error_output.cc
```diff
@@ -0,0 +1,150 @@
...
+            "0.59999999999999987' to key [] for new parent element of name '',"
+            " reverting to previous value '1 2 3 0.40000000000000002 0.5 "
+            "0.59999999999999987'."
+            , error.Message());
+
+  sdf::Param param4("key", "bool", "15", false, "a", "b", errors,
+                    "description");
+  error = errors.front();
+  errors.erase(errors.begin());
+  ASSERT_EQ(sdf::ErrorCode::PARAMETER_ERROR, error.Code());
```
azeey, line 122: I think it would be clearer to index into the `errors` object than to pop each error and assign to `error`. Here's an example: https://github.com/ignitionrobotics/sdformat/blob/4fa1be3140b827eaf9a954a5800df04442b23ef9/test/integration/frame.cc#L743-L762

marcoag: Sure, changed to this format.

test/integration/error_output.cc
```diff
@@ -0,0 +1,150 @@
...
+// Test Param class for sdf::Errors outputs
+TEST(Error, ErrorOutput)
+{
+  std::stringstream buffer;
+  std::streambuf * old = std::cerr.rdbuf(buffer.rdbuf());
+  sdf::Errors errors;
+  sdf::Error error;
+  EXPECT_NO_THROW(sdf::Param param1("key", "not_valid_type", "true", false,
+                                    errors, "description"));
+  error = errors.front();
```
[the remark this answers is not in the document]
marcoag: Sure, changed `EXPECT`s to `ASSERT`.

test/integration/error_output.cc
```diff
@@ -0,0 +1,132 @@
...
+TEST(Error, ErrorOutput)
+{
+  std::stringstream buffer;
+  sdf::testing::RedirectConsoleStream redir(
+    sdf::Console::Instance()->GetMsgStream(), &buffer);
+
+  sdf::Errors errors;
+  ASSERT_NO_THROW(sdf::Param param1("key", "not_valid_type", "true", false,
+                                    errors, "description"));
+  ASSERT_EQ(errors[0].Code(), sdf::ErrorCode::UNKNOWN_PARAMETER_TYPE);
```
azeey, line 40: Sorry, I only meant to use `ASSERT` to check the size of the `errors` object. Here is my suggestion https://github.com/ignitionrobotics/sdformat/pull/1010

azeey (approved):

Issue #31, opened 2013-10-23
Use common friction parameters across physics engines


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/31)) by John Hsu (Bitbucket: [hsu](https://bitbucket.org/%7B0a186eae-abf0-4514-a951-23db5eccc286%7D/), GitHub: [hsu](https://github.com/hsu)).**

----------------------------------------
Currently each physics engine has its own friction parameters. We should make some common parameters that each can use.

chapulina:
The DART implementation on Ignition Physics has been [using](https://github.com/ignitionrobotics/ign-physics/blob/56a04fe204afa77e6a16d84b27ec289823eb330a/dartsim/src/SDFFeatures.cc#L544-L554) the `<ode>` tag until this issue is resolved.

PR #955, opened 2022-04-07
Added Friction and ODE classes

ahcorde:

# 🎉 New feature

related with this issue https://github.com/ignitionrobotics/sdformat/issues/954

## Summary

Added Friction and ODE classes

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/Surface.hh
+++ b/include/sdf/Surface.hh
@@ -64,7 +64,7 @@ namespace sdf
     /// \brief Default constructor
     public: ODE();
 
-    /// \brief Load the contact based on a element pointer. This is *not* the
+    /// \brief Load the ODE based on a element pointer. This is *not* the
     /// usual entry point. Typical usage of the SDF DOM is through the Root
     /// object.
     /// \param[in] _sdf The SDF Element pointer
@@ -90,18 +90,18 @@ namespace sdf
 
     /// \brief Get the fdir
     /// \returns ODE fdir
-    public: const ignition::math::Vector3d Fdir1() const;
+    public: const ignition::math::Vector3d &Fdir1() const;
 
     /// \brief Set fdir
-    /// \param[in] _mu2 ODE fdir
-    public: void SetFdir1(ignition::math::Vector3d _fdir);
+    /// \param[in] _fdir ODE fdir
+    public: void SetFdir1(const ignition::math::Vector3d &_fdir);
 
-    /// \brief Get the Mu2
-    /// \returns ODE mu2
+    /// \brief Get the slip1
+    /// \returns ODE slip1
     public: double Slip1() const;
 
     /// \brief Set Slip1
-    /// \param[in] _mu2 ODE Slip1
+    /// \param[in] _slip1 ODE Slip1
     public: void SetSlip1(double _slip1);
 
     /// \brief Get the Slip2
@@ -109,7 +109,7 @@ namespace sdf
     public: double Slip2() const;
 
     /// \brief Set Slip2
-    /// \param[in] _mu2 ODE Slip2
+    /// \param[in] _slip2 ODE Slip2
     public: void SetSlip2(double _slip2);
 
     /// \brief Get a pointer to the SDF element that was used during
@@ -128,7 +128,7 @@ namespace sdf
     /// \brief Default constructor
     public: Friction();
 
-    /// \brief Load the contact based on a element pointer. This is *not* the
+    /// \brief Load the friction based on a element pointer. This is *not* the
     /// usual entry point. Typical usage of the SDF DOM is through the Root
     /// object.
     /// \param[in] _sdf The SDF Element pointer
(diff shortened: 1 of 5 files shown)


scpeters:
@azeey this might be the right time to try to clean up our friction parameters, per https://github.com/ignitionrobotics/sdformat/issues/31 and specifically this comment: https://github.com/ignitionrobotics/sdformat/issues/31#issuecomment-951137292

azeey:
> @azeey this might be the right time to try to clean up our friction parameters, per #31 and specifically this comment: [#31 (comment)](https://github.com/ignitionrobotics/sdformat/issues/31#issuecomment-951137292)

I think @ahcorde is pressed for time and needs this for the USD-SDF work. But yeah, we might be able to make time for it in the next couple months.

scpeters:
> > @azeey this might be the right time to try to clean up our friction parameters, per #31 and specifically this comment: [#31 (comment)](https://github.com/ignitionrobotics/sdformat/issues/31#issuecomment-951137292)
> 
> I think @ahcorde is pressed for time and needs this for the USD-SDF work. But yeah, we might be able to make time for it in the next couple months.

oh I didn't look at the labels and thought this was for deepmind. Since it is creating an API, I'd like to think a little extra about its design

scpeters:
> oh I didn't look at the labels and thought this was for mujoco. Since this PR is creating an API, I'd like to think a little extra about its design

my main concern is that the fact that all friction parameters are in the `<ode>` block feels like a hack. We started trying to clean this up with the [FrictionPyramid](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/physics/SurfaceParams.hh#L34-L161) class in gazebo classic but never pushed that model structure into SDFormat.

so, do we want to encode the hacky SDFormat structure in our C++ API in libsdformat12 and then deprecate it in libsdformat13? The advantage of that would be that we could get it merged faster, and that it maintains a transparent mapping between the DOM API and the SDFormat data structure. Or do we want to try for a better API now?

scpeters:
> > oh I didn't look at the labels and thought this was for mujoco. Since this PR is creating an API, I'd like to think a little extra about its design
> 
> my main concern is that the fact that all friction parameters are in the `<ode>` block feels like a hack. We started trying to clean this up with the [FrictionPyramid](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/physics/SurfaceParams.hh#L34-L161) class in gazebo classic but never pushed that model structure into SDFormat.
> 
> so, do we want to encode the hacky SDFormat structure in our C++ API in libsdformat12 and then deprecate it in libsdformat13? The advantage of that would be that we could get it merged faster, and that it maintains a transparent mapping between the DOM API and the SDFormat data structure. Or do we want to try for a better API now?

after discussing offline, it sounds reasonable to proceed with this approach for `libsdformat12` since it matches the specification and deprecate it in `libsdformat13` if we change how the friction parameters are specified per #31

include/sdf/Surface.hh
```diff
@@ -58,6 +58,103 @@ namespace sdf
...
+    /// \param[in] _mu2 ODE mu2
+    public: void SetMu2(double _mu2);
+
+    /// \brief Get the fdir
+    /// \returns ODE fdir
+    public: const ignition::math::Vector3d Fdir1() const;
+
+    /// \brief Set fdir
+    /// \param[in] _mu2 ODE fdir
+    public: void SetFdir1(ignition::math::Vector3d _fdir);
```
azeey, line 97: The parameter should be a const ref.

src/Surface.cc
```diff
@@ -33,15 +33,248 @@ class sdf::Contact::Implementation
...
+  // Check that sdf is a valid pointer
+  if (!_sdf)
+  {
+    errors.push_back({ErrorCode::ELEMENT_MISSING,
+        "Attempting to load a Friction, but the provided SDF "
+        "element is null."});
+    return errors;
+  }
+
+  // Check that the provided SDF element is a <contact>
```
azeey, line 240: `<contact>` -> `<friction>`.

src/Surface.cc
```diff
@@ -33,15 +33,248 @@ class sdf::Contact::Implementation
...
+    return errors;
+  }
+
+  if (_sdf->HasElement("ode"))
+  {
+    Errors err = this->dataPtr->ode.Load(_sdf->GetElement("ode"));
+    errors.insert(errors.end(), err.begin(), err.end());
+  }
+
+  // \todo(nkoenig) Parse the remaining collide properties.
```
azeey, line 256: Is this TODO needed? If so, can you update it?

src/Surface_TEST.cc
```diff
@@ -126,3 +190,242 @@ TEST(DOMcontact, CollideBitmask)
   EXPECT_EQ(contact.CollideBitmask(), 0x67);
 }
 
+/////////////////////////////////////////////////
+TEST(DOMfriction, DefaultConstruction)
```
azeey, line 194: Since the test is named `DefaultConstruction`, I think it should test the default values, not ones set using the `Set*` functions.

src/Surface_TEST.cc
```diff
@@ -126,3 +190,242 @@ TEST(DOMcontact, CollideBitmask)
...
+  EXPECT_DOUBLE_EQ(friction1.ODE()->Slip1(), 3);
+  EXPECT_DOUBLE_EQ(friction1.ODE()->Slip2(), 4);
+  EXPECT_EQ(friction1.ODE()->Fdir1(),
+            ignition::math::Vector3d(1, 2, 3));
+}
+
+/////////////////////////////////////////////////
+TEST(DOMode, DefaultConstruction)
+{
+  sdf::ODE ode;
```
azeey, line 325: Since the test is named `DefaultConstruction`, I think it should test the default values, not ones set using the `Set*` functions.

azeey (changes requested):

src/Surface.cc
azeey: Can you update the username?

azeey (approved):

PR #958, opened 2022-04-08
Backport ``ign sdf --inertial-stats``

adityapande-1995:

# 🎉 New feature

## Summary
Backport of https://github.com/ignitionrobotics/sdformat/pull/936

## Test it
Added a test case to ``ign_TEST.cc`` and an example model sdf file.

## Checklist
- [x] Added tests

### DIFF
--- a/src/ign.cc
+++ b/src/ign.cc
@@ -178,7 +178,6 @@ extern "C" SDFORMAT_VISIBLE int cmdInertialStats(
     {
       std::cerr << "Error: " << error.Message() << std::endl;
     }
-    return -1;
   }
 
   if (root.WorldCount() > 0)
--- a/src/ign.cc
+++ b/src/ign.cc
@@ -253,7 +253,7 @@ extern "C" SDFORMAT_VISIBLE int cmdInertialStats(
     size_t spacePadding = maxLength - moiVector[i].size();
     // Print the matrix element
     std::cout << moiVector[i];
-    for (int j = 0; j < spacePadding; j++)
+    for (size_t j = 0; j < spacePadding; j++)
     {
       std::cout << " ";
     }
--- a/src/ign.cc
+++ b/src/ign.cc
@@ -16,6 +16,7 @@
 */
 
 #include <iostream>
+#include <string>
 #include <string.h>
 #include <vector>
 
--- a/src/ign.cc
+++ b/src/ign.cc
@@ -17,12 +17,17 @@
 
 #include <iostream>
 #include <string.h>
+#include <vector>
 
 #include "sdf/sdf_config.h"
 #include "sdf/Filesystem.hh"
+#include "sdf/Link.hh"
+#include "sdf/Model.hh"
 #include "sdf/Root.hh"
 #include "sdf/parser.hh"
 #include "sdf/system_util.hh"
+
+#include "ignition/math/Inertial.hh"
 
 #include "ign.hh"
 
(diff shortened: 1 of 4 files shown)


src/ign.cc
```diff
@@ -153,3 +159,114 @@ extern "C" SDFORMAT_VISIBLE int cmdPrint(const char *_path)
...
+
+  sdf::Root root;
+  sdf::Errors errors = root.Load(_path);
+  if (!errors.empty())
+  {
+    for (auto &error : errors)
+    {
+      std::cerr << "Error: " << error.Message() << std::endl;
+    }
+    return -1;
```
scpeters, line 181: I think we don't need to return here. There can be minor errors that wouldn't affect the inertial calculations. It's nice to print the errors, but just skip the `return` and that should be fine

jennuine: Wouldn't it be safer to return in case the error isn't minor? For example, it won't catch [inertial_invalid.sdf](https://github.com/ignitionrobotics/sdformat/blob/sdf9/test/sdf/inertial_invalid.sdf)

adityapande-1995: Yeah that just broke a test case. ``ign sdf --inertial-stats`` should inherently check for sdf errors I think, and if those pass only then we should return the statistics. Maybe that is safer for users.

scpeters: it currently doesn't work for the VIPER urdf. That might be fixed by #959, but it's not very helpful for debugging if it doesn't at least try to give you some inertial information if there's anything at all wrong with the model. We could add a `--strict` flag to make this configurable if we can't agree, but I think printing the warning messages should be enough of a clue that there's something wrong

adityapande-1995: Updated the test case. On the other hand maybe printing out the false inertial stats can help with debugging.

src/ign.cc
```diff
@@ -153,3 +159,113 @@ extern "C" SDFORMAT_VISIBLE int cmdPrint(const char *_path)
...
+  std::cout << "---" << std::endl;
+
+  std::cout << "Moment of inertia matrix: " << std::endl;
+
+  // Pretty print the MOI matrix
+  std::stringstream ss;
+  ss << totalInertial.Moi();
+
+  std::string s;
+  auto maxLength = 0u;
```
scpeters, line 240: there is a windows compiler warning on line 247. try `size_t` instead of `auto`?

* https://build.osrfoundation.org/job/sdformat-ci-pr_any-windows7-amd64/3612/msbuild/

src/ign.cc
```diff
@@ -153,3 +159,113 @@ extern "C" SDFORMAT_VISIBLE int cmdPrint(const char *_path)
...
+    moiVector.push_back(s);
+    if (s.size() > maxLength)
+    {
+      maxLength = s.size();
+    }
+  }
+
+  for (int i = 0; i < 9; i++)
+  {
+    int spacePadding = maxLength - moiVector[i].size();
```
scpeters, line 253: windows compiler warning about this line

* https://build.osrfoundation.org/job/sdformat-ci-pr_any-windows7-amd64/3612/msbuild/

~~~
'initializing': conversion from 'size_t' to 'int', possible loss of data
~~~

adityapande-1995: Done https://github.com/ignitionrobotics/sdformat/pull/958/commits/6b532e3debef13c6fa5d8539057b7f6d8d57a43d

scpeters (approved):

scpeters:
> The ABI checker failed weirdly. Seems unrelated.

yes, it's unrelated. I'll merge this now

PR #966, opened 2022-04-12
Merge 12 -> main

scpeters:
# ➡️ Forward port

Port sdf12 to main

Branch comparison: https://github.com/ignitionrobotics/sdformat/compare/main...sdf12


CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ if (BUILD_SDF)
   ign_configure_project(
     NO_IGNITION_PREFIX
     REPLACE_IGNITION_INCLUDE_PATH sdf
-    VERSION_SUFFIX pre1)
+    VERSION_SUFFIX)
```
chapulina, line 38: We shouldn't change this

CMakeLists.txt
```diff
@@ -117,6 +117,7 @@ if (BUILD_SDF)
   ########################################
   # Find ignition utils
   ign_find_package(ignition-utils1 VERSION REQUIRED)
+  ign_find_package(ignition-utils1 COMPONENTS cli REQUIRED_BY usd)
```
chapulina, line 120: Can these calls be combined?

chapulina (changes requested):

CMakeLists.txt
scpeters: I think they need to be in separate calls if there are different keywords. Arguably the `cli` component could be `PRIVATE` since it is just used to build executables, but we haven't tried to mark that here. So I think it would be fine to combine these two lines into:

~~~
ign_find_package(ignition-utils1 REQUIRED COMPONENTS cli)
~~~

CMakeLists.txt
scpeters: fixed via amended commit in https://github.com/ignitionrobotics/sdformat/pull/966/commits/1943c11820c1f1fdb7c88457ce83f30c4f790327

CMakeLists.txt
scpeters: fixed in https://github.com/ignitionrobotics/sdformat/pull/966/commits/5600ab6b92dd9d7ee322f1d64725ccb8a5c41358

chapulina (approved):

PR #1000, opened 2022-04-22
Polyline geometry DOM

chapulina:
# 🎉 New feature

* Closes https://github.com/ignitionrobotics/sdformat/issues/296

## Summary

Expose polylines through the C++ API.

Polyline is different from other geometries because the user can specify more than one of them. So I'm proposing here an API that's slightly different and returns an `std::vector<>` as opposed to the raw pointers returned for the other geometries. The API will look slightly different from other geometries, but I think it would need to be different anyway.

## Test it

See the added tests. 

There's also this Gazebo PR:

* https://github.com/ignitionrobotics/ign-gazebo/pull/1456

## Checklist
- [x] Added tests
- [x] ~~Updated migration guide (as needed)~~

### DIFF
--- a/include/sdf/Polyline.hh
+++ b/include/sdf/Polyline.hh
@@ -96,6 +96,11 @@ namespace sdf
     /// \brief Remove all points from the polyline.
     public: void ClearPoints();
 
+    /// \brief Get the polyline's points. Each point has 2D coordinates in
+    /// meters.
+    /// \return The polyline's points.
+    public: const std::vector<ignition::math::Vector2d> &Points() const;
+
     /// \brief Get a pointer to the SDF element that was used during
     /// load.
     /// \return SDF element pointer. The value will be nullptr if Load has
--- a/include/sdf/Polyline.hh
+++ b/include/sdf/Polyline.hh
@@ -81,7 +81,7 @@ namespace sdf
     public: uint64_t PointCount() const;
 
     /// \brief Get a point by its index.
-    /// \return Constand pointer to the point.
+    /// \return Constant pointer to the point.
     public: const ignition::math::Vector2d *PointByIndex(uint64_t _index) const;
 
     /// \brief Get a point by its index.
--- a/include/sdf/Polyline.hh
+++ b/include/sdf/Polyline.hh
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+#ifndef SDF_POLYLINE_HH_
+#define SDF_POLYLINE_HH_
+
+#include <ignition/math/Vector2.hh>
+#include <sdf/Error.hh>
+#include <sdf/Element.hh>
+#include <sdf/sdf_config.h>
+
+namespace sdf
+{
+  // Inline bracket to help doxygen filtering.
+  inline namespace SDF_VERSION_NAMESPACE {
+  //
+
+  // Forward declare private data class.
+  class PolylinePrivate;
+
+  /// \brief Polyline represents a 2D path. Multiple polylines can be combined
+  class SDFORMAT_VISIBLE Polyline
+  {
+    /// \brief Constructor
+    public: Polyline();
+
+    /// \brief Copy constructor
+    /// \param[in] _polyline Polyline to copy.
+    public: Polyline(const Polyline &_polyline);
+
+    /// \brief Move constructor
+    /// \param[in] _polyline Polyline to move.
+    public: Polyline(Polyline &&_polyline) noexcept;
+
+    /// \brief Destructor
+    public: virtual ~Polyline();
+
+    /// \brief Assignment operator.
+    /// \param[in] _polyline The polyline to set values from.
+    /// \return *this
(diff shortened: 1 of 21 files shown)


scpeters:
> Polyline is different from other geometries because the user can specify more than one of them.

This is correct, but I noticed that the `required` attribute is `0` instead of `*` in the spec where the poly line is included:

* https://github.com/ignitionrobotics/sdformat/blob/sdf12/sdf/1.9/geometry.sdf#L17

I guess this shows that we aren't following that attribute very closely, but perhaps we can update the spec to match the actual usage? (ie. I'm suggesting we change the required value from `0` to `*` on that line and in the other `sdf/1.*` folders)

chapulina:
> This is correct, but I noticed that the required attribute is 0 instead of * in the spec where the poly line is included:

I noticed that too. I think the required symbols can be a bit confusing, because `*` means "zero or more elements are required". I guess it would be clearer if it didn't have the word "required" 🤔 

https://github.com/ignitionrobotics/sdformat/blob/12e77821af7c2810c4d9d532ee8cb833cbb8e318/include/sdf/Element.hh#L109-L113

chapulina:
> I think the required symbols can be a bit confusing,

* Ticketed an issue here: https://github.com/ignitionrobotics/sdformat/issues/1002

> I'm suggesting we change the required value from 0 to * on that line

@scpeters , after you read that issue, I'd appreciate some feedback on whether you still think we should go ahead with this or if we should improve those required symbols before making any large changes like that.

scpeters:
> > I think the required symbols can be a bit confusing,
> 
> * Ticketed an issue here: [Enforce required elements #1002](https://github.com/ignitionrobotics/sdformat/issues/1002)
> 
> > I'm suggesting we change the required value from 0 to * on that line
> 
> @scpeters , after you read that issue, I'd appreciate some feedback on whether you still think we should go ahead with this or if we should improve those required symbols before making any large changes like that.

that's a fair point; let's not make any changes here to the spec

include/sdf/Polyline.hh
```diff
@@ -0,0 +1,99 @@
...
+    public: void SetHeight(const double _height);
+
+    /// \brief Get the polyline's points. Each point has 2D coordinates in
+    /// meters.
+    /// \return The polyline's points.
+    public: const std::vector<ignition::math::Vector2d> &Points() const;
+
+    /// \brief Add a point to the polyline.
+    /// \param[in] _point 2D point to add.
+    public: void AddPoint(const ignition::math::Vector2d &_point);
```
scpeters, line 86: this is useful for incremental construction, but it doesn't allow removing or modifying points. Should we also provide a `SetPoints(const std::vector<ignition::math::Vector2d>&)` method?

chapulina: Yeah I was looking for this kind of feedback before adding a bunch of methods. I was just catching up on the [DOM API](https://github.com/ignitionrobotics/sdf_tutorials/blob/9ca00b9bcd3d698ad9e0c986b4cdb065e25aaf73/developers/dom_api.md#current-state-of-mutability) documentation and it looks like things are still up in the air.

Do you think `SetPoints` is enough? Should I remove `AddPoint`? Should I add other functions, maybe `ClearPoints`?

scpeters: ok, now that I look more closely at our API, I see that we don't typically return a `std::vector` like this. Based on [Link.hh](https://github.com/ignitionrobotics/sdformat/blob/sdf12/include/sdf/Link.hh), I would expect the following API here:

~~~
    public: uint64_t PointCount() const;

    public: const ignition::math::Vector2d *PointByIndex(const uint64_t _index) const;

    public: ignition::math::Vector2d *PointByIndex(const uint64_t _index);

    public: bool AddPoint(const ignition::math::Vector2d &_collision);

    public: void ClearPoints();
~~~

chapulina: Sounds good, implemented that API in [b376f96](https://github.com/ignitionrobotics/sdformat/pull/1000/commits/b376f964e16c1fe1eaad9b8806e6f33fd2af9fc4)

chapulina: As I'm updating Gazebo to use the new API, I realize how inconvenient it is not to have the `const std::vector<...> &Points()` function. Would you be ok bringing it back? I'm not sure what's the motivation for hiding the vector from the public API and instead handling raw pointers. I think we could make better use of the containers provided by the standard, which provide their own APIs for iterating, appending, accessing elements, etc.

chapulina: See what changes were needed to use the new API here: a89c6a028e0b930323a49b7c5c7ac2dc0664e14c

The vector was nice because it matched the Common API for extruding the polylines, but that's not enough reason to drive the API design here.

include/sdf/Polyline.hh
```diff
@@ -0,0 +1,99 @@
...
+
+    /// \brief Get a pointer to the SDF element that was used during
+    /// load.
+    /// \return SDF element pointer. The value will be nullptr if Load has
+    /// not been called.
+    public: sdf::ElementPtr Element() const;
+
+    /// \brief Private data pointer.
+    private: PolylinePrivate *dataPtr;
```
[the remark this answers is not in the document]
chapulina: Here's the forward-port:

* https://github.com/ignitionrobotics/sdformat/pull/1003

(Opened while this is still being iterated on because I actually need to use this on Fortress)

include/sdf/Polyline.hh
scpeters: ok, well it looks like we've started using `std::vector` in the DOM API:

* https://github.com/ignitionrobotics/sdformat/blob/sdf12/include/sdf/Plugin.hh#L102

I think it's fine to have an API that returns an `std::vector`; please feel free to restore it. Sorry for the runaround.

chapulina: Done in [c8cafb6](https://github.com/ignitionrobotics/sdformat/pull/1000/commits/c8cafb6bf2a56258373dc3569f716bce0f463d12)

scpeters (approved):
fixed minor spelling error in https://github.com/ignitionrobotics/sdformat/pull/1000/commits/7a34fab4b39b4f59d3b3ad7d90f312800d1b19cb; thanks for iterating, looks good

PR #1006, opened 2022-05-01
Camera: fix default trigger topic

scpeters:
# 🦟 Bug fix

Fixes default trigger topic when unset

## Summary

In https://github.com/ignitionrobotics/ign-sensors/pull/215#discussion_r855757419 I noticed that if the trigger topic name is not explicitly set in SDFormat that ign-sensors uses `__default__/trigger` as the default trigger topic name. I believe this is due to the way the DOM object is loaded, so I have adjusted it in https://github.com/ignitionrobotics/sdformat/commit/361fa6aa8d7083cb1c0c50b291bd3b5e522cb3ae to only read the value of the `trigger_topic` element if that element exists. Thus it will return an empty string `""` if no trigger topic is specified, leaving the default behavior up to the downstream library (ign-sensors in the case I noticed).

## Checklist
- [ ] Added tests

### DIFF
--- a/src/Camera.cc
+++ b/src/Camera.cc
@@ -228,8 +228,11 @@ Errors Camera::Load(ElementPtr _sdf)
   this->dataPtr->triggered = _sdf->Get<bool>("triggered",
       this->dataPtr->triggered).first;
 
-  this->dataPtr->triggerTopic = _sdf->Get<std::string>("trigger_topic",
-      this->dataPtr->triggerTopic).first;
+  if (_sdf->HasElement("trigger_topic"))
+  {
+    this->dataPtr->triggerTopic = _sdf->Get<std::string>("trigger_topic",
+        this->dataPtr->triggerTopic).first;
+  }
 
   this->dataPtr->hfov = _sdf->Get<ignition::math::Angle>("horizontal_fov",
       this->dataPtr->hfov).first;
--- a/src/Camera.cc
+++ b/src/Camera.cc
@@ -228,11 +228,8 @@ Errors Camera::Load(ElementPtr _sdf)
   this->dataPtr->triggered = _sdf->Get<bool>("triggered",
       this->dataPtr->triggered).first;
 
-  if (_sdf->HasElement("trigger_topic"))
-  {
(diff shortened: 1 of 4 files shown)


src/Camera.cc
```diff
@@ -228,8 +228,11 @@ Errors Camera::Load(ElementPtr _sdf)
   this->dataPtr->triggered = _sdf->Get<bool>("triggered",
       this->dataPtr->triggered).first;
 
-  this->dataPtr->triggerTopic = _sdf->Get<std::string>("trigger_topic",
-      this->dataPtr->triggerTopic).first;
+  if (_sdf->HasElement("trigger_topic"))
+  {
+    this->dataPtr->triggerTopic = _sdf->Get<std::string>("trigger_topic",
```
chapulina, line 233: I understood we shouldn't need to call `HasElement` before using the `Get` call that passes a default value. If we call `HasElement`, how will it ever get to use the default value?

It looks like the use of `__default__` was intentional, judging by the documentation, should that be updated too?

https://github.com/ignitionrobotics/sdformat/blob/12e77821af7c2810c4d9d532ee8cb833cbb8e318/sdf/1.9/camera.sdf#L12-L14

WilliamLewww: I think I used `__default__/trigger` since a camera with no name is given the name `__default__`. I made it so that */trigger uses the name of the camera to create the prefix (i.e. /camera123 -> /camera123/trigger).

There is probably a better way to handle the camera name rather than to use `__default__/trigger`.

https://github.com/ignitionrobotics/sdformat/blob/12e77821af7c2810c4d9d532ee8cb833cbb8e318/sdf/1.9/camera.sdf#L4-L6

scpeters: I think the default value of `trigger_topic` should be an empty string and leave the behavior up to a downstream library (`ign-sensors`). I will update the documentation accordingly.

scpeters: https://github.com/ignitionrobotics/sdformat/pull/1006/commits/2948c0932465768abc2dc8e247beedad16fd41de

WilliamLewww (approved):

PR #858, opened 2022-02-19
Use the Plugin DOM in other DOM objects

nkoenig:
Signed-off-by: Nate Koenig <nate@openrobotics.org>

# 🎉 New feature

## Summary

A lot of file have changed, but most of the changes all follow an identical pattern.

There is a `Plugin` DOM object, but only the `GUI` DOM object was using it. This is particularly problematic when attempting to build and use a programmatically created DOM. For example, Ignition Gazebo is currently relying on the `Element()` function(s) to access plugin information. This will not work for programmatically create DOM objects.

I've also added input and output stream operators to the `Plugin` class. This is here to facilitate the use of [VectorSerialization](https://github.com/ignitionrobotics/ign-gazebo/blob/91a79da407b913b1e6bc6b96739a80f6e7254f18/include/ignition/gazebo/components/Serialization.hh#L231). This is also the reason why I choose to access a `std::vector` of `Plugins` rather than use the `ByIndex` and `ByName` pattern.

Example usage in https://github.com/ignitionrobotics/ign-gazebo/pull/1352.

## Test it
Run the tests.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #1021, opened 2022-05-18
Extract only one plugin with extraction operator

chapulina:
# 🦟 Bug fix

* Follow up to https://github.com/gazebosim/sdformat/pull/858
* Needed by https://github.com/gazebosim/gz-sim/pull/1404

## Summary

Plugin's extraction operator was consuming the entire input stream, although it should only be taking the first `<plugin>` element. 

This PR fixes the most basic use case, which requires the element to be exactly closed with `</plugin>`. I also didn't do a lot of sanity checking, like does it start with `<plugin>`, or is there a `<plugin>` nested inside another, etc. I feel like there must be a more reliable way of doing this, I'm all ears for ideas.

This operator is used on `gz-sim` to deserialize an `sdf::Plugins` object (i.e. `std::vector<sdf::Plugin>`), and the serialization is just concatenating them right now, using the [VectorSerializer](https://github.com/gazebosim/gz-sim/blob/f4198be19ae432bbcc584b73887178320e69ed90/include/ignition/gazebo/components/Serialization.hh#L231-L262).

## Checklist
- [x] Added tests


nkoenig:
I improved the stream extraction operator, and added more tests. See [135e251](https://github.com/gazebosim/sdformat/pull/1021/commits/135e25115636d94385a1307a411d82daf471871c)

chapulina:
Thanks for the updates, @nkoenig , LGTM. I just need an approval now.

include/sdf/parser.hh
```diff
@@ -458,6 +458,15 @@ namespace sdf
...
   std::string computeMergedModelProxyFrameName(const std::string &_modelName);
+
+  /// \brief A function that recursively reads XML elements from a stream and
+  /// returns the XML content as a string.
+  /// \param[in,out] _in Input stream to read from. This function will
+  /// consume the data on the stream from the start of the XML content to the
+  /// closing tag.
+  /// \return String that contains the XML content.
+  SDFORMAT_VISIBLE
+  std::string readXMLStream(std::istream &_in);
```
scpeters, line 469: it seems that this new function is only used internally, so I would propose adding it to `src/Utils.hh` to keep the surface area of parser.hh from growing further. This will require moving the stream operator implementations to Plugin.cc, which is a little tricky, but I've done it in https://github.com/gazebosim/sdformat/commit/22e99121d649103533106d15d24eaebc2f0052c8, which you should be able to cherry-pick to this branch

scpeters: I've cherry-picked that commit and hidden the function in `src/Utils.hh` in https://github.com/gazebosim/sdformat/pull/1021/commits/0a7241d4d85b7e2b3203af6abb35a27bb9540737

scpeters:
the API checker complaint is befuddling since `SDFImpl.hh` was not touched

* https://build.osrfoundation.org/job/sdformat-abichecker-any_to_any-ubuntu_auto-amd64/2282/API_5fABI_20report/

Screen Shot 2022-07-27 at 4 56 21 PM

src/Utils.cc
```diff
@@ -362,5 +362,87 @@ void copyChildren(ElementPtr _sdf, tinyxml2::XMLElement *_xml,
     }
   }
 }
+
+/////////////////////////////////////////////////
+std::string readXMLStream(std::istream &_in)
```
azeey, line 367: Would it be possible to use `tinyXML` to do this instead of manually parsing the stream? This seems brittle and I'm not sure if it handles escapes and `cdata` content.

include/sdf/parser.hh
scpeters: I've moved this to a separate pull request in https://github.com/gazebosim/sdformat/pull/1111 to see if that fixes the ABI complaints

azeey:
Can we close this since https://github.com/gazebosim/gz-sim/pull/1404 was merged without needing this PR?

PR #1027, opened 2022-05-23
Get the world name

ahcorde:

# 🎉 New feature

## Summary

Get just the world name instead of reading the whole file. This is required by this other PR https://github.com/gazebosim/gz-sim/pull/1484

## Checklist
- [x] Added tests


src/Root.cc
```diff
@@ -140,6 +140,29 @@ Root::Root()
...
+  tinyxml2::XMLDocument doc;
+  doc.LoadFile(_filename.c_str());
+  tinyxml2::XMLElement * pRootElement = doc.RootElement();
+  if (NULL != pRootElement)
+  {
+    tinyxml2::XMLElement * pworld = pRootElement->FirstChildElement("world");
+    if(pworld)
+    {
+      _worldName = pworld->Attribute("name");
+    }
```
chapulina, line 156: Error on `else`?

src/Root_TEST.cc
```diff
@@ -65,6 +66,16 @@ TEST(DOMRoot, MoveAssignmentOperator)
   EXPECT_EQ("test_root", root2.Version());
 }
 
+TEST(DOMRoot, GetWorldName)
```
chapulina, line 69: It would be interesting to test some failure cases

include/sdf/Root.hh
```diff
@@ -58,13 +58,13 @@ namespace sdf
     /// \brief Default constructor
     public: Root();
 
-    /// \brief Get the name of the world
+    /// \brief Get the name of the world without loading the entire world
```
azeey, line 61: Can you add that users shouldn't normally need to use this API. Also please mention that since this doesn't load the world, it might return the world name even if the file is not a valid SDFormat file.

azeey (commented):
In general, I have reservations about this approach because it doesn't go through the normal error checking process of the parser. It also results in loading the file into TinyXML twice, the second time when `sdf::Root::Load` is called. 

Here are some other ideas to consider:

* Avoid the need for a world name in the first place. Maybe we can just use the global topic namespace for single worlds and have world-specific namespaces when we add support for multiple worlds. BTW, this implementation of `WorldName` doesn't return all the world names in a file. What would be our plan when we want to add support for multiple worlds?
* Generate a unique topic namespace for a given world during load (eg. `/world/UUID/topic/name`) and then have some sort of topic and service remapping function in ign transport that maps `/world/<world name>` to `/world/UUID` once the world name is obtained from the fully loaded world.
* Implement a partial parser in SDFormat where `include` files are not processed. This would still suffer from the error checking issue I mentioned, but it wouldn't load the file twice.

include/sdf/Root.hh
```diff
@@ -58,13 +58,13 @@ namespace sdf
...
     /// \param[in] _filename Name of the SDF file to parse.
     /// \param[out] _worldName The name of the world
     /// \return Errors, which is a vector of Error objects. Each Error includes
     /// an error code and message. An empty vector indicates no error.
-    public: Errors GetWorldName(const std::string &_filename,
-                                std::string &_worldName);
+    public: Errors WorldName(const std::string &_filename,
```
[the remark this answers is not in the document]
ahcorde: https://github.com/gazebosim/sdformat/pull/1027/commits/a7a68576a6380d4bb670de5cb8fc79851e37b713

include/sdf/Root.hh
ahcorde: https://github.com/gazebosim/sdformat/pull/1027/commits/a7a68576a6380d4bb670de5cb8fc79851e37b713

src/Root.cc
azeey: Is this resolved?

azeey (approved):
Approving with my reservation stated above.

include/sdf/Root.hh
```diff
@@ -58,6 +59,17 @@ namespace sdf
...
 
+    /// \brief Get the name of the world without loading the entire world
+    /// Users shouldn't normally need to use this API.
+    /// This doesn't load the world, it might return the world name even if the
+    /// file is not a valid SDFormat file.
+    /// \param[in] _filename Name of the SDF file to parse.
+    /// \param[out] _worldNames A vector with world names
+    /// \return Errors, which is a vector of Error objects. Each Error includes
+    /// an error code and message. An empty vector indicates no error.
+    public: Errors WorldNameFromFile(const std::string &_filename,
```
scpeters, line 70: nit: should this be `WorldNamesFromFile` (plural)?

ahcorde: https://github.com/gazebosim/sdformat/pull/1027/commits/a6c9ef3e32dc5500fbb57c1b55c7247e021baf51

scpeters (approved):
LGTM with green CI

PR #1043, opened 2022-06-07
Don't include the gz/math.hh header from library code

jwnimmer-tri:
# 🦟 Bug fix

## Summary

Don't include the gz/math.hh header from library code. Using overly-broad include statements leads to slower builds and bloated object code. This commit swaps it out the `gz/math.hh` "everything" header for more targets files, instead.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests (n/a)
- [x] Updated documentation (n/a)
- [x] Updated migration guide (n/a)
- [x] Consider updating Python bindings (n/a)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #1044, opened 2022-06-07
Backport: Don't include the gz/math.hh header from library code (#1043)

azeey:
Backport of #1043 with additional changes for the `ignition` -> `gz` rename.

### DIFF
--- a/include/sdf/Param.hh
+++ b/include/sdf/Param.hh
@@ -30,7 +30,12 @@
 #include <variant>
 #include <vector>
 
-#include <ignition/math.hh>
+#include <ignition/math/Angle.hh>
+#include <ignition/math/Color.hh>
+#include <ignition/math/Pose3.hh>
+#include <ignition/math/Quaternion.hh>
+#include <ignition/math/Vector2.hh>
+#include <ignition/math/Vector3.hh>
 
 #include "sdf/Console.hh"
 #include "sdf/sdf_config.h"
--- a/include/sdf/Link.hh
+++ b/include/sdf/Link.hh
@@ -19,6 +19,7 @@
 
 #include <memory>
 #include <string>
+#include <ignition/math/Inertial.hh>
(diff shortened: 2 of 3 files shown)


scpeters:
we should make sure that gazebo11 and ignition-citadel packages still compile with this change

azeey:
> we should make sure that gazebo11 and ignition-citadel packages still compile with this change

Good point. I tested with citadel and there were some build failures in gazebo (gz-sim). Should I fix those or is this change not backportable?

scpeters:
> > we should make sure that gazebo11 and ignition-citadel packages still compile with this change
> 
> Good point. I tested with citadel and there were some build failures in gazebo (gz-sim). Should I fix those or is this change not backportable?

let's start by fixing any bugs we find, and then we can decided on whether to backport

scpeters:
gazebo11 needs some changes: https://github.com/osrf/gazebo/pull/3227

include/sdf/Param.hh
```diff
@@ -30,7 +30,12 @@
 #include <variant>
 #include <vector>
 
-#include <ignition/math.hh>
+#include <ignition/math/Angle.hh>
+#include <ignition/math/Color.hh>
+#include <ignition/math/Pose3.hh>
+#include <ignition/math/Quaternion.hh>
+#include <ignition/math/Vector2.hh>
+#include <ignition/math/Vector3.hh>
```
chapulina, line 38: Even though this is the right thing to do, we should minimize the chances of breaking downstream users on a stable version. That's why we decided to target https://github.com/gazebosim/gz-transport/pull/315 at Garden for example - that specific include change will definitely break lots of users. Here it could be argued that `Param.hh` is not as widely used as `Node.hh`, but I think the same principle applies.

chapulina (changes requested):

scpeters:
this may break downstream code, so let's skip it

PR #1053, opened 2022-06-19
Joint: rename parent/child *LinkName APIs

### LINKED ISSUE #979: Consider renaming Joint::ParentLinkName and Joint::ChildLinkName to Joint::ParentName and Joint::ChildName
## Desired behavior
As of SDFormat 1.8, joints can have frames as parent or child elements, so the functions `Joint::ParentLinkName` and `Joint::ChildLinkName` are no longer accurate. My suggestion would be to rename them to `Joint::ParentName` and `Joint::ChildName`.

## Alternatives considered
They could also be named `Joint::ParentFrameName` and `Joint::ChildFrameName` since all parent/child elements are frames, either explicit or implicit. However, I think this is a little more confusing if users are not familiar with the explicit/implicit frame concept.

scpeters:
# 🦟 Bug fix

Fixes #979

## Summary

Remove Link from the names of APIs for getting and setting the name of parent/child frames, since any frame can be specified, not just links.

## Checklist
- [ ] Added tests
- [X] Consider updating Python bindings (if the library has them)

### DIFF
--- a/python/src/sdf/pyJoint.cc
+++ b/python/src/sdf/pyJoint.cc
@@ -19,6 +19,8 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+#include <gz/utils/SuppressWarning.hh>
+
 #include "sdf/Joint.hh"
 #include "sdf/Sensor.hh"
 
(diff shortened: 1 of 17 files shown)


ahcorde (approved):

scpeters:
this will introduce some deprecation warnings in gz-physics and gz-sim after this is merged, but I'll fix that in follow-up pull requests

python/src/sdf/pyJoint.cc
```diff
@@ -34,6 +36,7 @@ namespace python
 void defineJoint(pybind11::object module)
 {
   pybind11::class_<sdf::Joint> jointModule(module, "Joint");
+  GZ_UTILS_WARN_IGNORE__DEPRECATED_DECLARATION
```
azeey, line 39: Should we even include the deprecated functions in the python API since we don't have a stable release of the bindings yet?

azeey (approved):

python/src/sdf/pyJoint.cc
scpeters: gz-mujoco is already using the deprecated functions, so I was going to leave them in until gz-mujoco can be updated to avoid breakage. Should I just remove it and then quickly fix it in a gz-mujoco PR?

azeey: Yeah, that sounds good to me.

scpeters: removed deprecated API from python binding in https://github.com/gazebosim/sdformat/pull/1053/commits/70023d623a8759be21804c08d71d8943aeeda113

PR #1054, opened 2022-06-21
Require eigen3 component of gz-math

scpeters:
# 🎉 New feature

Motivated by https://github.com/gazebosim/gz-math/pull/442

## Summary

Require and export the `eigen3` component of `gz-math` as a target dependency so that Eigen types may be used in the public API.

## Test it

Check that CI passes

## Checklist
- [ ] Added tests

### DIFF
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -1,6 +1,6 @@
 libignition-cmake3-dev
 libignition-common5-dev
-libignition-math7-dev
+libignition-math7-eigen3-dev
 libignition-tools2-dev
 libignition-utils2-dev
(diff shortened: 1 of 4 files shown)


chapulina (changes requested):
We're discussing this internally. I'd like to see a complete proposal for where Eigen is planned to be used and where, and what's the migration strategy for SDF and other libraries. There's a high risk of ending up in a mixed situation where parts of our API are using `gz-math` types and other parts are using `eigen`.

scpeters:
This is more contentious than I thought it would be. I will revisit in the future if the team reaches consensus to support this.

PR #1057, opened 2022-06-26
Migrate IGNITION- CMake variables

methylDragon:
https://github.com/gazebo-tooling/release-tools/issues/718


src/gz_TEST.cc
```diff
@@ -37,9 +37,9 @@ static std::string SdfVersion()
   return " --force-version " + std::string(SDF_VERSION_FULL);
 }
 
-static std::string IgnCommand()
+static std::string GzCommand()
 {
-  return std::string(GZ_PATH) + "/ign";
+  return std::string(GZ_PATH);
```
chapulina, line 42: The test failures seem to indicate we should have `/gz` here, but that would be different from the other libs. Gotta look more carefully to see what's going on

methylDragon: There was a

`get_filename_component(GZ_PATH ${GZ_PROGRAM} DIRECTORY)` which stripped the filename :>
I just used `GZ_PROGRAM` instead.

chapulina (approved):
Nice!

PR #1042, opened 2022-06-07
Bash completion for flags

mabelzhang:
# 🎉 New feature

Part of https://github.com/gazebosim/gz-tools/issues/1
Used together with https://github.com/gazebosim/gz-tools/pull/87

## Summary

Bash completion, installation, tests.

## Test it

```
$ . install/share/gz/gz1.completion
$ ign sdf -
--check           --describe        --force-version   --help            --inertial-stats  --print           --versions        -d                -h                -k                -p
```

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #1087, opened 2022-07-21
9 ➡️ 12

jennuine:
# ➡️ Forward port

Port `sdf9` to `sdf12`

Branch comparison: https://github.com/gazebosim/sdformat/compare/sdf12...sdf9

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,9 @@ if (BUILD_SDF)
   #################################################
   # Find ign command line utility:
   find_program(IGN_PROGRAM ign)
+  # Note that CLI files are installed regardless of whether the dependency is
+  # available during build time
+  set(IGN_TOOLS_VER 1)
 
   #################################################
   # Copied from catkin/cmake/empy.cmake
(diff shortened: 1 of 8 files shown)


cmake/SearchForStuff.cmake
```diff
@@ -0,0 +1,152 @@
+include (FindPkgConfig)
```
scpeters, line 1: we removed this `SearchForStuff.cmake` file when migrating to `ign-cmake2`. We don't want to re-add it. Let's try to figure out which pull request induced a change here and figure out how to apply that change elsewhere (probably the root CMakeLists.txt)

scpeters (changes requested):
don't add SearchForStuff.cmake

cmake/SearchForStuff.cmake
scpeters: looks like the changes were made in https://github.com/gazebosim/sdformat/pull/1042

jennuine: 9c81632bca096b86655a0501ca9e47be278ac609

scpeters (approved):

scpeters:
thanks!

Issue #820, opened 2022-01-12
Explicitly delimit EOM for logging


jwnimmer-tri:
## Desired behavior

The Console (i.e., logging) abstraction in libsdformat currently follows the semantics of something like `std::cerr` -- a simple text stream where the call sites conclude each call with a newline (`sdferr << ... << "\n";`).  Even if we switch to use the ignition-common console per #351, it still currently embodies the same abstraction.

However, in a lot of production code, log message telemetry is not just a dedicated console read by human eyes -- it is recorded in a structured form, sent over the network, and/or otherwise treated on a _message_ basis rather a _stream_ basis.

My request here is to change the Console to use the same semantics -- where there is a clear "end of message" demarcation.

Several C++ logging frameworks (e.g., log4cpp) do this by using the _logger lifetime_ to conclude the message.  My patch file in https://github.com/RobotLocomotion/drake/pull/16348 shows an example of this.  The `Console::ColorMsg` function returns a short-lived stream by-value, instead of a long-lived stream by-reference.  When the statement ends, the temporary is destroyed, and in its destructor the message is posted.

Another technique is what `spdlog` does -- the user makes a single function call to log, passing both the format string and all of the arguments all at once.  This would be a more significant rewrite of the calling-code than the stream lifetime change, but would be a better choice in case the `Console` needed to be used from languages beyond just C++.

## Alternatives considered

A different design would be to keep the streaming API unchanged, but to have the _implementation_ monitor the character buffer insertions for newlines, and replace all newlines with EOM / log posting.  I am not a fan of that design, because it removes the ability for a single log message to have embedded newlines, which I've often found to be useful.

## Additional context

For sdformat in particular, as I've explained in https://github.com/ignitionrobotics/sdformat/issues/334#issuecomment-683985626 the _library_ should never be printing to any console anyway.  The calls to parse should have a _structured_ way to report errors, and should be silent on `cerr`.  The current design and use of `sdf::Error` gets us most of the way there, but there are still a ton of places in the library code that stream to `sdferr` (or `sdfwarn`, etc.) instead of appending an `sdf::Error`.  If we fixed _that_ problem instead, then we could delete the sdformat-specific `Console` class and this whole discussion here is moot (though would still apply to the ignition-common logging API design).

azeey:
[...]

PR #1095, opened 2022-07-28
Element: add sdf::Errors output to API methods

marcoag:

# 🎉 New feature

Work towards #820

## Summary
Adds `sdf::Errors` for as an error report option to all the methods on the Element class.

## Test it
Calling the new functions that contain the `sdf::Errors` parameter should only add errors to the `sdf::Errors` data structure and not produce any printed output or throw any exceptions.

## Checklist
- [x] Added tests


include/sdf/Error.hh
```diff
@@ -85,6 +85,9 @@ namespace sdf
     /// expected, and an element of a different type was received.
     ELEMENT_INCORRECT_TYPE,
 
+    /// \Briev Generic error type for Elements
```
scpeters, line 88:
```suggestion
    /// \brief Generic error type for Elements
```

include/sdf/Element.hh
```diff
@@ -83,10 +83,20 @@ namespace sdf
...
+    public: ElementPtr Clone(sdf::Errors &_errors) const;
+
     /// \brief Copy values from an Element.
     /// \param[in] _elem Element to copy value from.
     public: void Copy(const ElementPtr _elem);
 
+    /// \brief Copy values from an Element.
+    /// \param[in] _elem Element to copy value from.
+    /// \param[out] _errors Vector of errors.
+    public: void Copy(const ElementPtr _elem, sdf::Errors &_errors);
```
scpeters, line 98: Since this is targeting `main`, we can change API if we do it this week. For any of these functions that currently return `void`, we could change the return type to `sdf::Errors` to avoid duplicating the functions

marcoag: The idea was to still maintain the old output as an option. If we have `sdf::Errors` as a return and only one function then I believe we would need a way to select one of the two ways of error report. I can think of a boolean flag but doesn't sound very appealing, any other way to solve this?

include/sdf/Error.hh
```diff
@@ -85,6 +85,9 @@ namespace sdf
     /// expected, and an element of a different type was received.
     ELEMENT_INCORRECT_TYPE,
 
+    /// \brief Generic error type for Elements
+    ELEMENT_ERROR,
```
scpeters, line 89: actually inserting an enum value breaks API; I'll split this into a separate PR

scpeters: #1099

src/Element.cc
```diff
@@ -170,6 +201,25 @@ ParamPtr Element::CreateParam(const std::string &_key,
...
 
+/////////////////////////////////////////////////
+ParamPtr Element::CreateParam(const std::string &_key,
+                              const std::string &_type,
+                              const std::string &_defaultValue,
+                              bool _required,
+                              sdf::Errors &_errors,
+                              const std::string &_description)
+{
+  ParamPtr param = std::make_shared<Param>(
```
azeey, line 212: This is duplicating code. See my comments on `AddValue`

src/Element.cc
```diff
@@ -181,8 +231,33 @@ void Element::AddAttribute(const std::string &_key,
...
 
+/////////////////////////////////////////////////
+void Element::AddAttribute(const std::string &_key,
+                           const std::string &_type,
+                           const std::string &_defaultValue,
+                           bool _required,
+                           sdf::Errors &_errors,
+                           const std::string &_description)
+{
+  this->dataPtr->attributes.push_back(
```
azeey, line 242: This is duplicating code. See my comments on `AddValue`.

src/Element.cc
```diff
@@ -941,6 +1109,18 @@ ElementPtr Element::GetElement(const std::string &_name)
   return result;
 }
 
+/////////////////////////////////////////////////
+ElementPtr Element::GetElement(const std::string &_name, sdf::Errors &_errors)
+{
+  ElementPtr result = this->GetElementImpl(_name);
```
azeey, line 1115: This is duplicating code. See my comments on `AddValue`.

src/gz_TEST.cc
```diff
@@ -1871,6 +1875,8 @@ TEST(inertial_stats, GZ_UTILS_TEST_DISABLED_ON_WIN32(SDF))
     std::string output =
       custom_exec_str(GzCommand() + " sdf --inertial-stats " +
                       path + SdfVersion());
+    std::cout << GzCommand() + " sdf --inertial-stats " +
```
azeey, line 1878: nit: Remove debug code?

test/integration/error_output.cc
```diff
@@ -106,7 +106,8 @@ TEST(Error, ErrorOutput)
   // will make reparse fail
   sdf::Param poseParam("", "pose", "1 2 3 0.4 0.5 0.6", false, "description");
   sdf::ElementPtr poseElem(new sdf::Element);
-  poseElem->AddAttribute("rotation_format", "string", "invalid_format", false);
+  poseElem->AddAttribute(
```
azeey, line 109: nit: Is this change necessary?

azeey (changes requested):
Looks good! I have some comments about deduplicating code and avoiding extra new member functions.

include/sdf/Element.hh
```diff
@@ -610,6 +698,51 @@ namespace sdf
...
+    /// \brief Create a new Param object and return it.
+    /// \param[in] _key Key for the parameter.
+    /// \param[in] _type String name for the value type (double,
+    /// int,...).
+    /// \param[in] _defaultValue Default value.
+    /// \param[in] _required True if the parameter is required to be set.
+    /// \param[out] _errors Vector of errors.
+    /// \param[in] _description Description of the parameter.
+    /// \return A pointer to the new Param object.
+    private: ParamPtr CreateParam(const std::string &_key,
```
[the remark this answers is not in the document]
marcoag: This one seems to use `SDF_ASSERT`, which throws an `sdf::AssertionInternalError`, changing the function to use `sdf::Error` instead will change behavior of any other function using `CreateParam`... dunno if we are ok with that, are we?

src/Element.cc
```diff
@@ -141,6 +141,17 @@ void Element::AddValue(const std::string &_type,
...
 }
 
+/////////////////////////////////////////////////
+void Element::AddValue(const std::string &_type,
+                       const std::string &_defaultValue,
+                       bool _required,
+                       sdf::Errors &_errors,
+                       const std::string &_description)
+{
+  this->dataPtr->value = this->CreateParam(this->dataPtr->name,
```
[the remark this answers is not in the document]
marcoag: Just realized that errors are sometimes using `SDF_ASSERT`  and others the `sdferr` stream. The first one throws an exception while the later prints using `sdf::Console`. Since these are two different behaviors I guess we can do either one of these:
* we don't care about changing this behavior and use either `sdferr` or `SDF_ASSERT` by default in the calling function.
* we add some "extra" info to `sdf::Errors`, maybe by creating a new enum that says `ASSERT_ERROR` or `SDFERR_...` so we can differentiate between both of them and keep the same behavior.

I also realized this might change behavior since a call to `SDF_ASSERT` will break the normal execution flow whereas if we add errors to `sdf::Errors` the function will keep executing. I think this can be solved with just a `return` and then with the throwing the `SDF_ASSERT` in the calling function but I just wanted to leave it here in case I miss something.

include/sdf/Element.hh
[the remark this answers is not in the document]
azeey: That's true. If Garden wasn't feature frozen, I'd say we should remove the SDF_ASSERT, but now that this is going in a minor release, I don't think we should change behavior like that. So, yeah, let's keep the second `CreateParam`, but I'd still think we should avoid duplicating code and write one `CreateParam` by calling the other.

src/Element.cc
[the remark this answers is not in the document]
azeey: Like I mentioned in another comment, I'd say let's keep the SDF_ASSERT. We can still call `AddValue` with a throwaway `Errors` object and call `SDF_ASSERT` if there are any errors.

chapulina:
Removed the `Garden` label because we're past feature freeze ❄️ .

This PR can either be merged into `main` and backported to `sdf13` after the stable release, or retargeted to `sdf13` and merged after the stable release (i.e. October).

src/Element.cc
```diff
@@ -975,6 +1155,30 @@ bool Element::HasElementDescription(const std::string &_name) const
 
 /////////////////////////////////////////////////
 ElementPtr Element::AddElement(const std::string &_name)
+{
+  ElementPtr elem = this->AddElementImpl(_name);
+  if (elem.get() == nullptr)
+  {
+    sdferr << "Missing element description for [" << _name << "]\n";
+  }
+  return elem;
```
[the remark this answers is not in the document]
marcoag: Yes, using the other other `AddElement` instead.

test/integration/error_output.cc
marcoag: no, removed.

include/sdf/Element.hh
```diff
@@ -610,6 +698,51 @@ namespace sdf
...
+    /// \brief Create a copy of this Element.
+    /// \param[out] _errors Vector of errors.
+    /// \return A copy of this Element, NULL if there was an error.
+    private: ElementPtr CloneImpl(sdf::Errors *const _errors = nullptr) const;
+
+    /// \brief Copy values from an Element.
+    /// \param[in] _elem Element to copy value from.
+    /// \param[out] _errors Vector of errors.
+    private: void CopyImpl(const ElementPtr _elem,
+                sdf::Errors *const _errors = nullptr);
```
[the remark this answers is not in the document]
marcoag: Yes, I'll do that.

src/gz_TEST.cc
marcoag: Removed.

src/Element.cc
marcoag: Removed duplicating code.

src/Element.cc
marcoag: Yes, fixed.

src/Element.cc
marcoag: Since this is private, I'll go ahead and change the function to just use `sdf::Errors`.

include/sdf/Element.hh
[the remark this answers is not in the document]
marcoag: Ok then. I went and created an `sdf::ErrorCode::THROW_ERROR` so the calling function can identify which errors to throw. Also added a `throwOrPrintErrors` function to `Utils.cc` to handle the printing or throwing as needed.

src/Element.cc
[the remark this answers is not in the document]
marcoag: Yes, will try to make sure we keep the same behavior.

src/Element.cc
```diff
@@ -1191,7 +1303,20 @@ void Element::RemoveFromParent()
 /////////////////////////////////////////////////
 void Element::RemoveChild(ElementPtr _child)
 {
-  SDF_ASSERT(_child, "Cannot remove a nullptr child pointer");
+  sdf::Errors errors;
+  RemoveChild(_child, errors);
```
azeey, line 1307: Missing `throwOrPrintErrors`

src/Utils.hh
```diff
@@ -85,6 +85,11 @@ namespace sdf
     const sdf::Error &_error,
     sdf::Errors &_errors);
 
+  /// \brief It will print the errors to sdferr or throw them using
+  /// SDF_ASSERT depeneing on their ErrorCode.
```
azeey, line 89: nit:
```suggestion
  /// SDF_ASSERT depending on their ErrorCode.
```

src/Utils.hh
```diff
@@ -85,6 +85,11 @@ namespace sdf
...
+  /// \param[in] _errors  The vector of errors.
+  void throwOrPrintErrors(const sdf::Errors _errors);
```
azeey, line 91: nit: use const ref
```suggestion
  void throwOrPrintErrors(const sdf::Errors& _errors);
```

test/integration/error_output.cc
```diff
@@ -146,3 +146,76 @@ TEST(Error, ErrorOutput)
...
+
+  sdf::Errors errors;
+  sdf::ElementPtr elem = std::make_shared<sdf::Element>();
+  elem->SetName("testElement");
+
+  elem->GetAny(errors, "test");
+  EXPECT_EQ(errors[0].Code(), sdf::ErrorCode::ELEMENT_ERROR);
+  EXPECT_NE(std::string::npos, errors[0].Message().find(
+    "Unable to find value for key [test]"));
+  ASSERT_EQ(errors.size(), 1u);
```
azeey, line 166: The `ASSERT` calls that check the size of the `errors` should come before indexing into it. Otherwise, we might get a segfault.

azeey (changes requested):

include/sdf/Error.hh
```diff
@@ -161,6 +161,10 @@ namespace sdf
     /// \brief The specified parameter (values of SDFormat elements
     /// or attributes) type is unknown.
     UNKNOWN_PARAMETER_TYPE,
+
+    /// \brief generic error to be thrown with SDF_ASSERT by the caller.
+    /// This has been created to help preserve behavior.
+    THROW_ERROR,
```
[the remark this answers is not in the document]
marcoag: Ok, sounds good.

src/Element.cc
marcoag: Added.

test/integration/error_output.cc
marcoag: True, my bad :man_facepalming:.

azeey (approved):
Great work!

azeey:
Can you fix the DCO failure? I think it's due to applied suggestions :(

marcoag:
Oh the suggestions DCO thing, I need to be more careful with that :man_facepalming:.

Fixed two out of the three DCOs failures. The last one, commit sha: [96286da](https://github.com/gazebosim/sdformat/pull/1095/commits/96286da3e45156ca9c1a63438d6f84a9ad66c3b7), gets tricky as there are other commits on top of it that I did not authored (see [history](https://github.com/gazebosim/sdformat/commits/marcoag/sdf_error_element)) so rebasing is not an option. The only solutions to this problem I know are either taking it as it is or creating a new PR with the cherry picked relevant commits, any preference?

chapulina:
> The only solutions to this problem I know are either taking it as it is or creating a new PR with the cherry picked relevant commits, any preference?

Another option is adding the other author's signature to the commit as you rebase. I often do that.

scpeters:
> > The only solutions to this problem I know are either taking it as it is or creating a new PR with the cherry picked relevant commits, any preference?
> 
> Another option is adding the other author's signature to the commit as you rebase. I often do that.

you have my permission to add the signature or to just squash it because it is a trivial change. I think rebasing would be a reasonable step

scpeters:
> Can you fix the DCO failure? I think it's due to applied suggestions :(

looks like this is resolved

PR #1105, opened 2022-08-09
ign -> gz : Remove redundant namespace references

methylDragon:
Part of: https://github.com/gazebo-tooling/release-tools/issues/784

### DIFF
--- a/src/Actor.cc
+++ b/src/Actor.cc
@@ -111,7 +111,7 @@ class sdf::ActorPrivate
   public: std::vector<Joint> joints;
 
   /// \brief The SDF element pointer used during load.
-  public: sdf::ElementPtr sdf;
+  public: ElementPtr sdf;
 };
 
 /////////////////////////////////////////////////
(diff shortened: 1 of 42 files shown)


scpeters:
I think the namespace reduction plan in https://github.com/gazebo-tooling/release-tools/issues/784 is useful for eliminating `ignition::` namespaces, but for this library it is just removing `sdf::` in lots of places, which looks like noise to me. I would rather just decline this.

chapulina (approved):
This PR isn't strictly helpful for the `ignition` migration because this repository doesn't release any code under that namespace, so we shouldn't be `using namespace ignition` here. 

But I think removing those redundant namespaces doesn't hurt, makes some liner shorter, and the code a bit cleaner. It may just introduce some confusion for functions that aren't clearly from this library, like `sdf::split`

chapulina:
I just saw @scpeters 's comment, I'm fine declining this PR if that's preferred.

methylDragon:
I have no preferences, but if needed, I can exclude stuff like `split` from the cleanup, it's not too difficult for me to do (like a 5-10 min job)

chapulina:
@methylDragon , I'll just close this one for expediency. We don't need to put more time into it. Let's focus on `ignition::` and `ignition::gazebo`.

PR #1123, opened 2022-08-31
Converter: add sdf::Errors output to API methods

marcoag:

# 🎉 New feature

Work towards https://github.com/gazebosim/sdformat/issues/820.

## Summary

It moves all error reporting in the `Converter` class to `sdf::Errors`. It also includes minimal changes to the parser so the `Converter` changes can compile and work adequately. A full update to add `sdf::Errors` support to the parser will follow up on a separate PR so this one is not to big.

After this PR is merged the parser might introduce more errors into the `sdf::Errors` structure than before. In order to reduce the time this behavior is changed it is recommended to **wait until the PR that addresses errors on the parser is ready** so it can be merged immediately after this one.

There is one warning that is still printed unless the warnings policy is set to `sdf::EnforcementPolicy::ERR`.

## Test it

Using any method of the `Converter` class should only print warnings and report all errors through `sdf::Errors`. If  the warnings policy is set to `sdf::EnforcementPolicy::ERR` warnings should be reported also in the `sdf::Errors` structure.

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/parser.hh
+++ b/include/sdf/parser.hh
@@ -337,16 +337,15 @@ namespace sdf
                    const ParserConfig &_config, SDFPtr _sdf);
 
   /// \brief Convert an SDF file to a specific SDF version.
+  /// \param[out] _sdf Pointer to the converted SDF document.
   /// \param[in] _filename Name of the SDF file to convert.
   /// \param[in] _version Version to convert _filename to.
   /// \param[in] _config Custom parser configuration
-  /// \param[out] _sdf Pointer to the converted SDF document.
-  /// \param[out] _errors Vector of errors.
-  /// \return True on success.
+  /// \return Vector of errors, successful when vector is empty.
   SDFORMAT_VISIBLE
-  bool convertFile(const std::string &_filename, const std::string &_version,
-                   const ParserConfig &_config, SDFPtr _sdf,
-                   sdf::Errors &_errors);
+  sdf::Errors convertFile(
+      SDFPtr _sdf, const std::string &_filename, const std::string &_version,
+      const ParserConfig &_config = ParserConfig::GlobalConfig());
 
   /// \brief Convert an SDF string to a specific SDF version.
   /// \param[in] _sdfString The SDF string to convert.
@@ -368,16 +367,15 @@ namespace sdf
                      const ParserConfig &_config, SDFPtr _sdf);
 
   /// \brief Convert an SDF string to a specific SDF version.
+  /// \param[out] _sdf Pointer to the converted SDF document.
   /// \param[in] _sdfString The SDF string to convert.
   /// \param[in] _version Version to convert _filename to.
   /// \param[in] _config Custom parser configuration
-  /// \param[out] _sdf Pointer to the converted SDF document.
-  /// \param[out] _errors Vector of errors.
-  /// \return True on success.
+  /// \return Vector of errors, successful when vector is empty.
   SDFORMAT_VISIBLE
-  bool convertString(const std::string &_sdfString, const std::string &_version,
-                     const ParserConfig &_config, SDFPtr _sdf,
-                     sdf::Errors &_errors);
+  sdf::Errors convertString(
+      SDFPtr _sdf, const std::string &_sdfString, const std::string &_version,
+      const ParserConfig &_config = ParserConfig::GlobalConfig());
 
   /// \brief Check that for each model, the canonical_link attribute value
   /// matches the name of a link in the model if the attribute is set and
--- a/include/sdf/parser.hh
+++ b/include/sdf/parser.hh
@@ -333,6 +333,7 @@ namespace sdf
   SDFORMAT_VISIBLE
   bool convertFile(const std::string &_filename, const std::string &_version,
                    const ParserConfig &_config, SDFPtr _sdf);
+
   /// \brief Convert an SDF file to a specific SDF version.
   /// \param[in] _filename Name of the SDF file to convert.
   /// \param[in] _version Version to convert _filename to.
--- a/include/sdf/parser.hh
+++ b/include/sdf/parser.hh
@@ -38,6 +38,8 @@ namespace sdf
   // Inline bracket to help doxygen filtering.
   inline namespace SDF_VERSION_NAMESPACE {
   //
+  // TODO(marcoag): Deprecate function overloads that do not use sdf::Errors,
+  // see: https://github.com/gazebosim/sdformat/issues/1186
   class Root;
 
   /// \brief Initialize the SDF interface from the embedded root spec file
--- a/include/sdf/parser.hh
+++ b/include/sdf/parser.hh
@@ -333,6 +333,17 @@ namespace sdf
   SDFORMAT_VISIBLE
   bool convertFile(const std::string &_filename, const std::string &_version,
                    const ParserConfig &_config, SDFPtr _sdf);
+  /// \brief Convert an SDF file to a specific SDF version.
+  /// \param[in] _filename Name of the SDF file to convert.
+  /// \param[in] _version Version to convert _filename to.
+  /// \param[in] _config Custom parser configuration
+  /// \param[out] _sdf Pointer to the converted SDF document.
+  /// \param[out] _errors Vector of errors.
+  /// \return True on success.
+  SDFORMAT_VISIBLE
+  bool convertFile(const std::string &_filename, const std::string &_version,
+                   const ParserConfig &_config, SDFPtr _sdf,
+                   sdf::Errors &_errors);
 
   /// \brief Convert an SDF string to a specific SDF version.
   /// \param[in] _sdfString The SDF string to convert.
@@ -352,6 +363,18 @@ namespace sdf
   SDFORMAT_VISIBLE
   bool convertString(const std::string &_sdfString, const std::string &_version,
                      const ParserConfig &_config, SDFPtr _sdf);
+
+  /// \brief Convert an SDF string to a specific SDF version.
+  /// \param[in] _sdfString The SDF string to convert.
+  /// \param[in] _version Version to convert _filename to.
+  /// \param[in] _config Custom parser configuration
+  /// \param[out] _sdf Pointer to the converted SDF document.
+  /// \param[out] _errors Vector of errors.
+  /// \return True on success.
+  SDFORMAT_VISIBLE
+  bool convertString(const std::string &_sdfString, const std::string &_version,
+                     const ParserConfig &_config, SDFPtr _sdf,
+                     sdf::Errors &_errors);
 
   /// \brief Check that for each model, the canonical_link attribute value
   /// matches the name of a link in the model if the attribute is set and
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -57,7 +57,8 @@ if (BUILD_TESTING)
     target_sources(UNIT_Converter_TEST PRIVATE
       Converter.cc
       EmbeddedSdf.cc
-      XmlUtils.cc)
+      XmlUtils.cc
+      Utils.cc)
   endif()
 
   if (TARGET UNIT_FrameSemantics_TEST)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -57,8 +57,8 @@ if (BUILD_TESTING)
     target_sources(UNIT_Converter_TEST PRIVATE
       Converter.cc
       EmbeddedSdf.cc
-      XmlUtils.cc
-      Utils.cc)
+      Utils.cc
+      XmlUtils.cc)
   endif()
 
   if (TARGET UNIT_FrameSemantics_TEST)
--- a/src/Converter.cc
+++ b/src/Converter.cc
@@ -84,9 +84,9 @@ void UpdatePose(tinyxml2::XMLElement *_elem,
 }
 
 /////////////////////////////////////////////////
-bool Converter::Convert(tinyxml2::XMLDocument *_doc,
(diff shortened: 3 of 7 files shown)


src/Converter.hh
```diff
@@ -36,10 +37,14 @@ namespace sdf
   {
     /// \brief Convert SDF to the specified version.
     /// \param[in] _doc SDF xml doc
-    /// \param[in] _toVersion Version number in string format
-    /// \param[in] _quiet False to be more verbose
+    /// \param[in] _toVersion Version number in string format.
+    /// \param[out] _errors Vector of errors.
```
scpeters, line 41: it feels a little weird to put an output variable in the middle of the argument list. I think I favor either making output variables the first argument (like [SemanticPose::Resolve](https://github.com/gazebosim/sdformat/blob/sdformat12_12.5.0/include/sdf/SemanticPose.hh#L73-L74)) so that subsequent input parameters could be optional.

Also, we sometimes pass `Errors` by reference and sometimes return from the function. It feels weird to both return `bool` and write to an `Errors` vector though.

@azeey what do you think?

marcoag: > it feels a little weird to put an output variable in the middle of the argument list. I think I favor either making output variables the first argument (like [SemanticPose::Resolve](https://github.com/gazebosim/sdformat/blob/sdformat12_12.5.0/include/sdf/SemanticPose.hh#L73-L74)) so that subsequent input parameters could be optional.

Sounds good to me, will apply the changes on this PR.

> Also, we sometimes pass `Errors` by reference and sometimes return from the function. It feels weird to both return `bool` and write to an `Errors` vector though.

This seems a bit more tricky since not all the time an error is added to the structure `false` is returned, meaning `errors.empty() != return false`, so I'm not sure if there would be trivial way of checking this without changing behavior.

src/CMakeLists.txt
```diff
@@ -57,7 +57,8 @@ if (BUILD_TESTING)
     target_sources(UNIT_Converter_TEST PRIVATE
       Converter.cc
       EmbeddedSdf.cc
-      XmlUtils.cc)
+      XmlUtils.cc
+      Utils.cc)
```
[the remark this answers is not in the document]
marcoag: Done in https://github.com/gazebosim/sdformat/pull/1123/commits/56126196e11f3aec82004e3d39036c339e3877c8.

include/sdf/parser.hh
```diff
@@ -333,6 +333,17 @@ namespace sdf
   SDFORMAT_VISIBLE
   bool convertFile(const std::string &_filename, const std::string &_version,
                    const ParserConfig &_config, SDFPtr _sdf);
+  /// \brief Convert an SDF file to a specific SDF version.
```
scpeters, line 336: nit: add blank line above

include/sdf/parser.hh
```diff
@@ -333,6 +333,17 @@ namespace sdf
...
+  /// \param[in] _filename Name of the SDF file to convert.
+  /// \param[in] _version Version to convert _filename to.
+  /// \param[in] _config Custom parser configuration
+  /// \param[out] _sdf Pointer to the converted SDF document.
+  /// \param[out] _errors Vector of errors.
+  /// \return True on success.
+  SDFORMAT_VISIBLE
+  bool convertFile(const std::string &_filename, const std::string &_version,
+                   const ParserConfig &_config, SDFPtr _sdf,
+                   sdf::Errors &_errors);
```
scpeters, line 346: hot take:

for the new APIs, let's return `sdf::Errors` and pass the `SDFPtr` as the first argument and make ParserConfig optional

~~~
sdf::Errors convertFile(
    SDFPtr _sdf,
    const std::string &_filename,
    const std::string &_version,
    const ParserConfig &_config = ParserConfig::GlobalConfig())
~~~

include/sdf/parser.hh
```diff
@@ -353,6 +364,18 @@ namespace sdf
...
+  /// \param[in] _sdfString The SDF string to convert.
+  /// \param[in] _version Version to convert _filename to.
+  /// \param[in] _config Custom parser configuration
+  /// \param[out] _sdf Pointer to the converted SDF document.
+  /// \param[out] _errors Vector of errors.
+  /// \return True on success.
+  SDFORMAT_VISIBLE
+  bool convertString(const std::string &_sdfString, const std::string &_version,
+                     const ParserConfig &_config, SDFPtr _sdf,
+                     sdf::Errors &_errors);
```
scpeters, line 377: I suggest:

~~~
sdf::Errors convertString(
    SDFPtr _sdf,
    const std::string &_sdfString,
    const std::string &_version,
    const ParserConfig &_config = ParserConfig::GlobalConfig())
~~~

scpeters (commented):
I made some suggestions on how the new APIs in parser.hh should be structured. The Converter.hh APIs are not public, so they are not as critical

include/sdf/parser.hh
marcoag: Wouldn't this leave it a bit weird in the sense that some of the overload functions return an `sdf::Errors` structure while others a `bool`? Also to add to the argument, the most common way of obtaining `sdf::Errors` in the library is through a parameter.

scpeters: all the `Load(sdf::ElementPtr)` functions return `Errors` as well as the `Resolve*` functions in the DOM classes. If a method already returns something like a `SDFPtr`, I think we need to pass the `Errors` by reference, but for APIs that would otherwise return `bool`, I think it makes more sense to return `Errors`

@azeey your thoughts?

marcoag: I had a deeper look at this suggestion and I found one more drawback: 

Because the old functions are now calling the new ones (i.e.: [here](https://github.com/gazebosim/sdformat/blob/ecb8fa225c1638095a28c1774d4598078f85d590/src/parser.cc#L1944)) it makes it very hard to figure out when to return `False` or `True` in these old functions if the new function is only returning `sdf::Errors`. This is due to the fact that some of the code down the line might add and `sdf::Error` and not return anything at that point (i.e.: [here](https://github.com/gazebosim/sdformat/blob/ecb8fa225c1638095a28c1774d4598078f85d590/src/Converter.cc#L171)). So I'm not sure if this change can be applied without breaking behavior or having some code duplication on the old functions...

azeey: Since the `Errors` object is going to be passed to multiple functions, I think it's better to pass it as an output parameter. Otherwise, a function will need to aggregate `Errors` from all the functions it calls, which would add a lot of overhead.

include/sdf/parser.hh
```diff
@@ -333,6 +333,17 @@ namespace sdf
...
                    const ParserConfig &_config, SDFPtr _sdf);
+  /// \brief Convert an SDF file to a specific SDF version.
+  /// \param[in] _filename Name of the SDF file to convert.
+  /// \param[in] _version Version to convert _filename to.
+  /// \param[in] _config Custom parser configuration
+  /// \param[out] _sdf Pointer to the converted SDF document.
+  /// \param[out] _errors Vector of errors.
+  /// \return True on success.
+  SDFORMAT_VISIBLE
+  bool convertFile(const std::string &_filename, const std::string &_version,
```
azeey, line 344: Can you add a TODO to deprecate the other overloads?

azeey (approved):

include/sdf/parser.hh
scpeters: > Since the `Errors` object is going to be passed to multiple functions, I think it's better to pass it as an output parameter.

I think it depends on the context. There are methods in our API that pass Errors along but don't check them. If you don't care about the Errors, then it is simpler to pass the Errors object by reference. If you do care about the result, however, then it's a little bit more work because you have to create a clean `Errors` object to pass in and then possibly insert those errors back into the vector of accumulated errors.

azeey: In the case where we do care about the Errors object, inserting the errors back into the vector of accumulated errors would happen regardless of whether we passed by reference or returned an `Errors` object. The only difference would be whether we have an extra line creating a new `Errors` object or not.

I think returning `Errors` is more convenient for APIs that are main entry points, such as `Root::Load`. You could argue `convertFile` is also a main entry point, but as @marcoag said, determining the boolean value of the returned `Errors` object is not straightforward for the functions that are being overloaded.

scpeters: my proposed API puts the `SDFPtr` as the first argument, which is different enough from the existing API that it should hopefully be obvious which overload is being used. Putting the output variables first allows optional `const` arguments at the end so we could reduce the number of overloads

include/sdf/parser.hh
marcoag: You mean to deprecate all overloads that we are creating for the error reporting right? where would you say we should put this TODO?

src/Converter.cc
```diff
@@ -237,47 +285,55 @@ void Converter::ConvertImpl(tinyxml2::XMLElement *_elem,
...
+      Unflatten(_elem, _errors);
     }
     else if (name != "convert")
     {
-      sdferr << "Unknown convert element[" << name << "]\n";
+      std::stringstream ss;
+      ss << "Unknown convert element["
+         << name
+         << "]";
+      _errors.push_back({ErrorCode::FATAL_ERROR, ss.str()});
```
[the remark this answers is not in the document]
marcoag: Definitely not, switched to `CONVERSION_ERROR` in https://github.com/gazebosim/sdformat/pull/1123/commits/fc741b23a57e481961db9352af92a41d8fe9d9f5.

include/sdf/parser.hh
marcoag: Added in https://github.com/gazebosim/sdformat/pull/1123/commits/fc741b23a57e481961db9352af92a41d8fe9d9f5.

src/Converter.hh
marcoag: Just realized, shall we make the `ParserConfig` also optional?

include/sdf/parser.hh
azeey: Good question. We should definitely create a ticket for it, but I'm not sure we'll remember to check, so I suggest putting a TODO in at least one of these files (probably parser.hh since it's more frequently updated/looked at)

include/sdf/parser.hh
azeey: The problem is not knowing which overload is being used. It's that the old function (one that doesn't take `sdf::Error`) is now implemented in terms of the new function. But the old function returns a bool so if we make the new function return an `sdf::Error`, how would we interpret that to a bool.

src/Converter.hh
azeey: Since this is an internal function, I recommend against making `ParserConfig` optional. Whoever is calling should be passing a config.

include/sdf/parser.hh
scpeters: `sdf::Errors` is a vector, so return `true` if it is empty: `return errors.empty();`

marcoag: > some of the code down the line might add and sdf::Error and not return anything at that point (i.e.: [here](https://github.com/gazebosim/sdformat/blob/ecb8fa225c1638095a28c1774d4598078f85d590/src/Converter.cc#L171)).

If we return `false` with that error the behavior of that function and any other users of it will change.

scpeters: > some of the code down the line might add and sdf::Error and not return anything at that point (i.e.: [here](https://github.com/gazebosim/sdformat/blob/ecb8fa225c1638095a28c1774d4598078f85d590/src/Converter.cc#L171)).
> 
> If we return `false` with that error the behavior of that function and any other users of it will change.
> 
> In short: in the current behavior, not always there is an error in the `sdf::Errors` false is returned.

Your example is a different function `Converter::Convert`. I'm only proposing this change for the `convertFile` and `convertString` functions. I'm not sure how I understand how your comment applies to `convertFile` and `convertString`

include/sdf/parser.hh
marcoag: Added an issue [here](https://github.com/gazebosim/sdformat/issues/1186) and the TODO in https://github.com/gazebosim/sdformat/pull/1123/commits/5f91d45a201662332d045ab845cda5ddea827e96.

include/sdf/parser.hh
marcoag: But `convertFile` calls `Converter::Convert` [here](https://github.com/gazebosim/sdformat/blob/5f91d45a201662332d045ab845cda5ddea827e96/src/parser.cc#L1992), so there's a chance that the `sdf::Errors` being passed there gets an error and no `false` is returned, isn't it?

scpeters: all the errors will be returned, so I don't think that would be a problem

here is what I would do for `convertFile`: dcbedd7654cccb055605578e7b7a10aad2c0b974

scpeters: > But `convertFile` calls `Converter::Convert` [here](https://github.com/gazebosim/sdformat/blob/5f91d45a201662332d045ab845cda5ddea827e96/src/parser.cc#L1992), so there's a chance that the `sdf::Errors` being passed there gets an error and no `false` is returned, isn't it?

an `sdf::Errors` vector is passed to `Converter::Convert`, but I believe `Convert` only returns `false` if it adds an error to the vector, so I think it will be fine with the changes I have proposed in #1188

include/sdf/parser.hh
scpeters: see #1188

include/sdf/parser.hh
marcoag: Merged proposal. Thanks!

PR #1147, opened 2022-09-16
Add a configuration option to resolve URIs

mjcarroll:
# 🎉 New feature

## Summary

This adds an option for the SDF parser to attempt to resolve any URIs in the document via a user-provided callback.
This works similarly to how `include/uri` is resolved, but works for all `uri` elements.
A `ParserConfig` parameter is added to opt in to this behavior.
In the case that the user-provided callback cannot resolve the URI (returns an empty string), the original contents of the `uri` tag will be preserved.

This is needed in support of: https://github.com/gazebosim/gz-sim/pull/1560

## Checklist
- [ ] Added tests


include/sdf/ParserConfig.hh
```diff
@@ -174,6 +174,12 @@ class SDFORMAT_VISIBLE ParserConfig
   /// \brief Get the preserveFixedJoint flag value.
   public: bool URDFPreserveFixedJoint() const;
 
+  /// \brief Set the storeResolvedURIs flag value.
```
ahcorde, line 177:
```suggestion
  /// \brief Set the storeResolvedURIs flag value.
  \param[in] _resolveURI
```

include/sdf/ParserConfig.hh
```diff
@@ -174,6 +174,12 @@ class SDFORMAT_VISIBLE ParserConfig
...
+  public: void SetStoreResovledURIs(bool _resolveURI);
+
+  /// \brief Get the storeResolvedURIs flag value.
```
ahcorde, line 180:
```suggestion
  /// \brief Get the storeResolvedURIs flag value.
  /// \return 
```

src/parser.cc
```diff
@@ -1570,6 +1570,25 @@ bool readXml(tinyxml2::XMLElement *_xml, ElementPtr _sdf,
...
+        if (resolvedUri.empty())
+        {
+          Error err(ErrorCode::URI_LOOKUP, "Unable to resolve uri[" + uri + "]",
+              _source, elemXml->GetLineNum());
+          err.SetXmlPath(_sdf->XmlPath() + "/uri");
+          _errors.push_back(err);
+        }
+        else
+        {
+          elemXml->SetText(resolvedUri.c_str());
```
azeey, line 1588: This means we'll lose the original URI, which is not ideal for certain use cases, e.g. saving a world to sdf. A few alternatives to consider:
1. Do the file resolution in `sdf::Mesh`. This requires passing the `ParserConfig` down to `sdf::Mesh` from `sdf::Model`. Tedious, but doable, and wouldn't break ABI.
2. Add a member variable to `sdf::ElementPrivate` to store the resolved URI and expose this variable through `sdf::Mesh`,  Unfortunately, this would break ABI since `ElementPrivate` is used by templates such as `sdf::Element::Get`.
3. (Hacky) Add the resolved URI as a custom attribute to the `<uri>` element, e.g. `gz:resolved_uri="path/to/file"`. Then read this attribute in `sdf::Mesh`. Note that this will likely be included in the generated SDF when saving worlds, but that won't break anything since it's a custom attribute.

mjcarroll: I ended up going with approach 1.  In addition to mesh, skybox and heightmap also make use of URI elements, so those were resolved as well.

mjcarroll:

src/Utils.hh
```diff
@@ -204,6 +204,58 @@ namespace sdf
...
+  /// \param[in] _config Specific parser configuration to use to load
+  /// \param[out] _objs Elements that match _sdfName in _sdf are added to this
+  /// vector, unless an error is encountered during load.
+  /// \return The vector of errors. An empty vector indicates no errors were
+  /// experienced.
+  template <typename Class>
+  sdf::Errors loadRepeated(sdf::ElementPtr _sdf, const std::string &_sdfName,
+      const ParserConfig &_config,
+      std::vector<Class> &_objs,
+      const std::function<void(Class &)> &_beforeLoadFunc = {})
```
scpeters, line 220: I looked and I can't find a single place where this `std::function` argument is used. `git blame` indicates that it was added in #381, which was a big PR, and I'm guessing we didn't notice that the argument wasn't used by the time it was ready to be merged.

Given that `_beforeLoadFunc` isn't used, I would suggest removing the parameter and the small code block where it is used, and adding `const ParserConfig &_config = ParserConfig::GlobalConfig()` in its place so we don't need to duplicate this method.

scpeters: ok, but now I see that the `Load(obj, _config)` call is different, so we need two different versions of this template function. never mind I guess

mjcarroll: Yeah, I took a moment to try and do some SFINAE template trickery to dispatch to the correct signature, but it ended up getting ugly quickly for such a simple function and I opted to just duplicate.

If all of the DOM objects eventually have a Load(obj, config), then we could remove the original method, but that would increase the scope of this pr substantially (for not much benefit at the moment)

scpeters: 👍

scpeters: hold on, I believe [loadUniqueRepeated](https://github.com/gazebosim/sdformat/blob/sdf13/src/Utils.hh#L104-L105) is able to handle this with `Args&&... _args`, since [World::Load passes a ParserConfig to Model::Load](https://github.com/gazebosim/sdformat/blob/sdf13/src/World.cc#L225)

that would require removing the `beforeLoadFunc` argument, but it's not used anyway

mjcarroll: Good call, I switched to the variadic implementation in [1746843](https://github.com/gazebosim/sdformat/pull/1147/commits/17468435d3b288f01eb20d22e27842b9b15c46f8)

mjcarroll:
@scpeters one more green here?

scpeters:
> @scpeters one more green here?

yes I'll take a look

src/Heightmap.cc
```diff
@@ -311,7 +353,23 @@ Errors Heightmap::Load(ElementPtr _sdf)
...
+      {
+        errors.push_back({ErrorCode::URI_LOOKUP,
+            "Parser configurations requested resolved uris, but uri ["
+            + uri + "] could not be resolved."});
+      }
+      else
+      {
+        uri = resolvedUri;
+      }
+    }
```
scpeters, line 371: consider a helper function for this code block that is repeated 3 times in this file

scpeters (changes requested):

src/Heightmap.cc
mjcarroll: Added a helper to utils, and used it accordingly.

test/integration/resolve_uris.cc
```diff
@@ -0,0 +1,420 @@
...
+    texture->Element()->GetElement("normal")->GetValue()->GetAsString();
+
+  auto heightmapCol = shapesLink->CollisionByName("heightmap_col");
+  ASSERT_NE(nullptr, heightmapCol);
+  auto heightmapColUri = heightmapCol->Geom()->HeightmapShape()->Uri();
+  auto heightmapColElem = heightmapCol->Geom()->HeightmapShape()->Element();
+  auto heightmapColElemUri =
+    heightmapColElem->GetElement("uri")->GetValue()->GetAsString();
+
+  /// DOM URIs are modified
```
scpeters, line 401: is this comment correct? it looks like the URI's are not modified in this case

mjcarroll: Good catch. Fixed.

scpeters (approved):
thanks for your patience!

Issue #199, opened 2018-09-05
URDF to SDF conversion ignores links without inertia


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/199)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
For example, this:

~~~
<?xml version="1.0" ?>
<robot name="always_ignored">
  <link name="link">
    <visual>
      <geometry>
        <sphere radius="1.0"/>
      </geometry>
    </visual>
  </link>
</robot>
~~~

Results in:

~~~
$ gz sdf -p /tmp/urdf_ball.urdf 
<sdf version='1.6'>
  <model name='always_ignored'/>
</sdf>
~~~

But when inertia is added, the link is not ignored:

~~~
<?xml version="1.0" ?>
<robot name="always_ignored">
  <link name="link">
    <visual>
      <geometry>
        <sphere radius="1.0"/>
      </geometry>
    </visual>
    <inertial>
      <mass value="1"/>
      <inertia ixx="1" ixy="0.0" ixz="0.0" iyy="1" iyz="0.0" izz="1"/>
    </inertial>
  </link>
</robot>
~~~

~~~
$ gz sdf -p /tmp/urdf_ball.urdf 
<sdf version='1.6'>
  <model name='always_ignored'>
    <link name='link'>
      <pose frame=''>0 0 0 0 -0 0</pose>
      <inertial>
        <pose frame=''>0 0 0 0 -0 0</pose>
        <mass>1</mass>
        <inertia>
          <ixx>1</ixx>
          <ixy>0</ixy>
          <ixz>0</ixz>
          <iyy>1</iyy>
          <iyz>0</iyz>
          <izz>1</izz>
        </inertia>
      </inertial>
      <visual name='link_visual'>
        <pose frame=''>0 0 0 0 -0 0</pose>
        <geometry>
          <sphere>
            <radius>1</radius>
          </sphere>
        </geometry>
      </visual>
    </link>
  </model>
</sdf>
~~~

---

It could be argued that:

* it is good practice to add inertia for simulation
* URDF will rarely be used to describe static objects, where inertia is ignored

[...]

PR #1171, opened 2022-09-26
Offer an API option to display new warnings on problems with URDF to SDF conversions

j-rivero:
# 🎉 New feature

Should help providing an API option to have explicit warnings for https://github.com/gazebosim/sdformat/issues/199 and https://github.com/gazebosim/sdformat/issues/1007.

## Summary
The PR tries to address the lack of the explicit information on decisions made that end up ignoring some link definitions in the user input which can lead to many problems see https://github.com/gazebosim/sdformat/issues/199 and https://github.com/gazebosim/sdformat/issues/1007.

For keeping backwards compatibility, no new warning is raised unless the consumer of sdformat readFile 

The PR implements a new API function:
```c++
bool readFile(const std::string &_filename, SDFPtr _sdf, bool _enable_new_warnings);
```
that allows users of sdf to enable new extra warnings.

If the parameter is enabled, the problems with links being ignored silently (not really, but the issues were stored in the logs) are promoted to warnings using `gzwarn` in the `CreateSDF` function.

## Test it
Should be fairly trivial to change Gazebo or other consumer of the `readFile` API to use the new parameter. Examples of code that triggers problems can be found in https://github.com/gazebosim/sdformat/issues/199 and https://github.com/gazebosim/sdformat/issues/1007. Simple patch for gazebo9:
```diff
diff --git a/tools/gz.cc b/tools/gz.cc
index 7a8f3d6b16..116c94541f 100644
--- a/tools/gz.cc
+++ b/tools/gz.cc
@@ -1070,6 +1070,10 @@ bool SDFCommand::RunImpl()
     }
   }
 
+  bool new_warnings = true;
+  if (common::getEnv("GAZEBO11_BACKWARDS_COMPAT_WARNINGS_ERRORS"))
+    new_warnings = false;
+
   if (!sdf::init(sdf))
   {
     std::cerr << "ERROR: SDF parsing the xml failed" << std::endl;
@@ -1124,7 +1128,7 @@ bool SDFCommand::RunImpl()
     if (!boost::filesystem::exists(path))
       std::cerr << "Error: File doesn't exist[" << path.string() << "]\n";
 
-    if (!sdf::readFile(path.string(), sdf))
+    if (!sdf::readFile(path.string(), sdf, new_warnings))
     {
       std::cerr << "Error: SDF parsing the xml failed\n";
       return false;
```

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/parser.hh
+++ b/include/sdf/parser.hh
@@ -115,6 +115,21 @@ namespace sdf
   SDFORMAT_VISIBLE
   bool readFile(const std::string &_filename, SDFPtr _sdf);
 
+  /// \brief Populate the SDF values from a file
+  ///
+  /// This populates the given sdf pointer from a file. If the file is a URDF
+  /// file it is converted to SDF first. All files are converted to the latest
+  /// SDF version
+  /// \param[in] _filename Name of the SDF file
+  /// \param[in] _sdf Pointer to an SDF object.
+  /// \param[in] _enable_new_warnings Consider previous issues in debug log as
+  ///            warnings. This allow to keep log level backwards compatible
+  ///            while providing an option to increase the warnings with new
+  ///            relevant issues.
+  /// \return True if successful.
+  bool readFile(const std::string &_filename, SDFPtr _sdf,
+      bool _enable_new_warnings);
+
   /// \brief Populate the SDF values from a string
   ///
   /// This populates the sdf pointer from a string. If the string is a URDF
--- a/include/sdf/parser_urdf.hh
+++ b/include/sdf/parser_urdf.hh
@@ -42,6 +42,15 @@ namespace sdf
     /// \brief constructor
     public: URDF2SDF();
 
+    /// \brief constructor
+    /// \param[in] _enable_new_warnings Consider previous issues in debug log as
+    ///            warnings.
+    public: explicit URDF2SDF(const bool _enable_new_warnings) :
+            URDF2SDF()
+      {
+        enable_new_warnings = _enable_new_warnings;
+      };
+
     /// \brief destructor
     public: ~URDF2SDF();
 
@@ -81,6 +90,9 @@ namespace sdf
     // cppcheck-suppress unusedPrivateFunction
     // cppcheck-suppress unmatchedSuppression
     private: void ListSDFExtensions(const std::string &_reference);
+
+    /// when to raise new warnings from previous log issues
+    private: bool enable_new_warnings;
   };
   }
 }
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -54,12 +54,15 @@ inline namespace SDF_VERSION_NAMESPACE {
 /// \param[in] _sdf Pointer to an SDF object.
 /// \param[in] _convert Convert to the latest version if true.
 /// \param[out] _errors Parsing errors will be appended to this variable.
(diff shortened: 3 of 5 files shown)


include/sdf/parser.hh
```diff
@@ -115,6 +115,21 @@ namespace sdf
...
+  /// file it is converted to SDF first. All files are converted to the latest
+  /// SDF version
+  /// \param[in] _filename Name of the SDF file
+  /// \param[in] _sdf Pointer to an SDF object.
+  /// \param[in] _enable_new_warnings Consider previous issues in debug log as
+  ///            warnings. This allow to keep log level backwards compatible
+  ///            while providing an option to increase the warnings with new
+  ///            relevant issues.
+  /// \return True if successful.
+  bool readFile(const std::string &_filename, SDFPtr _sdf,
```
azeey, line 130: While I think this is super helpful, I'm not sure adding a new overload is the best option. In Citadel, we only have two existing overloads, so this adds a third. In Garden, we have 6 overloads already! and how this will interact with all of them when forward ported is not clear. Starting from Fortress, these functions take a `ParserConfig` object, so adding it to that makes sense, but for citadel, I would prefer if we could use a different approach. Perhaps an env variable? Or backport` ParserConfig`?

scpeters: before handling `sdf9`, I'd like to review a pull request that targets `sdf12` and uses `ParserConfig`. Once we have completed that review, we can decide how we want to backport it to `sdf9`. Does that sound ok?

j-rivero: sure, I'll try to create it in the next days.

aaronchongth: Hello @azeey, @j-rivero, @scpeters! Circling back to this discussion, I've implemented the ideas for `sdf12`, https://github.com/gazebosim/sdformat/pull/1238,
* changing `sdfdbg` to `sdfwarn` whenever links are ignored during urdf->sdf conversion
* adding a setting in `ParserConfig` to convert urdf links with no inertia or masses smaller than 1e-6 to frames

I found that using warnings directly do not trigger any regressions, users will however start seeing them in their workflows without any change in behavior (might be a good thing, since these links are most likely mistakes)

Let me know what yall think!

azeey: I think converting the debug messages to warnings (without any config flags) is reasonable since my expectation is that users generally do not rely on this behavior being the normal, i.e, we want to avoid spamming users who expect links to be ignored and they're okay with it, but I don't think that's most users.

aaronchongth: We just got https://github.com/gazebosim/sdformat/issues/199 merged into `sdf12`, changing debug messages to warnings, with more explicit information if links or joints were ignored, and suggestions to resolve those issues. There's no behavior change (but with 1 bug fix), and `ParserConfig` was not used. If it looks ok, we can probably try to cherry pick that back to `sdf9` for gazebo11 and citadel?

aaronchongth: @j-rivero @azeey https://github.com/gazebosim/sdformat/pull/1258 is the cherry-pick towards `sdf9`, it looks like the new warnings are not causing any regressions. Do yall think we still need an option to display newer warnings?

azeey:
Closing since #1238 and #1258 have been merged.

PR #1188, opened 2022-10-14
Different API for convertFile and convertString

scpeters:
Targeting #1123: proposes an alternative API for `sdf::convertFile` and `sdf::convertString`.

### DIFF
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -1944,17 +1944,17 @@ bool convertFile(const std::string &_filename, const std::string &_version,
 bool convertFile(const std::string &_filename, const std::string &_version,
                  const ParserConfig &_config, SDFPtr _sdf)
 {
-  sdf::Errors errors;
-  bool result = convertFile(_filename, _version, _config, _sdf, errors);
+  sdf::Errors errors = convertFile(_sdf, _filename, _version, _config);
   throwOrPrintErrors(errors);
-  return result;
+  return errors.empty();
 }
 
 /////////////////////////////////////////////////
-bool convertFile(const std::string &_filename, const std::string &_version,
-                 const ParserConfig &_config, SDFPtr _sdf,
-                 sdf::Errors &_errors)
+sdf::Errors convertFile(SDFPtr _sdf, const std::string &_filename,
+                        const std::string &_version,
+                        const ParserConfig &_config)
 {
+  sdf::Errors errors;
   std::string filename = sdf::findFile(_filename, true, false, _config);
 
   if (filename.empty())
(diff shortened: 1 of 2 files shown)


src/parser.cc
```diff
@@ -1944,17 +1944,17 @@ bool convertFile(const std::string &_filename, const std::string &_version,
...
   throwOrPrintErrors(errors);
-  return result;
+  return errors.empty();
 }
 
 /////////////////////////////////////////////////
-bool convertFile(const std::string &_filename, const std::string &_version,
-                 const ParserConfig &_config, SDFPtr _sdf,
-                 sdf::Errors &_errors)
+sdf::Errors convertFile(SDFPtr _sdf, const std::string &_filename,
```
marcoag, line 1953: @scpeters Wouldn't this break API?

scpeters: this is targeting your branch, which hasn't been merged to `sdf13` yet. It changes the proposed API. Comparing the branch to `sdf13`, nothing is broken: https://github.com/gazebosim/sdformat/compare/sdf13...scpeters/sdf_error_converter

marcoag: True, my bad.

src/parser.cc
```diff
@@ -1944,17 +1944,17 @@ bool convertFile(const std::string &_filename, const std::string &_version,
 bool convertFile(const std::string &_filename, const std::string &_version,
                  const ParserConfig &_config, SDFPtr _sdf)
 {
-  sdf::Errors errors;
-  bool result = convertFile(_filename, _version, _config, _sdf, errors);
+  sdf::Errors errors = convertFile(_sdf, _filename, _version, _config);
   throwOrPrintErrors(errors);
-  return result;
+  return errors.empty();
```
marcoag, line 1949: So I believe here is the origin of the discussion [here](https://github.com/gazebosim/sdformat/pull/1123#discussion_r995358960). There is a change of behavior here:

- Now: any `sdf::Error` will make this function return `false`
- Before: There can be some `sdf::Errors` and still return `true`

Example of a case for the later: 
`convertFile` -> `Convert` -> `ConvertImpl`: 

1.  `ConvertImpl` returns an `sdf::Error` but no boolean:
https://github.com/gazebosim/sdformat/blob/84e83b1d4e82989bc403c2946f5cce6cd721b5ad/src/Converter.cc#L171
2. `Convert` can now return extra `sdf::Errors` from `ConvertImpl`:
https://github.com/gazebosim/sdformat/blob/84e83b1d4e82989bc403c2946f5cce6cd721b5ad/src/Converter.cc#L186
3. Finally when `convertFile` sued to return `True` will now return `False` due to the `sdf::Error` check:
https://github.com/gazebosim/sdformat/blob/702dc371e2881c8b1755823f79682f1e32df6113/src/parser.cc#L1987

isn't it?

While I understand your proposal seems much better I'm worried it might change the cases this function is returning `true` or `false` and I believe we don't want that to happen.

azeey: It's true there are number of places where functions called from `ConvertImpl` would add to the `errors` vector without causing the resulting return value of `Convert` to be `false`.  But looking at those locations closely, it seems like errors are only added when there's a syntax issue with the `.convert` files. Since these files are internal and users are not allowed to change them, it's unlikely these errors will be generated for our end users. The only potential error that I found not in the category is in `CheckDeprecation` when the error policy is `ERR`.https://github.com/gazebosim/sdformat/blob/84e83b1d4e82989bc403c2946f5cce6cd721b5ad/src/Converter.cc#L1251. Even then the `deprecated` tag is only used in https://github.com/gazebosim/sdformat/blob/84e83b1d4e82989bc403c2946f5cce6cd721b5ad/sdf/1.2/1_0.convert#L122-L123. I don't think we support that version of sdformat, so, unless there are I missed other behavior changes, I think it would be safe to change the API as suggested in this PR.

marcoag: Thanks for having a deep look into it @azeey. Will go ahead and merge then! Thanks for the contribution here @scpeters.

PR #1190, opened 2022-10-19
Fix assumptions that `CMAKE_INSTALL_*DIR` variables are relative

lopsided98:
# 🦟 Bug fix

## Summary

The `CMAKE_INSTALL_*DIR` variables may be absolute, in which case they must not be concatenated with `CMAKE_INSTALL_PREFIX`. If an absolute path is needed, `CMAKE_INSTALL_FULL_*DIR` must be used. In other cases, more complex logic may be required to make packages relocatable in the case that `CMAKE_INSTALL_*DIR` is relative.

This patch replaces the custom `LIB_INSTALL_DIR` and `BIN_INSTALL_DIR` variables with the standard GNUInstallDirs variables throughout the code. The custom variables are still honored if specified manually to maintain compatibility, although perhaps this isn't necessary.

Some more work will be needed to port this to newer versions.

## Checklist
- [ ] Added tests


CMakeLists.txt
```diff
@@ -304,12 +311,22 @@ else (build_errors)
...
-  set(PC_CONFIG_INSTALL_DIR ${LIB_INSTALL_DIR}/pkgconfig)
+  set(PC_CONFIG_INSTALL_DIR ${CMAKE_INSTALL_FULL_LIBDIR}/pkgconfig)
   file(RELATIVE_PATH
     PC_CONFIG_RELATIVE_PATH_TO_PREFIX
-    "${CMAKE_INSTALL_PREFIX}/${PC_CONFIG_INSTALL_DIR}"
+    "${PC_CONFIG_INSTALL_DIR}"
     "${CMAKE_INSTALL_PREFIX}"
   )
+  if(IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
+    set(SDFORMAT_PKG_CONFIG_LIBDIR "${CMAKE_INSTALL_LIBDIR}")
```
scpeters, line 321: this should still be relative to the `${prefix}` variable to ensure relocatability. An additional `file(RELATIVE_PATH` call can be used to compute the proper relative path

lopsided98: This makes sense assuming `CMAKE_INSTALL_*DIR` is a subdirectory of `CMAKE_INSTALL_PREFIX`. In nixpkgs, we often pass something like `-DCMAKE_INSTALL_PREFIX=/nix/store/<hash>-sdformat-9.9.0 -DCMAKE_INSTALL_LIBDIR=/nix/store/<hash>-sdformat-9.9.0/lib -DCMAKE_INSTALL_INCLUDEDIR=/nix/store/<hash>-sdformat-9.9.0-dev/include` in order to install the headers into a separate package.

If we use `file(RELATIVE_PATH <variable> "${CMAKE_INSTALL_PREFIX}" "${CMAKE_INSTALL_INCLUDEDIR}")` in that case, the pkg-config file end up with a line like: `includedir=${prefix}/../../../nix/store/<hash>-sdformat-9.9.0-dev/include`, which works but is a bit weird.

I came up with a simpler alternative solution using `string(REPLACE)`, based on how CMake solves the same problem in `configure_package_config_file()`: https://github.com/Kitware/CMake/blob/bcf203226a6877824b91f2ffa9b5cc0373900273/Modules/CMakePackageConfigHelpers.cmake#L279. The only potential downside to this method I can see is that it would fail if `CMAKE_INSTALL_FULL_*DIR` contained the value of `CMAKE_INSTALL_PREFIX` somewhere other than at the start of the string, but this seems pretty unlikely. This scenario would break `configure_package_config_file()` as well.

scpeters (approved):

PR #1193, opened 2022-11-03
Fix static URDF models with fixed joints 

scpeters:
# 🦟 Bug fix

Allows URDF models with fixed joint reduction to be declared as `<static>`

## Summary

Our URDF parser extensions allow a `<static>1</static>` tag inside a `<gazebo/>` element (with no `reference` attribute) to declare the model as static. Currently, the `SDFExtension` data structure does not indicate whether any value for `<static>` was specified or not, so every `<gazebo>` block will cause a `<static>0</static>` value to be added to the resulting SDFormat model. As of #1148, every URDF fixed joint that is reduced (ie. without `<preserveFixedJoint>1</preserveFixedJoint>`) will add a `<gazebo>` block with `<frame>` elements, which would prevent those models from being marked as static. This is illustrated with the `fixed_joint_static.urdf` example added in https://github.com/gazebosim/sdformat/commit/2dc5c1f9c79119502f838c8353b1a42731abb692 with a failing test.

A fix is included in https://github.com/gazebosim/sdformat/commit/6580927ecd1a096d10b60989b6e9a7a137db27c0?w=1 (without whitespace) by adding a flag to indicate when the `<static>` element has been set in an `SDFExtension`.

## Checklist
- [X] Added tests


adityapande-1995 (approved):

src/SDFExtension.cc
```diff
@@ -74,6 +75,7 @@ SDFExtension::SDFExtension(const SDFExtension &_ge)
   this->material = _ge.material;
   this->visual_blobs = _ge.visual_blobs;
   this->collision_blobs = _ge.collision_blobs;
+  this->isSetStaticFlag = _ge.isSetStaticFlag;
```
adityapande-1995, line 78: Can we do something about this line in codecov ?

scpeters: I just removed the SDFExtension copy constructor and destructor so the class can follow the rule of zero. That should improve coverage.

scpeters: coverage checker is green; removed 45 missed lines

PR #1211, opened 2022-12-07
Warn child joint that resolves to world

liamhan0905:
## Summary

SDFormat doesn't allow 
- [already implemented] joint child to be world
- [new] joint child to resolve to world (e.g. joint child can be a frame that's attached to a world)

The latter is implemented in this PR

I originally created a new test with a example sdf. However, after pushing the code, I realized that there's actually a very similar TEST that checks for the same thing. Therefore the latest push cleans up all the unnecessary noise.

## Test it
~~The first commit only contains the test so it doesn't catch when the resolved child link is the world. Therefore, it'll result in a FAIL test. But the second commit contains the logic to catch when the resolved child link is the world. Hence the test will PASS~~
- ~~checkout `1390e88f550a5a40329b433b1f2324c721b2bb9b` commit and rebuild. Run the `gazebo_garden/build/sdformat13/bin/INTEGRATION_joint_dom` test. This should result in one test that FAILS.~~
- ~~checkout `58319f00e436bd82263de6767b269cb3d3815a17` commit and rebuild. Run the test again and this time, the test will PASS.~~

## Checklist
- [ ] Added tests


liamhan0905:
@azeey now that I think about it, maybe this can be much simpler... 
1. I don't think I need to add another TEST since `worldJointInvalidChildWorld` test does pretty much the same.
2. I won't have to add additional example sdf since it's already pretty much the same as `world_joint_child_resolved_to_world` sdf

azeey (approved):
Looks great! And props on realizing that we don't need another test.

/cc @scpeters

scpeters:
> Looks great! And props on realizing that we don't need another test.
> 
> /cc @scpeters

good call; I guess I had made the test but didn't add all the checking logic. It would have been more obvious if I had split the error cases into separate files, but I was lazy. Nice job catching it!

PR #1213, opened 2022-12-14
Allowing relative paths in URDF

FracturedShader:
# 🦟 Bug fix
## Summary
When parsing a URDF file, the `_source` parameter is set to "urdf file" which breaks relative path resolution later on. This commit changes the call to `sdf::readDoc` to use `filename` as the `_source` parameter instead.

## Additional Information
I came across this issue while testing out a physics plugin I am developing for PhysX 4.1 integration. Searching all repos pulled by colcon reveals no other uses of the phrase "urdf file" as a special case. Testing has revealed no new issues.

### DIFF
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -415,7 +415,7 @@ bool readFileInternal(const std::string &_filename, SDFPtr _sdf,
   {
     URDF2SDF u2g;
     TiXmlDocument doc = u2g.InitModelFile(filename);
-    if (sdf::readDoc(&doc, _sdf, "urdf file", _convert, _errors))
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

azeey:
Thanks for the contribution. Since this is a small change, would you mind setting the target branch to `sdf9` so the bug fix is goies into earlier versions as well?

FracturedShader:
> Thanks for the contribution. Since this is a small change, would you mind setting the target branch to `sdf9` so the bug fix is goies into earlier versions as well?

No problem. Rebased onto `sdf9` and modified the commit to fit.

azeey (approved):

PR #1220, opened 2023-01-13
Make throwOrPrintErrors a member of sdf::Error

marcoag:

# 🎉 New feature

Related to https://github.com/gazebosim/sdformat/pull/1141.

## Summary

Because of the request [here](https://github.com/gazebosim/sdformat/pull/1141#discussion_r1043663060) and as per IM, moving `throwOrPrintErrors` into `Error.cc` so now an Error has a method to be thrown or printed and this can be used in current SDFormat header files.

## Test it
Throw any `Error` object using the new method.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/sdf/Error.hh
+++ b/include/sdf/Error.hh
@@ -251,7 +251,7 @@ namespace sdf
     /// \brief It will print the error to _out or throw it using
     /// SDF_ASSERT depending on its ErrorCode.
     /// \param[in] _out ostream to use for printing errors.
-    public: void throwOrPrintError(sdf::Console::ConsoleStream &_out) const;
+    public: void ThrowOrPrintError(sdf::Console::ConsoleStream &_out) const;
 
     /// \brief Safe bool conversion.
     /// \return True if this Error's Code() != NONE. In otherwords, this is
--- a/include/sdf/Error.hh
+++ b/include/sdf/Error.hh
@@ -247,6 +247,10 @@ namespace sdf
     /// \sa Element::SetXmlPath
     public: void SetXmlPath(const std::string &_xmlPath);
 
+    /// \brief It will print the error to sdferr or throw it using
+    /// SDF_ASSERT depending on its ErrorCode.
+    public: void throwOrPrintError();
+
     /// \brief Safe bool conversion.
     /// \return True if this Error's Code() != NONE. In otherwords, this is
     /// true when there is an error.
--- a/include/sdf/Error.hh
+++ b/include/sdf/Error.hh
@@ -22,6 +22,7 @@
 #include <optional>
 #include <gz/utils/ImplPtr.hh>
 #include <sdf/sdf_config.h>
+#include "sdf/Console.hh"
 #include "sdf/system_util.hh"
 
 #ifdef _WIN32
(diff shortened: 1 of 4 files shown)


include/sdf/Error.hh
```diff
@@ -247,6 +247,10 @@ namespace sdf
     /// \sa Element::SetXmlPath
     public: void SetXmlPath(const std::string &_xmlPath);
 
+    /// \brief It will print the error to sdferr or throw it using
+    /// SDF_ASSERT depending on its ErrorCode.
+    public: void throwOrPrintError();
```
azeey, line 252: If it's a member function, the first letter should be uppercase.

azeey, line 252: Since it's not modifying anything, it should be a `const` function

src/Utils.cc
```diff
@@ -167,18 +167,11 @@ void enforceConfigurablePolicyCondition(
 }
 
 /////////////////////////////////////////////////
-void throwOrPrintErrors(const sdf::Errors& _errors)
+void throwOrPrintErrors(sdf::Errors& _errors)
```
azeey, line 170: Shouldn't have to make this change if the member function is `const`.

src/Error.cc
```diff
@@ -118,6 +120,19 @@ void Error::SetXmlPath(const std::string &_xmlPath)
   this->dataPtr->xmlPath = _xmlPath;
 }
 
+/////////////////////////////////////////////////
+void Error::throwOrPrintError()
```
azeey, line 124: I think it would be a good idea for this to take an `ostream` object as input so it can be used with `sdfwarn` and others.

azeey (changes requested):
quick nit: Since the title of the PR becomes part of the changelog, mind changing the tense to match the rest of the changelog? We commonly use imperative ("Move throwOrPrintErrors...") or past tense ("Moved throwOrPrintErrors...")

src/Error.cc
marcoag: `sdferr` and `sdfwarn` seem to be `ConsoleStream` shall it take that one as input instead of the `ostream`?

include/sdf/Error.hh
marcoag: Done in https://github.com/gazebosim/sdformat/pull/1220/commits/f6f171594b4e6a7864c696318a961d1cb4bea061.

src/Utils.cc
marcoag: Changed back in https://github.com/gazebosim/sdformat/pull/1220/commits/6483e3ab20de09fc28f83e16ed45a67297419466.

include/sdf/Error.hh
marcoag: True, done in https://github.com/gazebosim/sdformat/pull/1220/commits/6483e3ab20de09fc28f83e16ed45a67297419466.

src/Error.cc
marcoag: Added a `ConsoleStream` for now in https://github.com/gazebosim/sdformat/pull/1220/commits/6483e3ab20de09fc28f83e16ed45a67297419466. Let me know if we should do something different.

azeey: Yeah, I thought `ConsoleStream` inherited from `ostream`, but that's not the case, so what you have is good.

azeey (approved):

PR #1231, opened 2023-02-08
Add missing sdf files from xsd generation

mjcarroll:
These `.sdf` files were not being used as part of the schema generation process.  I have re-added them here.

### DIFF
--- a/sdf/1.10/CMakeLists.txt
+++ b/sdf/1.10/CMakeLists.txt
@@ -36,6 +36,7 @@ set (sdfs
   mesh_shape.sdf
   model.sdf
   model_state.sdf
+  navsat.sdf
   noise.sdf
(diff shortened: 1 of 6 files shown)


azeey (approved):
LGTM! 

Not sure why we're not just globing for `*.sdf` files here.

mjcarroll:
> Not sure why we're not just globing for *.sdf files here.

Likely a leftover from local development.  If you glob, you have to be sure to rebuild the cmake cache to get newly-created files.  If someone was using `make && make install`, it wouldn't have used any new files.

scpeters:
The `collision_engine` and `urdf` files were not included from any other files, so they were removed from 1.10 in https://github.com/gazebosim/sdformat/pull/1113, but the navsat miss was a definite bug

PR #1235, opened 2023-02-08
Go back to SDF_ASSERT instead of FATAL_ERROR

marcoag:
# 🦟 Bug fix

## Summary

It seems quite hard to guarantee the same flow of the execution of `SDF_ASSERT` when using a `FATAL_ERROR` instead. Because of this, as explained by [this comment](https://github.com/gazebosim/sdformat/pull/1141#issuecomment-1410826445), this PR reverts all previous `SDF_ASSERT` changes into `FATAL_ERROR` done by the following PRs:

https://github.com/gazebosim/sdformat/pull/1123
https://github.com/gazebosim/sdformat/pull/1095

## Checklist
- [ ] Added tests

### DIFF
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -559,6 +559,11 @@ namespace sdf
     /// \param[in] _child Pointer to the child to remove.
     public: void RemoveChild(ElementPtr _child);
 
+    /// \brief Remove a child element.
+    /// \param[in] _child Pointer to the child to remove.
+    /// \param[out] _errors Vector of errors.
+    public: void RemoveChild(ElementPtr _child, sdf::Errors &_errors);
+
     /// \brief Remove all child elements.
     public: void ClearElements();
 
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -559,11 +559,6 @@ namespace sdf
     /// \param[in] _child Pointer to the child to remove.
     public: void RemoveChild(ElementPtr _child);
 
-    /// \brief Remove a child element.
-    /// \param[in] _child Pointer to the child to remove.
-    /// \param[out] _errors Vector of errors.
-    public: void RemoveChild(ElementPtr _child, sdf::Errors &_errors);
-
     /// \brief Remove all child elements.
     public: void ClearElements();
 
--- a/src/Element.cc
+++ b/src/Element.cc
@@ -1283,7 +1283,7 @@ void Element::RemoveChild(ElementPtr _child)
 }
 
 /////////////////////////////////////////////////
-void Element::RemoveChild(ElementPtr _child, sdf::Errors &_errors)
+void Element::RemoveChild(ElementPtr _child, sdf::Errors &)
 {
   SDF_ASSERT(_child, "Cannot remove a nullptr child pointer");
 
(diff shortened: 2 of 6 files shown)


include/sdf/Element.hh
```diff
@@ -559,11 +559,6 @@ namespace sdf
     /// \param[in] _child Pointer to the child to remove.
     public: void RemoveChild(ElementPtr _child);
 
-    /// \brief Remove a child element.
-    /// \param[in] _child Pointer to the child to remove.
-    /// \param[out] _errors Vector of errors.
-    public: void RemoveChild(ElementPtr _child, sdf::Errors &_errors);
```
azeey, line 565: We can't remove this because it breaks ABI :(.

azeey (changes requested):
Unfortunately, we can't remove `Element::RemoveChild` because it would break ABI since it's already been released. We should make sure it works the same as the other `RemoveChild` with a test. Otherwise, LGTM.

include/sdf/Element.hh
marcoag: Oh true! shall we deprecate it instead and remove it on the the next release?

marcoag:
Brought `Element::RemoveChild` with the `sdf::Errors` param back and added a test to make sure both throw: https://github.com/gazebosim/sdformat/pull/1235/commits/d9474b39d27604fc70f48187423d561d830f340c.

include/sdf/Element.hh
marcoag: I'll open a PR to `main` then. Would an `sdf::Error` with an `ErrorCode::WARNING` message suffice?

azeey (approved):

azeey:
Ah, now there's a compiler warning saying `_errors` is not used.

marcoag:
Yup, any suggestion on how to avoid it or shall we just live with it?

src/Element.cc
```diff
@@ -1311,12 +1285,7 @@ void Element::RemoveChild(ElementPtr _child)
 /////////////////////////////////////////////////
 void Element::RemoveChild(ElementPtr _child, sdf::Errors &_errors)
```
azeey, line 1286: This should avoid the compiler warning
```suggestion
void Element::RemoveChild(ElementPtr _child, sdf::Errors &)
```

include/sdf/Element.hh
azeey: Not sure what the `ErrorCode::WARNING` would be for.

azeey (approved):

include/sdf/Element.hh
marcoag: Deprecation message?

azeey: Oh, we'd use the `SDF_DEPRECATED` macro for that.

marcoag: `GZ_DEPRECATED` maybe?

azeey: Yeah, `GZ_DEPRECATED`. I guess `SDF_DEPRECATED` is deprecated 😁 

We can discuss this on the followup PR, but if there's a chance the `sdf::Error` might be used by `RemoveChild`, it might be better to deprecate the overload that doesn't take `sdf::Error`.

PR #1256, opened 2023-03-21
Fix Element Set method return value

marcoag:
# 🦟 Bug fix

## Summary

Discovered [here](https://github.com/gazebosim/sdformat/pull/1141#discussion_r1140648979), the Element Set method would `return true` ignoring the result coming from the `this->dataPtr->value->Set(_value);`. This could cause the Set method to sometimes error but still return `true`.

 This PR fixes it and it makes the method return the result of `this->dataPtr->value->Set(_value);` instead of the hardcoded `true`.

## Checklist
- [x] Added tests

### DIFF
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -653,8 +653,7 @@ namespace sdf
   {
     if (this->dataPtr->value)
     {
-      this->dataPtr->value->Set(_value);
-      return true;
+      return this->dataPtr->value->Set(_value);
     }
     return false;
   }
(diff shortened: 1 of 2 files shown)


scpeters:
> @scpeters Do you think this is safe for Gazebo-classic?

I looked and I don't see a single place where the return value of `Element::Set` is checked. So that's maybe concerning on its own, but it means that this pull request shouldn't affect gazebo-classic. Thanks for checking

scpeters (approved):

azeey (approved):

marcoag:

Issue #552, opened 2021-04-29
Existing GSG violations: non-trivially destructible types at a static storage duration?


EricCousineau-TRI:
## Environment
N/A

## Description
* Expected behavior: All static storage duration stuff (`static`, global `const`, etc.) should be trivially destructible per GSG
  * https://ignitionrobotics.org/docs/all/contributing#style-guides
  * https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables
* Actual behavior: It's still a tad mixed atm:

https://github.com/osrf/sdformat/blob/689ee8410a97917e2087ffd5045861d63258ebd4/include/sdf/Types.hh#L64-L72

## Steps to reproduce
N/A

## Output
N/A

EricCousineau-TRI:
May not be easiest to fix given that some of these are part of public API.

Some options:
- Using things ilke `std::string` is important, then make it a function which returns a copy (or even a const reference to) something that is wrap in something like Drake's `never_destroyed<>` class:  https://drake.mit.edu/styleguide/cppguide.html#Static_and_Global_Variables (drake::never_destroyed<T>)
  - Deprecate old spelling.
- Change the type, and just change immediate downstream usages. (Ew).

FYI @azeey

PR #1264, opened 2023-03-28
Fix GSG violations on non-trivially destructible types

aaronchongth:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/sdformat/issues/552

## Checklist
- [ ] Added tests

### DIFF
--- a/include/sdf/Types.hh
+++ b/include/sdf/Types.hh
@@ -35,7 +35,7 @@ namespace sdf
   inline namespace SDF_VERSION_NAMESPACE {
   //
 
-  const std::string kSdfScopeDelimiter = "::";
+  constexpr char kSdfScopeDelimiter[] = "::";
 
   /// \brief The source path replacement if it was parsed from a string,
   /// instead of a file.
--- a/include/sdf/Types.hh
+++ b/include/sdf/Types.hh
@@ -36,12 +36,21 @@ namespace sdf
   inline namespace SDF_VERSION_NAMESPACE {
   //
 
-  static const std::string &SdfScopeDelimiter()
+  namespace internal
   {
-    static const gz::utils::NeverDestroyed<std::string> delimiter{"::"};
-    return delimiter.Access();
-  }
-  static const std::string &kSdfScopeDelimiter = SdfScopeDelimiter();
+    /// \brief Initializes the scope delimiter as a function-local static
+    /// variable so it can be used to initialize kSdfScopeDelimiter.
+    /// \note This should not be used directly in user code. It will likely be
+    /// removed in libsdformat 15 with kSdfScopeDelimiter.
+    SDFORMAT_VISIBLE const std::string &SdfScopeDelimiter();
+  }  // namespace internal
+
+  constexpr std::string_view kScopeDelimiter{"::"};
+
+  // Deprecated because it violates the Google Style Guide as it is not
+  // trivially destructible. Please use sdf::kScopeDelimiter instead.
+  GZ_DEPRECATED(14)
+  inline const std::string &kSdfScopeDelimiter = internal::SdfScopeDelimiter();
 
   /// \brief The source path replacement if it was parsed from a string,
   /// instead of a file.
--- a/include/sdf/Types.hh
+++ b/include/sdf/Types.hh
@@ -22,6 +22,7 @@
 #include <cstdint>
 #include <sstream>
 #include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
@@ -35,7 +36,7 @@ namespace sdf
   inline namespace SDF_VERSION_NAMESPACE {
   //
 
-  constexpr char kSdfScopeDelimiter[] = "::";
+  constexpr std::string_view kSdfScopeDelimiter = "::";
 
   /// \brief The source path replacement if it was parsed from a string,
   /// instead of a file.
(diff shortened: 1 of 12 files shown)


src/Sensor.cc
```diff
@@ -40,7 +41,7 @@ using namespace sdf;
 
 /// Sensor type strings. These should match the data in
 /// `enum class SensorType` located in Sensor.hh.
-const std::vector<std::string> sensorTypeStrs =
+constexpr std::array<std::string_view, 27> kSensorTypeStrs =
```
azeey, line 44: Can you add a comment here to remind us that we need to update the size template parameter if we add anything to this array.

src/Types.cc
```diff
@@ -142,7 +142,7 @@ std::string JoinName(
   else if (scopeNameEndsWithDelimiter || localNameStartsWithDelimiter)
     return _scopeName + _localName;
   else
-    return _scopeName + kSdfScopeDelimiter + _localName;
+    return _scopeName + std::string(kSdfScopeDelimiter) + _localName;
```
azeey, line 145: Ah, this is what I was talking about. We shouldn't break this type of code without a tick-tock.

src/parser_urdf.cc
```diff
@@ -56,9 +56,9 @@ typedef std::map<std::string, std::vector<SDFExtensionPtr> >
 StringSDFExtensionPtrMap g_extensions;
 bool g_reduceFixedJoints;
 bool g_enforceLimits;
-const char kCollisionExt[] = "_collision";
-const char kVisualExt[] = "_visual";
-const char kLumpPrefix[] = "_fixed_joint_lump__";
+constexpr char kCollisionExt[] = "_collision";
```
azeey, line 59: not sure if this is necessary, but okay.

azeey (changes requested):
I like the use of `std::string_view`. Much cleaner than trying to use `const char*`. The only main issue is how we tick-tock `kSdfScopeDelimiter`. Otherwise, LGTM!

src/parser_urdf.cc
aaronchongth: I didn't understand enough what `constexpr` does, you're right this is redundant. I've reverted it, https://github.com/gazebosim/sdformat/pull/1264/commits/53f017607b93819cea33b85fb5b8c04a436b432d

src/Sensor.cc
aaronchongth: https://github.com/gazebosim/sdformat/pull/1264/commits/b62510fd538f51c6c866ff9a1b477559ddff1e84

src/Filesystem.cc
```diff
@@ -210,7 +210,7 @@ void DirIter::close_handle()
 
 #else  // Windows
 
-static const char preferred_separator = '\\';
+constexpr char kSdfFileSystemPreferredSeparator[] = '\\';
```
azeey, line 213: This should be reverted. The separator is actually a single `\` character. Two backslashes are necessary to escape it.

aaronchongth: My bad, that was a horrific misunderstanding of the code, https://github.com/gazebosim/sdformat/pull/1264/commits/e05d4ecc1d70ed67ca1798e4eeb9882066eef645

include/sdf/Types.hh
```diff
@@ -35,7 +36,7 @@ namespace sdf
   inline namespace SDF_VERSION_NAMESPACE {
   //
 
-  const std::string kSdfScopeDelimiter = "::";
+  constexpr std::string_view kSdfScopeDelimiter = "::";
```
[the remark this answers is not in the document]
aaronchongth: Thanks for the tip! I started using it in https://github.com/gazebosim/sdformat/pull/1264/commits/3f5afcf72e39c1c572d94fbb3127e7d5665b0dd6. Let me know what you think

include/sdf/Types.hh
```diff
@@ -35,7 +36,12 @@ namespace sdf
   inline namespace SDF_VERSION_NAMESPACE {
   //
 
-  const std::string kSdfScopeDelimiter = "::";
+  static const std::string &SdfScopeDelimiter()
```
azeey, line 39: Per the [GSG](https://google.github.io/styleguide/cppguide.html#Internal_Linkage), we should avoid using `static` for the function here as well as for `kSdfScopeDelimiter` since this is a header file.

src/Camera.cc
```diff
@@ -23,7 +24,7 @@ using namespace sdf;
 
 /// \brief String names for the pixel formats.
 /// \sa Image::PixelFormat.
-static std::array<std::string, 19> kPixelFormatNames =
+static const std::array<const char *, 19> kPixelFormatNames =
```
azeey, line 27: Did `string_view` not work? I see it in `ParticleEmitter`, but not here. I'm okay with either, but it would be nice to make them all the same.

aaronchongth: The only difference is the `const`, I've just committed https://github.com/gazebosim/sdformat/pull/1264/commits/632570e4151919ac5e6f1b58d63088420d634757 to see how the macOs and win runners will react

azeey: seems to be working!

aaronchongth: yeah, looks like the compiler only considers `const std::array<const std::string_view, ...>` as literal

include/sdf/Types.hh
aaronchongth: Thanks for helping with this! https://github.com/gazebosim/sdformat/pull/1265

azeey (approved):

PR #1290, opened 2023-06-09
Remove robot not found error when parsing fails

### LINKED ISSUE #151: Trying to parse an invalid file gives URDF parser error
**Original report ([archived issue](https://osrf-migration.github.io/sdformat-gh-pages/#!/osrf/sdformat/issues/151)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

When trying to parse an invalid SDF file, we get two error messages:

~~~
Error [parser_urdf.cc:3423] Unable to call parseURDF on robot model
Error [parser.cc:340] parse as old deprecated model file failed.
~~~

These messages are very misleading, because they suggest we're trying to parse an URDF file or a deprecated file, when in reality, it might simply be a malformed SDF file.

What happens is that the [readString](https://github.com/osrf/sdformat/blob/1b5b72c771f4d63341115e0886780ca45cc7a7e2/src/parser.cc#L332) function first tries to load as an SDF, and if that fails, it silently tries to load the file as URDF.

marcoag:
# 🦟 Bug fix

Fixes #151

## Summary
When a file fails to be parsed as SDF the `parser.cc` tries to parse it as URDF generating the following error:

```
Error:   Could not find the 'robot' element in the xml file
             at line 109 in ./urdf_parser/src/model.cpp
```

This becomes confusing the file could just be a malformed SDF. 

This PR adds a check for the `robot` tag before attempting the parse in order to avoid the misleading error. It also adds an error in case it's not found indicating that the file could not be identified as SDF nor URDF. 

NOTE: I'm targeting `sdf13` as I consider this a bug fix but let me know if this should go into `master`. 

## Checklist
- [ ] Added tests

### DIFF
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -768,15 +768,16 @@ bool readFileInternal(const std::string &_filename, const bool _convert,
     return false;
   }
 
-  // Suppress deprecation for sdf::URDF2SDF
-  if (readDoc(&xmlDoc, _sdf, filename, _convert, _config, _errors))
+  tinyxml2::XMLElement *sdfXml = xmlDoc.FirstChildElement("sdf");
+  if (sdfXml)
   {
-    return true;
+    // Suppress deprecation for sdf::URDF2SDF
+    return readDoc(&xmlDoc, _sdf, filename, _convert, _config, _errors);
   }
   else
   {
     tinyxml2::XMLElement *robotXml = xmlDoc.FirstChildElement("robot");
-    if (robotXml && URDF2SDF::IsURDF(filename))
+    if (robotXml)
     {
       URDF2SDF u2g;
       auto doc = makeSdfDoc();
@@ -794,7 +795,7 @@ bool readFileInternal(const std::string &_filename, const bool _convert,
     }
     else
     {
-      sdferr << "XML does not seem to be an SDF or an URDF file.\n";
+      sdferr << "XML does not seem to be an SDFormat or an URDF file.\n";
     }
   }
 
@@ -853,10 +854,11 @@ bool readStringInternal(const std::string &_xmlString, const bool _convert,
     sdferr << "Error parsing XML from string: " << xmlDoc.ErrorStr() << '\n';
     return false;
   }
-  if (readDoc(&xmlDoc, _sdf, std::string(kSdfStringSource), _convert, _config,
-              _errors))
+  tinyxml2::XMLElement *sdfXml = xmlDoc.FirstChildElement("sdf");
+  if (sdfXml)
   {
-    return true;
+    return readDoc(&xmlDoc, _sdf, std::string(kSdfStringSource), _convert,
+                   _config, _errors);
   }
   else
   {
@@ -881,10 +883,10 @@ bool readStringInternal(const std::string &_xmlString, const bool _convert,
     }
     else
     {
-      sdferr << "XML does not seem to be an SDF or an URDF string.\n";
+      sdferr << "XML does not seem to be an SDFormat or an URDF string.\n";
+      return false;
     }
   }
-
   return false;
 }
 
--- a/src/parser.cc
+++ b/src/parser.cc
@@ -784,12 +784,14 @@ bool readFileInternal(const std::string &_filename, const bool _convert,
       u2g.InitModelFile(filename, _config, &doc);
       if (sdf::readDoc(&doc, _sdf, filename, _convert, _config, _errors))
       {
-        sdfdbg << "parse from urdf file [" << _filename << "].\n";
+        sdfdbg << "Converting URDF file [" << _filename << "] to SDFormat"
+               << " and parsing it.\n";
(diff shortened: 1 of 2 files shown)


src/parser.cc
```diff
@@ -773,20 +773,28 @@ bool readFileInternal(const std::string &_filename, const bool _convert,
...
   }
-  else if (URDF2SDF::IsURDF(filename))
+  else
   {
-    URDF2SDF u2g;
-    auto doc = makeSdfDoc();
-    u2g.InitModelFile(filename, _config, &doc);
-    if (sdf::readDoc(&doc, _sdf, filename, _convert, _config, _errors))
+    tinyxml2::XMLElement *robotXml = xmlDoc.FirstChildElement("robot");
+    if (robotXml && URDF2SDF::IsURDF(filename))
```
azeey, line 779: Calling `IsURDF` seems unnecessary at this point and wastes cycles since it actually fully parses the file. IMO, if the first tag is `robot`, we can assume it's a URDF and proceed.

src/parser.cc
```diff
@@ -773,20 +773,28 @@ bool readFileInternal(const std::string &_filename, const bool _convert,
...
+      auto doc = makeSdfDoc();
+      u2g.InitModelFile(filename, _config, &doc);
+      if (sdf::readDoc(&doc, _sdf, filename, _convert, _config, _errors))
+      {
+        sdfdbg << "parse from urdf file [" << _filename << "].\n";
+        return true;
+      }
+      else
+      {
+        sdferr << "parse as old deprecated model file failed.\n";
```
azeey, line 791: This error message is not really correct. Not 100% sure what it should say. Maybe "Failed to parse the URDF file after converting to SDFormat"?

src/parser.cc
```diff
@@ -773,20 +773,28 @@ bool readFileInternal(const std::string &_filename, const bool _convert,
...
+        return false;
+      }
     }
     else
     {
-      sdferr << "parse as old deprecated model file failed.\n";
-      return false;
+      sdferr << "XML does not seem to be an SDF or an URDF file.\n";
```
azeey, line 797: SDF -> SDFormat.

src/parser.cc
```diff
@@ -852,20 +860,28 @@ bool readStringInternal(const std::string &_xmlString, const bool _convert,
...
+
+      if (sdf::readDoc(&doc, _sdf, std::string(kUrdfStringSource), _convert,
+                      _config, _errors))
+      {
+        sdfdbg << "Parsing from urdf.\n";
+        return true;
+      }
+      else
+      {
+        sdferr << "parse as old deprecated model file failed.\n";
```
azeey, line 878: See my comment above about this error message

azeey (commented):
This is a much needed UX improvement, so thanks!! I've left some minor comments, but overall it looks great! Would you be able to add some tests?

src/parser.cc
marcoag: Sounds about right. I was not sure how much removing this parsing would modify behavior though so I opted to minimize changes, if you think it's OK I'll remove it then.

src/parser.cc
marcoag: I agree this is confusing although I was trying to keep the scope of this PR focused on the "fix" of removing the "robot" related message, dunno if this is comes as part of it or it should go into `main` as a behavior change, what do you think?

marcoag:
When adding the tests I realized that we need to check for the `<sdf>` tag and ONLY if not present then we could print the "nor SDFormat nor URDF" fail. Otherwise the error will show up on malformed SDFormat files and it gets confusing.

Added relevant changes in https://github.com/gazebosim/sdformat/pull/1290/commits/029b204e98cf5c01399b8e4297bea4f1dbaeac20.

src/parser.cc
marcoag: Changed in https://github.com/gazebosim/sdformat/pull/1290/commits/029b204e98cf5c01399b8e4297bea4f1dbaeac20.

src/parser.cc
marcoag: Removed in https://github.com/gazebosim/sdformat/pull/1290/commits/029b204e98cf5c01399b8e4297bea4f1dbaeac20.

src/parser.cc
marcoag: Done in https://github.com/gazebosim/sdformat/pull/1290/commits/41bb5bd0a1e29442d84cad4503677c9d4286e9ac.

src/parser.cc
marcoag: Done in https://github.com/gazebosim/sdformat/commit/41bb5bd0a1e29442d84cad4503677c9d4286e9ac.

azeey (approved):

PR #1317, opened 2023-08-30
Bugifx: Resolving Inertial Pose in `Collision::CalculateInertial()`

jasmeet0915:
# 🦟 Bug fix

## Summary
This PR fixes the resolving of inertial pose in `Collision::CalculateInertial()` by adding the calculated inertial pose returned from `Geometry::CalculateInertial()` to the pose of the collision.

Also shifted the `this->dataPtr->sdf` based check for loading the `<auto_inertia_params>` element from `Collision::CalculateInertial()` to `Collision::Load()`. This would allow users to programmatically create collisions and call CalculateInertial() function on them.

## Checklist
- [ ] Added tests


src/Collision.cc
```diff
@@ -287,7 +294,7 @@ void Collision::CalculateInertial(
     // Else resolve collision pose in Link Frame and then set as inertial pose
     if (this->dataPtr->poseRelativeTo.empty())
     {
-      _inertial.SetPose(this->dataPtr->pose);
+      _inertial.SetPose(_inertial.Pose() * this->dataPtr->pose);
```
azeey, line 297: Using notation from http://sdformat.org/tutorials?tut=specify_pose&cat=specification&
`X_LI`:  Pose of the inertial `I` relative to the link `L` .
`X_LC`:  Pose of the collision relative to `L`
`X_CG`: Pose of the inertial returned by the auto MOI calculator, which we take as being relative to the collision

Since we want `_inertial` to be in frame `L`, we would want

`X_LG = X_LC * L_CG and `X_LG = X_LI`.

So, I think the terms are swapped. What do you think?

jasmeet0915: Oh yeah that makes sense. I have swapped the terms in ca09b95b25a686845ea0e2341717053f79dc2640

azeey (approved):

PR #1319, opened 2023-08-31
Correct spelling of Python variable in CMake

mjcarroll:
Not critical, but does seem to have issues, at least on Windows.  

I believe that the problem is that COMMAND is expected to be one argument, so passing the `python.exe` plus the script name violates that.  I'm not sure why this hasn't been an issue before, but it causes all sorts of hilariously bad behavior on Windows (popping open the Microsoft store of VS code rather than trying to run the script).

### DIFF
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -12,4 +12,3 @@ python3-gz-math7
 python3-psutil
 python3-pybind11
 python3-pytest
-ruby-dev
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,15 +110,6 @@ if (BUILD_SDF)
     gz_build_warning("Python psutil package not found. Memory leak tests will be skipped")
   endif()
 
-  ################################################
-  # Find ruby executable to produce xml schemas
-  find_program(RUBY ruby)
-  if (NOT RUBY)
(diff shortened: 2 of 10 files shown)


peci1:
Got hit by this issue, too (in my case, opening GVim.exe :) ).

The resolution is much simpler, though:

https://github.com/gazebosim/sdformat/blob/881cbb7f4bab89d052095a43a11cf30059a174f4/sdf/1.5/CMakeLists.txt#L70

See, Pyhton ? :-D It's only in the 1.5 SDF CMakeLists, elsewhere it's okay.

@mjcarroll I think you can revert the other changes in your PR as they don't really make sense. CMake docs say `ARGS` is only for backwards compatibility and is ignored. So my interpretation is that command is always the first thing after `COMMAND`, while other things are args.

mjcarroll:
Good call, reverted.

> I guess it didn't manifest on the other operating systems as their default is to run .py files with an interpreter instead of an IDE.

Yeah, we have a shebang line in the `xmlschema.py`, so it should just do the right thing on platforms that understand that.  Windows is not one of those, unfortunately, so it would have all sorts of silly side effects.

azeey (approved):

azeey:
The title of the PR should be renamed though.

PR #1323, opened 2023-09-07
Remove hard dependence on ruby

mjcarroll:
The dep will get pulled in by gz-tools


scpeters (approved):
yes, we did a lot of work to not need ruby here, so we should definitely remove this, and likewise we need to indicate that python is now required

scpeters:
follow-up: https://github.com/gazebo-release/sdformat14-release/pull/2

PR #1325, opened 2023-09-23
ParserConfig: save auto inertial values by default

scpeters:
# 🦟 Bug fix

Follow-up to https://github.com/gazebosim/sdformat/pull/1299#discussion_r1302493127

## Summary

While reviewing #1299, I suggested adding a parameter to `ParserConfig` so that inertial values could be computed by default, but the calculation could be skipped by setting the appropriate parameter before loading, which is needed by `gz-sim`. At least, my intent was to point out that saving auto-inertial values should be the default, but I didn't articulate it explicitly, and the implementation skips by default instead.

This changes the default configuration of `ParserConfig` to save auto inertial values instead of skipping. Without this change `gz sdf --inertial-stats` doesn't resolve auto-inertial values (confirm by building https://github.com/gazebosim/sdformat/commit/ab438f3ea149701c47b6a444292ac2de0927cb55 and observing the failure of `UNIT_gz_TEST`).

Minor changes:

* https://github.com/gazebosim/sdformat/commit/36de5422781f1ce0e0edc09680d0eeef8fa583ef: fix xml syntax in `test/sdf/inertial_stats.sdf`
* https://github.com/gazebosim/sdformat/commit/ab438f3ea149701c47b6a444292ac2de0927cb55: add `test/sdf/inertial_stats_auto.sdf` based on `test/sdf/inertial_stats.sdf` with `//inertial/@auto=true` and `//collision/density` values that yield equivalent inertial values
* https://github.com/gazebosim/sdformat/commit/4b1112e26529220d4fcfb6fbdde6ded63599312c: minor changes to error messages

## Checklist
- [X] Added tests

### DIFF
--- a/src/Geometry.cc
+++ b/src/Geometry.cc
@@ -312,6 +312,7 @@ void Geometry::SetPolylineShape(const std::vector<Polyline> &_polylines)
   this->dataPtr->polylines = _polylines;
 }
 
+/////////////////////////////////////////////////
 std::optional<gz::math::Inertiald> Geometry::CalculateInertial(
   sdf::Errors &_errors, const ParserConfig &_config,
   double _density, sdf::ElementPtr _autoInertiaParams)
--- a/src/Link.cc
+++ b/src/Link.cc
@@ -191,9 +191,9 @@ Errors Link::Load(ElementPtr _sdf, const ParserConfig &_config)
       {
         Error err(
           ErrorCode::WARNING,
-          "Inertial was used with auto=true for link, " +
-          this->dataPtr->name + " but user defined inertial values "
-          "were found. They would be overwritten by computed ones."
+          "Inertial was used with auto=true for the link named " +
+          this->dataPtr->name + ", but user-defined inertial values were "
+          "found, which will be overwritten by the computed inertial values."
         );
         enforceConfigurablePolicyCondition(
           _config.WarningsPolicy(), err, errors);
@@ -610,7 +610,8 @@ void Link::ResolveAutoInertials(sdf::Errors &_errors,
     {
       _errors.push_back({ErrorCode::ELEMENT_MISSING,
                         "Inertial is set to auto but there are no "
-                        "<collision> elements for the link."});
+                        "<collision> elements for the link named " +
+                        this->Name() + "."});
       return;
     }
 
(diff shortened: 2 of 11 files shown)


azeey:
I added a test in https://github.com/gazebosim/sdformat/pull/1325/commits/c550e3d8aaa90b231e83fd2ee6746c6f7e3d8ed4  that fails because a MOI calculator callback is not set in `sdf::ParserConfig`. `Mesh::CalculateInertial` throws an exception because the callback is used without checking if it's set. I tried adding
```c++
  if (!customCalculator)
  {
    _errors.push_back(
        {sdf::ErrorCode::WARNING,
         "Custom moment of inertia calculator for meshes not set via "
         "sdf::ParserConfig::RegisterCustomInertiaCalc"});
    return std::nullopt;
  }
```
but I got another error from `Collision.cc` saying the "Inertia Calculated for collision is invalid" : https://github.com/gazebosim/sdformat/blob/c550e3d8aaa90b231e83fd2ee6746c6f7e3d8ed4/src/Collision.cc#L277-L282

Do we want it to just use the default inertial if a callback is not set?

scpeters:
> I added a test in [c550e3d](https://github.com/gazebosim/sdformat/commit/c550e3d8aaa90b231e83fd2ee6746c6f7e3d8ed4) that fails because a MOI calculator callback is not set in `sdf::ParserConfig`. `Mesh::CalculateInertial` throws an exception because the callback is used without checking if it's set. I tried adding
 
> but I got another error from `Collision.cc` saying the "Inertia Calculated for collision is invalid" :
> 
> https://github.com/gazebosim/sdformat/blob/c550e3d8aaa90b231e83fd2ee6746c6f7e3d8ed4/src/Collision.cc#L277-L282
> 
> Do we want it to just use the default inertial if a callback is not set?

I think we could use the default inertial with a warning that could be upgraded to an error with the right `ParserConfig` setting

scpeters:
updated in https://github.com/gazebosim/sdformat/pull/1325/commits/d6d8e7e20d1f2c337b224a4d2ebce900af75ea52

azeey (approved):
Returning the default Inertial instead of a nullopt makes more sense. Now, we only get the error from `Mesh.cc`.

Issue #1329, opened 2023-09-27
Use //link/inertial/density element in parser


scpeters:
## Desired behavior

After describing the `//collision/density` element, the [Proposal for Automatic Moments of Inertia Calculations](http://sdformat.org/tutorials?tut=auto_inertial_params_proposal&cat=pose_semantics_docs&) states: "A `//link/inertial/density` element would also be added in the spec to allow users to specify the density values on a link level instead of specifying the same values for each collision."

This element ~~should be~~ was added to the spec in https://github.com/gazebosim/sdformat/pull/1299 but still needs to be used in the parser.

scpeters:
see https://github.com/gazebosim/sdformat/pull/1335

PR #1304, opened 2023-07-28
Callback based API for Implementing Custom Moment of Inertia Calculators for Meshes

jasmeet0915:
# 🎉 New feature

Based on [this](http://sdformat.org/tutorials?tut=auto_inertial_params_proposal&branch=test_equation_rendering&repo=https%3A%2F%2Fgithub.com%2Fjasmeet0915%2Fsdf_tutorials) proposal.

> **Note:** This PR would require merging #1298 and #1299 

## Summary
 * Added functions to register custom calculator and get custom calculator in `ParserConfig`.
 * Added an `InterfaceMoiCalculator` class that can be used to send appropriate data to the calculator from sdformat.
 * Added a `<moi_calculator_params>` tag that can be used to access custom calculator params from the SDF by using tags with namespaces. (See [this](http://sdformat.org/tutorials?tut=custom_elements_attributes_proposal&cat=pose_semantics_docs&) for reference)
 * Added `MassMatrix()` function to `sdf::Mesh` class and updated the `Collision::MassMatrix()` function to pass the parser config as well as `sdf::ElementPtr` to `<moi_calculator_params>` tag.

## Demo & Usage Example

**Demo 1:** This demo shows the automatic inertia calculation feature on a rubber ducky model which is a non-convex mesh. On the left, we have the rubber ducky mesh with automatic calculations enabled and on the right, the mesh uses the default values.
> **Note:** The inertial values are due to the scale of the mesh. You can see the banana for scale in between the 2 ducks. A density value for the duck was used which was calculated by using the mass and volume data of the duck found online.

> **Note:** The `voxel_size` inertia param given in the snippet below is just for showing how the `<auto_inertia_params>` element could be used. This is not actually used by the calculator since we are not using a voxelization-based calculator.
> 
<details>
  <summary>SDF snippet for the duck mesh with auto inertial</summary>

  ```xml
<?xml version="1.0" ?>
<sdf version="1.6">
  <model name="duck">
    <link name="duck_link">
      <pose>0 0 0 0 0 0</pose>
      <inertial auto="true" />
      <collision name="duck_collision">
      	<pose>0 0 0 0 0 0</pose>
        <density>111.8</density>
        <auto_inertia_params>
          <gz:voxel_size>0.01</gz:voxel_size>
        </auto_inertia_params>
        <geometry>
          <mesh>
            <uri>meshes/duck_collider.dae</uri>
          </mesh>
        </geometry>
      </collision>
      <visual name="duck_visual">
[...]

PR #1330, opened 2023-09-27
Collision: add optional density parameter to API

scpeters:
# 🎉 New feature

Quick API change in support of https://github.com/gazebosim/sdformat/issues/1329

## Summary

This adds a `std::optional<double>` parameter to override the density in `Collision::CalculateInertial` to support the `//link/inertial/density` parameter. This is submitted as a quick API change before the Harmonic release without the implementation of the parameter. If we don't merge this, then we can add an overloaded `Collision::CalculateInertial` method afterwards.

cc @jasmeet0915 

## Checklist
- [ ] Added tests

### DIFF
--- a/include/sdf/Collision.hh
+++ b/include/sdf/Collision.hh
@@ -18,6 +18,7 @@
 #define SDF_COLLISION_HH_
 
 #include <memory>
+#include <optional>
 #include <string>
 #include <gz/math/Pose3.hh>
 #include <gz/math/Vector3.hh>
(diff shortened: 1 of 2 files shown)


jasmeet0915:
Thank you for the changes @scpeters. I would also like to point out that a link-level `<auto_inertia_params>` element (`//link/inertial/auto_inertia_params`) was also added in #1304 for similar reasons as the link-level `<density>` element (this is also mentioned in the [proposal](http://sdformat.org/tutorials?tut=auto_inertial_params_proposal&cat=pose_semantics_docs&)). So should we further modify the API to take an optional `sdf::ElementPtr` for the same?
cc: @azeey  @quarkytale

azeey:
I would prefer to wait until after Harmonic to add an overload that also takes into account `auto_inertia_params`.

scpeters:
we will add an overloaded API after the harmonic stable release and include the implementation at that time

Issue #867, opened 2022-03-05
`ToString` doesn't skip default elements and attributes on `sdf::ElementPtr` returned from `ToElement` 


azeey:
## Environment
* OS Version: any
* Source or binary build? `sdf12`

## Description
When using `sdf::Element::ToString`, libsdformat currently checks to see if an element or attribute is set by the user and skips printing it if it's not. This helps avoid verbose SDFormat files with elements and attributes having default values. However, this behavior only works if the `sdf::ElementPtr` was obtained from loading a file/string via `sdf::readFile`/`sdf::readString` or calling the `Element` function on DOM objects (e.g., `sdf::Model::Element`). It does not work if the `sdf::ElementPtr` is obtained form calling the `ToElement` function on DOM objects (e.g., `sdf::Model::ToElement`). 

Example:

```xml
<?xml version="1.0" ?>
<sdf version="1.6">
  <model name="test_model">
    <link name="link"/>
  </model>
</sdf>
```

Using `model->Element()->ToString("")`

```xml
<model name='test_model'>
  <link name='link'/>
</model>
```

Using `model->ToElement()->ToString("")`

```xml
<model name='test_model'>
  <static>false</static>
  <self_collide>false</self_collide>
  <allow_auto_disable>true</allow_auto_disable>
  <enable_wind>false</enable_wind>
  <pose>0 0 0 0 -0 0</pose>
  <link name='link'>
    <pose>0 0 0 0 -0 0</pose>
    <inertial>
      <pose>0 0 0 0 -0 0</pose>
      <mass>1</mass>
      <inertia>
        <ixx>1</ixx>
        <ixy>0</ixy>
        <ixz>0</ixz>
        <iyy>1</iyy>
        <iyz>0</iyz>
        <izz>1</izz>
      </inertia>
    </inertial>
    <enable_wind>false</enable_wind>
  </link>
</model>
```

Relates to : https://github.com/ignitionrobotics/sdformat/pull/575, https://github.com/ignitionrobotics/sdformat/pull/842

PR #1335, opened 2023-10-08
Use //link/inertial/density for auto-inertials

### LINKED ISSUE #1329: Use //link/inertial/density element in parser

scpeters:
# 🎉 New feature

Closes #1329. cc @jasmeet0915 

## Summary

This adds support for specifying density and `auto_inertia_params` in the `//link/inertial/density` and `//link/inertial/auto_inertia_params` elements respectively, which will apply to all collisions within a link. The following APIs are added:

* `static double Collision::DensityDefault()`: accessor for default density value
* overload of `Collision::CalculateInertial` with a `std::optional<double>` input to represent the value from `//link/inertial/density` if it has been set and an `sdf::ElementPtr` to the `<auto_inertia_params>` element or `nullptr` if it has not been set
* `double Link::Density() const`
* `void Link::SetDensity(double)`
* `sdf::ElementPtr AutoInertiaParams() const`
* `void SetAutoInertiaParams(const sdf::ElementPtr)`

The density precedence for Collisions is now:

1. Density explicitly set in `//collision/density` or by `Collision::SetDensity`.
2. Density explicitly set in `//link/inertial/density` or by `Link::SetDensity`.
3. Default density in `Collision::DensityDefault()`.

The following table summarizes this behavior:

| Density to use       | Collision density set | Collision density not set |
| -------------------- | --------------------- | ------------------------- |
| Link density set     | Collision density     | Link density              |
| Link density not set | Collision density     | Default density           |

The precedence rules for AutoInertiaParams match the rules for density.

## Test it

A test has has been added in `UNIT_Link_TEST` with the 4 combinations of link / collision density being set or not from the table above. Also the `test/sdf/inertial_stats_auto.sdf` file, which is tested with `gz sdf --inertial-stats` in `UNIT_gz_TEST`, is modified so that a `//collision/density` value is converted to `//link/inertial/density`, and the test still passes.

## Checklist
- [X] Added tests


azeey:
Per https://github.com/gazebosim/sdformat/pull/1330#issuecomment-1737879295, and https://github.com/gazebosim/sdformat/pull/1330#issuecomment-1737539126 could we pass along the  `//link/inertial/auto_inertia_params` in the new member function and avoid needing yet another overload?

scpeters:
> Per [#1330 (comment)](https://github.com/gazebosim/sdformat/pull/1330#issuecomment-1737879295), and [#1330 (comment)](https://github.com/gazebosim/sdformat/pull/1330#issuecomment-1737539126) could we pass along the `//link/inertial/auto_inertia_params` in the new member function and avoid needing yet another overload?

good call, thanks for the reminder

I've started work in https://github.com/gazebosim/sdformat/compare/scpeters/inertial_density...scpeters/inertial_density_auto_params, but I didn't write the test yet, since I think it will need a mock Geometry class in order to confirm which ElementPtr is passed to CalculateInertial.

azeey:
> > Per [#1330 (comment)](https://github.com/gazebosim/sdformat/pull/1330#issuecomment-1737879295), and [#1330 (comment)](https://github.com/gazebosim/sdformat/pull/1330#issuecomment-1737539126) could we pass along the `//link/inertial/auto_inertia_params` in the new member function and avoid needing yet another overload?
> 
> good call, thanks for the reminder
> 
> I've started work in [scpeters/inertial_density...scpeters/inertial_density_auto_params](https://github.com/gazebosim/sdformat/compare/scpeters/inertial_density...scpeters/inertial_density_auto_params), but I didn't write the test yet, since I think it will need a mock Geometry class in order to confirm which ElementPtr is passed to CalculateInertial.

Instead of a mock Geometry class, would it be easier to provide a mesh inertia calculator callback in the test that will receive the `autoInertiaParams` and check to see if it's the one from the link? i.e, https://github.com/gazebosim/sdformat/blob/11205c9059cd3991564b97d525054fcabc40faff/src/Geometry.cc#L341 will eventually call the callback provided by the test with the `autoInertiaParams` passed to `Geometry`.

scpeters:
> Instead of a mock Geometry class, would it be easier to provide a mesh inertia calculator callback in the test that will receive the `autoInertiaParams` and check to see if it's the one from the link? i.e,
> 
> https://github.com/gazebosim/sdformat/blob/11205c9059cd3991564b97d525054fcabc40faff/src/Geometry.cc#L341
> 
> will eventually call the callback provided by the test with the `autoInertiaParams` passed to `Geometry`.

I adapted the lambda callback from [Geometry_TEST.cc](https://github.com/gazebosim/sdformat/blob/11205c9059cd3991564b97d525054fcabc40faff/src/Geometry_TEST.cc#L498-L523) for an `auto_inertia_params` test in https://github.com/gazebosim/sdformat/pull/1335/commits/4eae4b8cf2a96512492a80acbbc4e4c9de3f255b and merged in the other changes from that branch

scpeters:
I fixed a compiler warning and coverage; CI is happy now

src/Collision.cc
```diff
@@ -256,23 +262,59 @@ void Collision::CalculateInertial(
...
+      std::to_string(DensityDefault()) + " kg/m^3. "
     );
     enforceConfigurablePolicyCondition(
       _config.WarningsPolicy(), densityMissingErr, _errors
     );
   }
 
+  // If this Collision's auto inertia params have not been set, then use the
+  // params passed into this function.
+  sdf::ElementPtr autoInertiaParams = this->dataPtr->autoInertiaParams;
```
azeey, line 310: We should document on sdf_tutorials that this works by completely replacing the parameters in the link, i.e, the values inside `auto_inertia_params` of the collision do not "merge" with the ones lin the parent link.

src/Collision.cc
```diff
@@ -256,23 +262,59 @@ void Collision::CalculateInertial(
...
+/////////////////////////////////////////////////
+void Collision::CalculateInertial(
+  sdf::Errors &_errors,
+  gz::math::Inertiald &_inertial,
+  const ParserConfig &_config,
+  const std::optional<double> &_density,
+  sdf::ElementPtr _autoInertiaParams)
+{
+  // Order of precedence for density:
+  double density;
```
azeey, line 278: This logic made me realize that the link could just call `SetDensity` and `SetAutoInertiaParams` on the collision object before calling `CalculateInertial` and we wouldn't need a new overload. I like this because it's more "functional", but not sure if it's worth the overload.

src/Link_TEST.cc
```diff
@@ -302,6 +316,353 @@ TEST(DOMLink, ResolveAutoInertialsWithNoCollisionsInLink)
   EXPECT_NE(nullptr, root.Element());
 }
 
+/////////////////////////////////////////////////
+TEST(DOMLink, ResolveAutoInertialsWithDifferentDensity)
+{
+  std::string sdf = "<?xml version=\"1.0\"?>"
```
azeey, line 322: nit: I think string literals would be great here like https://github.com/gazebosim/sdformat/blob/11205c9059cd3991564b97d525054fcabc40faff/src/Root_TEST.cc#L242, but if I won't block on this.

src/Collision.cc
scpeters: sure, let's document how it's implemented. I suppose we could merge the parameters in some way but we'd need to define the merge strategy first. the current approach is simplest

src/Collision.cc
scpeters: If the Collision density / `auto_inertia_params` values are unset and we call `Collision::SetDensity` / `Collision::SetAutoInertiaParams`, then those values will no longer be unset. This might or might not be important. It could matter to the `ToElement` functions, but I realized while writing this that I didn't update `Link::ToElement` or `Collision::ToElement` in this pull request.

Currently `Link::ToElement` does not export a `//inertial/@auto` attribute and explicitly exports all inertia values. `Collision::ToElement` unconditionally exports a `//collision/density` element whether the value was set or unset in the original SDFormat model.

I like the current, functional approach but am not sure what we expect from the output of the `ToElement` functions.

src/Link_TEST.cc
scpeters: https://github.com/gazebosim/sdformat/pull/1335/commits/b41b48b612db47a30409b80b4d1bb793ab734377

src/Collision.cc
azeey: > If the Collision density / auto_inertia_params values are unset and we call Collision::SetDensity / Collision::SetAutoInertiaParams, then those values will no longer be unset.

Good point. We won't have a way of knowing if those parameters were set by the user in the `<collision>` or not. So, let's go with the current approach.

>  but I realized while writing this that I didn't update Link::ToElement or Collision::ToElement in this pull request.

I didn't remember that either. In general, how do we remember to add things to `ToElement` when we add something to the spec? I wonder if we can write a test (not in this PR, of course) that calls `ToElement` and checks the output against the spec.

 > Currently Link::ToElement does not export a //inertial/@auto attribute and explicitly exports all inertia values. Collision::ToElement unconditionally exports a //collision/density element whether the value was set or unset in the original SDFormat model.

We have a general problem with values not set by the user showing up in `ToElement` (https://github.com/gazebosim/sdformat/issues/867). In most of our DOM objects, we don't keep track of whether the element/attribute was set by the user, so it's hard to figure out what to do in `ToElement`. Now that you've made `Collision::density` optional, I think we can actually call `Element::SetExplicitlySetInFile` on the element so it's printed properly according to the print config (https://github.com/gazebosim/sdformat/blob/b41b48b612db47a30409b80b4d1bb793ab734377/src/Element.cc#L592)

Regarding `//inertial/@auto`, the behavior was supposed to be determined by [ConfigureResolveAutoInertials](https://github.com/gazebosim/sdformat/blob/b41b48b612db47a30409b80b4d1bb793ab734377/include/sdf/ParserConfig.hh#L55), but I don't think it's working correctly because it's exporting all inertia values unconditionally. 

I think the logic should be the following in pseudocode

```
auto==true && SAVE_CALCULATION:
  export all inertia
auto==true && not SAVE_CALCULATION:
 export `//inertial/auto`
```

But it looks like we need more enum values for "skip calculation during load and do not save calculation" and "calculate during load but do not save calculation".

src/Collision.cc
azeey: Yes, I think the current approach is simplest and since we don't have anything currently using these, let's hold off on the merge idea. I just wanted to make sure we document the behavior since it's easy for users to think the values are being merged.

src/Collision.cc
scpeters: I've updated `Collision::ToElement` in https://github.com/gazebosim/sdformat/pull/1335/commits/8d91b6a962a90670bfafb64b41acb37e54c2040c and `Link::ToElement` in https://github.com/gazebosim/sdformat/pull/1335/commits/9460341a6eb3aa01b2ef023e0c3599bbb6e46b47

> We have a general problem with values not set by the user showing up in `ToElement` (#867). In most of our DOM objects, we don't keep track of whether the element/attribute was set by the user, so it's hard to figure out what to do in `ToElement`. Now that you've made `Collision::density` optional, I think we can actually call `Element::SetExplicitlySetInFile` on the element so it's printed properly according to the print config

I didn't use `Element::SetExplicitlySetInFile`; please advise further if it's necessary

scpeters: > Regarding `//inertial/@auto`, the behavior was supposed to be determined by [ConfigureResolveAutoInertials](https://github.com/gazebosim/sdformat/blob/b41b48b612db47a30409b80b4d1bb793ab734377/include/sdf/ParserConfig.hh#L55), but I don't think it's working correctly because it's exporting all inertia values unconditionally.
> 
> I think the logic should be the following in pseudocode
> 
> ```
> auto==true && SAVE_CALCULATION:
>   export all inertia
> auto==true && not SAVE_CALCULATION:
>  export `//inertial/auto`
> ```
> 
> But it looks like we need more enum values for "skip calculation during load and do not save calculation" and "calculate during load but do not save calculation".

Can we address this in a follow-up ticket?

azeey: > I didn't use Element::SetExplicitlySetInFile; please advise further if it's necessary

I removed `<density>` from `link_4`  in `test/sdf/inertial_stats_auto.sdf` and ran `gz sdf -p` and it didn't print the default density, so I think we're good.

> Can we address this in a follow-up ticket?

sure!

azeey (approved):

src/Collision.cc
scpeters: > > Can we address this in a follow-up ticket?
> 
> sure!

https://github.com/gazebosim/sdformat/issues/1371

PR #1338, opened 2023-10-10
Bumps in Ionic: sdformat15

azeey:
Part of https://github.com/gazebo-tooling/release-tools/issues/1027

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -6,7 +6,7 @@ jobs:
   build:
 
     env:
-      PACKAGE: sdformat14
+      PACKAGE: sdformat15
     runs-on: macos-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


iche033:
I think the homebrew build needs a sdformat15.rb formula?

scpeters:
> I think the homebrew build needs a sdformat15.rb formula?

there wasn't an sdformat15 Alias, so maybe the script failed

azeey:
https://github.com/osrf/homebrew-simulation/pull/2465

I guess the `bump_dependency.bash` script wants all `main` branches bumped first, which includes creating aliases in Homebrew.  I thought we could just run the script, which would hopefully create the formula and do the bumps in the same go. Since that doesn't seem to be working well, I suggest we merge all the bump PRs first. I'll create the aliases for Homebrew.

azeey:
I was wondering why the macOS Github Action passed. There's a hardcoded `sdformat14` in workflow file, so it was using that formula.

iche033 (approved):

PR #1346, opened 2023-11-14
Merge sdf9 ➡️  sdf12

azeey:
# ➡️  Forward port

Port `sdf9 ` ➡️  `sdf12`

Branch comparison: https://github.com/gazebosim/sdformat/compare/sdf12...sdf9

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: Ubuntu
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - 'sdf[0-9]?'
+      - 'main'
 
 jobs:
   bionic-ci:
@@ -8,7 +13,7 @@ jobs:
     name: Ubuntu Bionic CI
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Compile and test
         id: ci
         uses: ignition-tooling/action-ignition-ci@bionic
@@ -17,7 +22,7 @@ jobs:
     name: Ubuntu Focal CI
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Compile and test
         id: ci
         uses: ignition-tooling/action-ignition-ci@focal
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -10,10 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Add ticket to inbox
-        uses: technote-space/create-project-card-action@v1
+        uses: actions/add-to-project@v0.5.0
         with:
-          PROJECT: Core development
-          COLUMN: Inbox
-          GITHUB_TOKEN: ${{ secrets.TRIAGE_TOKEN }}
-          CHECK_ORG_PROJECT: true
-
+          project-url: https://github.com/orgs/gazebosim/projects/7
+          github-token: ${{ secrets.TRIAGE_TOKEN }}


iche033 (approved):

azeey:
@j-rivero There's a new cmake warning in the homebrew build, I think because `psutil` is not being installed anymore.

azeey:
I'll go ahead and merge since this isn't the cause for the homebrew warning on Jenkins.

scpeters:
> @j-rivero There's a new cmake warning in the homebrew build, I think because `psutil` is not being installed anymore.

the need to install this pip package has historically been encoded in [sdformat-default-devel-homebrew-amd64.bash](https://github.com/gazebo-tooling/release-tools/blob/37e22107c76c49e2575f799a035bb60243aa276e/jenkins-scripts/sdformat-default-devel-homebrew-amd64.bash#L7), but I believe this script is no longer being called as part of #1010

would it be better to add a `.github/ci/macos_requirements.txt` file with needed `pip` packages similar to the `.github/ci/packages.apt` file and then update the [project-default-devel-homebrew-amd64.bash](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash) script to look for that file and install those packages?

j-rivero:
> @j-rivero There's a new cmake warning in the homebrew build, I think because `psutil` is not being installed anymore.

Ouch, I think that this is a regression introduced while migrating to `gz-collections.yaml`. Should be fixed by https://github.com/gazebo-tooling/release-tools/pull/1072

azeey:
> > @j-rivero There's a new cmake warning in the homebrew build, I think because `psutil` is not being installed anymore.
> 
> Ouch, I think that this is a regression introduced while migrating to `gz-collections.yaml`. Should be fixed by [gazebo-tooling/release-tools#1072](https://github.com/gazebo-tooling/release-tools/pull/1072)

That PR seems to be removing cmake warnings from Jenkins. I thought having the warning was useful since it notified us that `psutil` was not installed. My understanding is that before the recent gz-collection changes, it used to be installed, and now it's not, which is what's causing the warnings.

j-rivero:
> My understanding is that before the recent gz-collection changes, it used to be installed, and now it's not, which is what's causing the warnings.

The gz-collection,yaml changes affects the way of generating the Jenkins jobs and how that Jenkins job configuration is being generated. It should not affect the way of provisioning the CI nodes (in this case Mac nodes for Brew) . The gz-collection.yaml migration should not impact (and hardly can do it) the software/dependencies installed.

> I thought having the warning was useful since it notified us that `psutil` was not installed.

Might have been useful in this case if you think that having `psutil` on Mac is useful. Until we can explicitly enforce using `gz-cmake` what dependencies we really require to be in a build it is hard to known which optional support we expect to find or which one we don't have interest into or it is not viable, specially on Windows/Mac.

azeey:
I have no strong opinion on whether `psutil` is useful on Mac or not. I was just pointing out that this is changing behavior from what we were doing since the dependency was explicitly included in https://github.com/gazebo-tooling/release-tools/blob/37e22107c76c49e2575f799a035bb60243aa276e/jenkins-scripts/sdformat-default-devel-homebrew-amd64.bash#L7. If we no longer need to run the `psutil` tests on Mac, that's okay with me.

j-rivero:
> I was just pointing out that this is changing behavior from what we were doing since the dependency was explicitly included in https://github.com/gazebo-tooling/release-tools/blob/37e22107c76c49e2575f799a035bb60243aa276e/jenkins-scripts/sdformat-default-devel-homebrew-amd64.bash#L7. If we no longer need to run the `psutil` tests on Mac, that's okay with me.

Understood. It is a regression since we are provisioning software in the CI scripts. I'll fix that.

scpeters:
> would it be better to add a `.github/ci/macos_requirements.txt` file with needed `pip` packages similar to the `.github/ci/packages.apt` file and then update the [project-default-devel-homebrew-amd64.bash](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash) script to look for that file and install those packages?

this is another option

Issue #1358, opened 2024-01-23
Consider [[nodiscard]] on all public C++ getters to prevent usage bugs


Ryanf55:
## Current behavior

The following code compiles with default compiler settings when building a camera plugin in Gazebo:
```c++
void MyCameraPlugin::PreUpdate(
    const UpdateInfo &_info,
    EntityComponentManager &_ecm)
{
  // All the boilerplate to get the SDF object
  Entity cameraEntity = this->impl->cameraSensorEntity;
  auto comp = _ecm.Component<components::Camera>(cameraEntity);
  if (!comp)
      return;
  sdf::Sensor &sensor = comp->Data();
  sdf::Camera *cameraSdf = sensor.CameraSensor();
  if(!cameraSdf)
      return;
 

// Now, at some point during an algorithm:
cameraSdf.ImageWidth();
 ```
 
 Calling any of the "getters" in `gs/sdformat14/sdf/Cmaera.hh` without assigning it a value would be a bug. 
 
 
 ## Desired behavior

 C++17 can protect you against this with a compiler directive called [nodiscard](https://en.cppreference.com/w/cpp/language/attributes/nodiscard)
 A great talk on the value of this compiler directive is seen here: https://youtu.be/teUA5U6eYQY?si=GluASrB8dnTNf0HM&t=1602

If, instead the header had the following, the compiler will warn if you forget to assign the return value. Thus the compiler catches a bug while you are developing code, or in CI if you have warnings treated as errors.
```c++
    public: [[nodiscard]] uint32_t ImageWidth() const;
```

## Alternatives considered

* Strict code reviews on projects that can't use C++17.

## Implementation suggestion

Adding [[nodiscard]] may introduce warnings on previously compiling code. Yes, these would all be bugs anyways,  but I would be hesitant to pull this into SDF14. I suggest considering it for the next version of SDF.

I wasn't able to find anywhere that specified the C++

mjcarroll:
I think this is a reasonable feature and can help us catch bugs sooner.

> I wasn't able to find anywhere that specified the C++ standard used for the public header of sdformat, but nodiscard is C++17 or above.

We are already using C++17 in `sdformat`: https://github.com/gazebosim/sdformat/blob/e7d7cefc8a4775c6bf0287be06da5c4c6cf9c493/CMakeLists.txt#L39-L40

[...]

PR #1360, opened 2024-01-30
Add nodiscard and example usage in Actor

Ryanf55:
# 🎉 New feature

Relates to  #1358

## Summary

* Add a macro to start using nodiscard as recommended in 1358.
* Demonstrate usage of the macro in Actor.hh
* If this PR is the right approach, then a future PR (or PR's) can fold this across to the rest of the repo. 

In theory, clang-tidy should be able to automatically apply this across the whole repo, but clang-tidy doesn't seem to work with this repo. If anyone knows how to use it, that would be much preferable because automated code refactors are much less error-prone and quicker than me doing it manually.

## Test it

* Compilation is the only thing needed
* If you want, try calling a getter in an application, and ignoring the return value. If you compile with C++17, it should warn.
* Same as above, but if you compile with C++14, you should not expect a warning.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/sdf/system_util.hh
+++ b/include/sdf/system_util.hh
@@ -36,6 +36,7 @@
 #if __cplusplus >= 201703L
 #define SDFORMAT_WARN_UNUSED [[nodiscard]]
 #else
+#error "HELLO"
 #define SDFORMAT_WARN_UNUSED
 #endif
 
--- a/include/sdf/system_util.hh
+++ b/include/sdf/system_util.hh
@@ -29,5 +29,15 @@
  */
 #define SDFORMAT_HIDDEN GZ_SDFORMAT_HIDDEN
 
+/** \def SDFORMAT_WARN_UNUSED
+ * Use to represent "[[warn_if_unused]]" compiler directive on C++17 or above
+ */
+
+#if __cplusplus >= 201703L
+#define SDFORMAT_WARN_UNUSED [[nodiscard]]
+#else
+#define SDFORMAT_WARN_UNUSED
+#endif
+
 // SDF_VISIBLE_HH_
 #endif
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,8 @@ if(COMMAND CMAKE_POLICY)
 endif(COMMAND CMAKE_POLICY)
 
 project (sdformat15 VERSION 15.0.0)
+# set(CMAKE_CXX_CLANG_TIDY clang-tidy -checks=-*,modernize-use-nodiscard)
+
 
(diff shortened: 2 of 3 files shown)


include/sdf/system_util.hh
```diff
@@ -29,5 +29,16 @@
...
+/** \def SDFORMAT_WARN_UNUSED
+ * Use to represent "[[warn_if_unused]]" compiler directive on C++17 or above
+ */
+
+#if __cplusplus >= 201703L
+#define SDFORMAT_WARN_UNUSED [[nodiscard]]
+#else
+#error "HELLO"
+#define SDFORMAT_WARN_UNUSED
+#endif
```
scpeters, line 41: per https://github.com/gazebosim/sdformat/issues/1358#issuecomment-1916778503, we don't need to define this macro, we can use `[[nodiscard]]` directly in the header file. apologies for the back-and-forth

Ryanf55: No worries. While I think I could do this to the model, this is perhaps a great candidate for using clang-tidy's automated refactoring capabilities. Although I don't have experience with it, perhaps someone at Gazebo does? 

I don't really expect to have time to do it to the entire repo, at least manually. It's not quick work because you can't find-replace or regex it.

scpeters: I haven't configured `clang-tidy` before. I think it would be worth adding `[[nodiscard]]` to this class by hand and then comparing to what clang-tidy generates

scpeters: I installed `clang-tidy` on a debian machine and built Ionic up to `sdformat15` in a colcon workspace and then was able to apply the `[[nodiscard]]` attribute automatically using the following command after sourcing the workspace:

~~~
clang-tidy --fix -checks=-*,modernize-use-nodiscard src/sdformat/include/sdf/*.hh --  $(pkg-config --cflags sdformat15)
~~~

Ryanf55: Fantastic. I'll go ahead and do the work manually and post a diff.

scpeters: untested changes to all header files in https://github.com/gazebosim/sdformat/commit/1d96a587497de7705b3cf2840b2ab2a8306d75fb

I'm pretty sure it pushes some lines over 80 characters, which would need to be fixed

azeey:
@Ryanf55, @scpeters we don't have much time left before the code freeze. Is this likely to make it by next Wednesday (08/28)?

Ryanf55:
I'm going to close because I don't have time to see this project through. Anyone else is welcome to continue!

PR #1368, opened 2024-02-02
Backport #1367 to Harmonic: Fix find Python3 logic and macOS workflow

scpeters:
# Backport

Backport #1367 to Harmonic. The patch applies cleanly to `sdf14` but will require small modifications to backport to `sdf13` since Python3 was not required in Garden.

## Original commit message:

Fix find Python3 logic and macOS workflow

* Find Python3 with find_package instead of GzPython, adapting the approach from gz-sim.
* macos workflow: use brew --prefix, not /usr/local
* macos workflow: set Python3_EXECUTABLE cmake var

**Note to maintainers**: Remember to use **Rebase-and-Merge**.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -19,11 +19,11 @@ jobs:
     # Workaround for https://github.com/actions/setup-python/issues/577
     - name: Clean up python binaries
       run: |
-        rm -f /usr/local/bin/2to3*;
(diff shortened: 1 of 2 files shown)


scpeters:
something's wrong with the python version in the macOS workflow; I'm going to try something...

azeey (approved):

PR #1370, opened 2024-02-02
Backport #1367 to Garden: Fix find Python3 logic and macOS workflow

scpeters:
# Backport

Backport #1367 to Garden. The patch required small modifications since Python3 is required in `sdf14` and later but is not required in Garden.

## Original commit message:

Fix find Python3 logic and macOS workflow

* Find Python3 with find_package instead of GzPython, adapting the approach from gz-sim.
* macos workflow: use brew --prefix, not /usr/local
* macos workflow: set Python3_EXECUTABLE cmake var

**Note to maintainers**: Remember to use **Rebase-and-Merge**.


j-rivero (approved):
good to my eyes

CMakeLists.txt
```diff
@@ -78,9 +78,31 @@ if (BUILD_SDF)
...
   set(GZ_TOOLS_VER 2)
 
+  #################################################
+  # Find python
+  if (SKIP_PYBIND11)
+    message(STATUS "SKIP_PYBIND11 set - disabling python bindings")
+    find_package(Python3 COMPONENTS Interpreter)
+  else()
+    find_package(Python3 COMPONENTS Interpreter Development
+    )
```
j-rivero, line 88: Nitpick: single line for the find_package command

CMakeLists.txt
```diff
@@ -78,9 +78,31 @@ if (BUILD_SDF)
   # available during build time
   set(GZ_TOOLS_VER 2)
 
+  #################################################
+  # Find python
+  if (SKIP_PYBIND11)
+    message(STATUS "SKIP_PYBIND11 set - disabling python bindings")
+    find_package(Python3 COMPONENTS Interpreter)
```
j-rivero, line 85: It looks like we are looking for the Interpreter unconditionally. Both cases of `SKIP_PYBIND11` look for ir. I see that is being used for `embedSdf.py`, if it is mandatory could we move it out of the if block to the main `CMakeLists.txt` course.

If that is valid, if can probably integrate this block with the `[add_directory(python)]`(//github.com/osrf/sdformat/blob/8de28a126ca5ba83b7431187e7714496e07f1168/CMakeLists.txt#L163) a bit below in this file to avoid extra checks in different locations.

CMakeLists.txt
```diff
@@ -78,9 +78,31 @@ if (BUILD_SDF)
...
+  else()
+    find_package(Python3 COMPONENTS Interpreter Development
+    )
+    if (NOT Python3_Development_FOUND)
+      GZ_BUILD_WARNING("Python development libraries are missing: Python interfaces are disabled.")
+    else()
+      set(PYBIND11_PYTHON_VERSION 3)
+      find_package(pybind11 2.4 CONFIG QUIET)
```
j-rivero, line 93: Dobule checking: do we prefer to call it with `QUITE` and add custom STATUS messages below instead of removing the `QUIET` and the messages?

j-rivero (commented):
Minor things :)

CMakeLists.txt
scpeters: Without the changes in this PR, the `include(GzPython)` line will unconditionally search for the Python interpreter, and then we would separately search for the Development component. I think that can be problematic on systems with multiple versions of python that don't all have the Development component, since it could first find an Interpreter from a non-Development version of Python, and then find a different version of python with the Development component.

So my goal here is to unconditionally search for Python3 one time with the full set of components that are needed, either `INTERPRETER` or `INTERPRETER DEVELOPMENT`. Does that make sense?

CMakeLists.txt
scpeters: 👍 I'll change it

CMakeLists.txt
scpeters: it changes the formatting of the console messages, and I think it's subjective; I'm flexible on this

CMakeLists.txt
scpeters: https://github.com/gazebosim/sdformat/pull/1370/commits/c9340301f156b6dba64c94b5adb67f6f9a3a19c8

CMakeLists.txt
j-rivero: It makes total sense yes and actually the CMake documentation recommends to do it in an atomic call. Reducing the call to just one outside of the if block and moving the block to merge the `add_directory(python)` would result in:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index fa983cee..dd7bf5d2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,29 +78,6 @@ if (BUILD_SDF)
   # available during build time
   set(GZ_TOOLS_VER 2)
 
-  #################################################
-  # Find python
-  if (SKIP_PYBIND11)
-    message(STATUS "SKIP_PYBIND11 set - disabling python bindings")
-    find_package(Python3 COMPONENTS Interpreter)
-  else()
-    find_package(Python3 COMPONENTS Interpreter Development
-    )
-    if (NOT Python3_Development_FOUND)
-      GZ_BUILD_WARNING("Python development libraries are missing: Python interfaces are disabled.")
-    else()
-      set(PYBIND11_PYTHON_VERSION 3)
-      find_package(pybind11 2.4 CONFIG QUIET)
-
-      if (pybind11_FOUND)
-        message (STATUS "Searching for pybind11 - found version ${pybind11_VERSION}.")
-      else()
-        GZ_BUILD_WARNING("pybind11 is missing: Python interfaces are disabled.")
-        message (STATUS "Searching for pybind11 - not found.")
-      endif()
-    endif()
-  endif()
-
   #################################################
   # Copied from catkin/cmake/empy.cmake
   function(find_python_module module)
@@ -159,8 +136,27 @@ if (BUILD_SDF)
   add_subdirectory(sdf)
   add_subdirectory(conf)
   add_subdirectory(doc)
-  if (pybind11_FOUND AND NOT SKIP_PYBIND11)
-    add_subdirectory(python)
+
+  #################################################
+  # Find python
+  find_package(Python3 COMPONENTS Interpreter Development)
+  if (SKIP_PYBIND11)
+    message(STATUS "SKIP_PYBIND11 set - disabling python bindings")
+  else()
+    if (NOT Python3_Development_FOUND)
+      GZ_BUILD_WARNING("Python development libraries are missing: Python interfaces are disabled.")
+    else()
+      set(PYBIND11_PYTHON_VERSION 3)
+      find_package(pybind11 2.4 CONFIG QUIET)
+
+      if (pybind11_FOUND)
+        message (STATUS "Searching for pybind11 - found version ${pybind11_VERSION}.")
+        add_subdirectory(python)
+      else()
+        GZ_BUILD_WARNING("pybind11 is missing: Python interfaces are disabled.")
+        message (STATUS "Searching for pybind11 - not found.")
+      endif()
+    endif()
   endif()
 endif(BUILD_SDF)
```

whatever you prefer, the atomic call seems to be the most critical point here.

CMakeLists.txt
j-rivero: up to you, no strong feeling either.

j-rivero (approved):
Subjective options to take or not. Changes fine.

CMakeLists.txt
scpeters: since this is a backport; I'm going to leave this as is to match the behavior in newer branches

CMakeLists.txt
scpeters: to avoid deviation from newer branches, I'm going to keep this as is

PR #1373, opened 2024-02-22
Resolve URIs relative to file path

iche033:
# 🎉 New feature

~~Marked as draft to get feedback. I'll add test and update other `resolveURI` instances if we decide this is the way to go.~~

This PR is ready for review

## Summary

Alternative approach to https://github.com/gazebosim/gz-physics/pull/599 for resolving mesh URI relative to `FilePath()`.

The following URIs are commonly used:

```xml
          <mesh>
            <uri>some_local_path/box.obj</uri>
          </mesh>
```

~~so I extended `sdf::findFile` to support searching files relative to the `model.sdf` file path~~. 
    * Update: Extended `resolveURI` to take additional search paths when resolving URIs.

To actually load the mesh, we'll also need to call `SetStoreResolvedURIs` in gz-sim.  Note, I also fixed the typo in the `SetStoreResolvedURIs` function name. We'll need to deprecate the original function in `main`. 

## Alternatives considered

The other way is to specify a findFile user callback but it is difficult to do as the user callback needs to take in a different file path (path to model.sdf) for each URI. Given that this relative path URI usage is quite common, I added support directly to `sdf::findFile` instead.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/sdf/ParserConfig.hh
+++ b/include/sdf/ParserConfig.hh
@@ -180,6 +180,10 @@ class SDFORMAT_VISIBLE ParserConfig
   public: bool URDFPreserveFixedJoint() const;
 
   /// \brief Set the storeResolvedURIs flag value.
+  /// \sa SetStoreResolvedURIs
+  public: void SetStoreResovledURIs(bool _resolveURI);
+
+  /// \brief Set the storeResolvedURIs flag value.
   /// \param[in] _resolveURI True to make the parser attempt to resolve any
   /// URIs found and store them.  False to preserve original URIs
   ///
@@ -188,12 +192,20 @@ class SDFORMAT_VISIBLE ParserConfig
   /// If the FindFileCallback provides a non-empty string, the URI will be
   /// stored in the DOM object, and the original (unresolved) URI will be
   /// stored in the underlying Element.
-  public: void SetStoreResovledURIs(bool _resolveURI);
+  public: void SetStoreResolvedURIs(bool _resolveURI);
 
   /// \brief Get the storeResolvedURIs flag value.
   /// \return True if the parser will attempt to resolve any URIs found and
   /// store them.  False to preserve original URIs
   public: bool StoreResolvedURIs() const;
+
+  /// \brief Set the path to the SDF file.
+  /// \param[in] _path Full path to SDF file.
+  public: void SetFilePath(const std::string &_path);
+
+  /// \brief Get the path to the SDF file.
+  /// \return Full path to SDF file.
+  public: const std::string &FilePath() const;
 
   /// \brief Private data pointer.
   GZ_UTILS_IMPL_PTR(dataPtr)
--- a/include/sdf/ParserConfig.hh
+++ b/include/sdf/ParserConfig.hh
@@ -199,13 +199,13 @@ class SDFORMAT_VISIBLE ParserConfig
   /// store them.  False to preserve original URIs
   public: bool StoreResolvedURIs() const;
 
-  /// \brief Set the path to the SDF file.
-  /// \param[in] _path Full path to SDF file.
-  public: void SetFilePath(const std::string &_path);
+  /// \brief Set the path to search for URIs with relative path.
+  /// \param[in] _path Path to search.
+  public: void SetRelativeURISearchPath(const std::string &_path);
 
-  /// \brief Get the path to the SDF file.
-  /// \return Full path to SDF file.
-  public: const std::string &FilePath() const;
+  /// \brief Get the path to search for URIs with relative path.
+  /// \return Path to search
+  public: const std::string &RelativeURISearchPath() const;
 
   /// \brief Private data pointer.
   GZ_UTILS_IMPL_PTR(dataPtr)
--- a/include/sdf/ParserConfig.hh
+++ b/include/sdf/ParserConfig.hh
@@ -199,14 +199,6 @@ class SDFORMAT_VISIBLE ParserConfig
   /// store them.  False to preserve original URIs
   public: bool StoreResolvedURIs() const;
 
-  /// \brief Set the path to search for URIs with relative path.
-  /// \param[in] _path Path to search.
-  public: void SetRelativeURISearchPath(const std::string &_path);
-
-  /// \brief Get the path to search for URIs with relative path.
-  /// \return Path to search
-  public: const std::string &RelativeURISearchPath() const;
-
   /// \brief Private data pointer.
   GZ_UTILS_IMPL_PTR(dataPtr)
 };
--- a/src/Heightmap.cc
+++ b/src/Heightmap.cc
@@ -137,8 +137,8 @@ Errors HeightmapTexture::Load(ElementPtr _sdf, const ParserConfig &_config)
     std::unordered_set<std::string> paths;
     if (!this->dataPtr->sdf->FilePath().empty())
     {
-      paths.insert(
-          std::filesystem::path(this->dataPtr->sdf->FilePath()).parent_path());
+      paths.insert(std::filesystem::path(
+          this->dataPtr->sdf->FilePath()).parent_path().string());
     }
     this->dataPtr->diffuse = resolveURI(
         _sdf->Get<std::string>(errors, "diffuse", this->dataPtr->diffuse).first,
@@ -155,8 +155,8 @@ Errors HeightmapTexture::Load(ElementPtr _sdf, const ParserConfig &_config)
     std::unordered_set<std::string> paths;
     if (!this->dataPtr->sdf->FilePath().empty())
     {
-      paths.insert(
-          std::filesystem::path(this->dataPtr->sdf->FilePath()).parent_path());
+      paths.insert(std::filesystem::path(
+          this->dataPtr->sdf->FilePath()).parent_path().string());
     }
     this->dataPtr->normal = resolveURI(
         _sdf->Get<std::string>(errors, "normal", this->dataPtr->normal).first,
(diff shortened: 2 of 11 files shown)


include/sdf/ParserConfig.hh
```diff
@@ -188,13 +192,21 @@ class SDFORMAT_VISIBLE ParserConfig
...
   /// store them.  False to preserve original URIs
   public: bool StoreResolvedURIs() const;
 
+  /// \brief Set the path to the SDF file.
+  /// \param[in] _path Full path to SDF file.
+  public: void SetFilePath(const std::string &_path);
+
+  /// \brief Get the path to the SDF file.
+  /// \return Full path to SDF file.
+  public: const std::string &FilePath() const;
```
scpeters, line 208: I would be more specific with this naming, like `URIRelativePath` because `FilePath` is already used in the `Element` class and many other DOM classes that are loaded directly from a file, but since `ParserConfig` represents configuration, not data loaded from a file, I would use a different name

scpeters (commented):
the concept looks good to me, and I left a few comments

include/sdf/ParserConfig.hh
iche033: I gave a shot at naming: `SetRelativeURISearchPath`. Not tied to it, happy to change to a better name.

src/SDF.cc
```diff
@@ -115,6 +116,16 @@ std::string findFile(const std::string &_filename, bool _searchLocalPath,
     {
       return path;
     }
+    else if (!_config.FilePath().empty())
+    {
+      auto p = std::filesystem::path(_config.FilePath()).parent_path();
```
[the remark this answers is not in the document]
iche033: moved path manipulation logic out of this function in [a83998c](https://github.com/gazebosim/sdformat/pull/1373/commits/a83998cd64d0055c5feeb1063edf9608fea9596f).

I didn't add the second arg though (`_stripNonFolderSuffix`) as the new naming should hopefully suggest that it expects a dir path and not a file.

src/Heightmap.cc
```diff
@@ -324,11 +325,14 @@ Errors Heightmap::Load(ElementPtr _sdf, const ParserConfig &_config)
     return errors;
   }
 
+  auto config = _config;
+  auto path = std::filesystem::path(this->dataPtr->filePath).parent_path();
+  config.SetRelativeURISearchPath(path.string());
```
scpeters, line 330: I see that `SetRelativeURISearchPath` is called here to unconditionally override the previous value in the `ParserConfig` object passed to `Load`. I think that probably makes sense, I'm trying to think it through to be sure.

Do we want to emit a warning if the value isn't empty before overwriting it? I haven't had a chance to think this through yet.

cc @azeey

src/SDF.cc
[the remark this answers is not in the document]
scpeters: 👍

include/sdf/ParserConfig.hh
scpeters: looks good

src/Heightmap.cc
```diff
@@ -324,11 +325,14 @@ Errors Heightmap::Load(ElementPtr _sdf, const ParserConfig &_config)
     return errors;
   }
 
+  auto config = _config;
+  auto path = std::filesystem::path(this->dataPtr->filePath).parent_path();
```
azeey, line 329: `filePath` may not be set if, for example, the object was constructed programmatically or the SDF input was a string. Can you guard against that?

include/sdf/ParserConfig.hh
```diff
@@ -188,13 +192,21 @@ class SDFORMAT_VISIBLE ParserConfig
...
+  public: void SetStoreResolvedURIs(bool _resolveURI);
 
   /// \brief Get the storeResolvedURIs flag value.
   /// \return True if the parser will attempt to resolve any URIs found and
   /// store them.  False to preserve original URIs
   public: bool StoreResolvedURIs() const;
 
+  /// \brief Set the path to search for URIs with relative path.
+  /// \param[in] _path Path to search.
+  public: void SetRelativeURISearchPath(const std::string &_path);
```
azeey, line 204: I'm concerned that this will be confusing for users, especially since there's a chance for the value to be overwritten. It seems to me we're adding a public API to implement something that should be hidden from the user.

I propose an alternative that does not require adding anything new to ParserConfig:
Let's extend `resolveURI` to take an optional parent directory as an argument. Then it could do the following:
* If the URI is relative:
  1. First do the normal `sdf::findFile` call
  2. If that fails, prepend the parent directory and try again. (Or we can swap these steps)
 * If the URI is absolute, just do the normal `sdf::findFile`.

src/Heightmap.cc
scpeters: I like @azeey's suggestion in https://github.com/gazebosim/sdformat/pull/1373#discussion_r1505134436

include/sdf/ParserConfig.hh
iche033: thanks for suggestions. I implemented the proposed changes in [70e897e](https://github.com/gazebosim/sdformat/pull/1373/commits/70e897e875362e4e6924674d5699ed81fcdae55a). The `resolveURI` function now takes an optional set of search paths for resolving URIs.

test/integration/resolve_uris.cc
```diff
@@ -418,3 +419,67 @@ TEST(ResolveURIs, BadCallback)
...
+  EXPECT_TRUE(std::filesystem::path(cubemapUri).is_absolute());
+  EXPECT_TRUE(std::filesystem::path(scriptUri).is_absolute());
+  EXPECT_TRUE(std::filesystem::path(meshVisualUri).is_absolute());
+  EXPECT_TRUE(std::filesystem::path(meshColUri).is_absolute());
+
+  EXPECT_TRUE(sdf::filesystem::exists(cubemapUri));
+  EXPECT_TRUE(sdf::filesystem::exists(scriptUri));
+  EXPECT_TRUE(sdf::filesystem::exists(meshVisualUri));
+  EXPECT_TRUE(sdf::filesystem::exists(meshColUri));
+}
```
azeey, line 485: I would like to see a test where the input is an SDF string, not a file, so as to exercise the situation where `sdf::Element::FilePath` returns an empty string. It's not clear to me what `parent_path` returns in that case.

azeey (changes requested):

test/integration/resolve_uris.cc
iche033: added test for loading sdf from string. In this case, `parent_path` will be empty string so it's then up to the user callback function to search for the uri file. [c130219](https://github.com/gazebosim/sdformat/pull/1373/commits/c130219fce8a874cca1e419fefa4d524df9e16b8)

src/Heightmap.cc
iche033: it should just return an empty string. I added a check nevertheless [443dfd2](https://github.com/gazebosim/sdformat/pull/1373/commits/443dfd200d7f0f341d42b10a280de58c235d39d9)

azeey:
@iche033 Looks like there's a build error on windows.

azeey (approved):

src/Utils.cc
```diff
@@ -382,11 +384,26 @@ void copyChildren(ElementPtr _sdf, tinyxml2::XMLElement *_xml,
...
   {
+    std::string sep("://");
+    if (!_searchPaths.empty() &&
+        _inputURI.find(sep) == std::string::npos &&
+        !std::filesystem::path(_inputURI).is_absolute())
+    {
+      for (const auto &sp : _searchPaths)
+      {
+        std::string fullPath = sdf::filesystem::append(sp, _inputURI);
+        resolvedURI = sdf::findFile(fullPath, true, false);
```
scpeters, line 401: I think we could add a comment here to clarify the `true, false` arguments: "Search local paths but don't use callbacks"

src/Utils.cc
```diff
@@ -382,11 +384,26 @@ void copyChildren(ElementPtr _sdf, tinyxml2::XMLElement *_xml,
...
+        if (!resolvedURI.empty())
+          return resolvedURI;
+      }
+    }
+
     resolvedURI = sdf::findFile(_inputURI, true, true, _config);
```
scpeters, line 407: we could add a comment here about the `true, true` arguments: "Search local paths and use registered callbacks"

test/integration/resolve_uris.cc
```diff
@@ -418,3 +419,151 @@ TEST(ResolveURIs, BadCallback)
...
+
+/////////////////////////////////////////////////
+TEST(ResolveURIs, ResolvedRelativeURIs)
+{
+  const std::string testFile =
+    sdf::testing::TestFile("sdf", "resolve_relative_uris.sdf");
+  const std::string sdfDir = sdf::testing::TestFile("sdf");
+
+  sdf::ParserConfig config;
+  config.SetStoreResovledURIs(true);
```
scpeters, line 431:
```suggestion
  config.SetStoreResolvedURIs(true);
```

test/integration/resolve_uris.cc
```diff
@@ -418,3 +419,151 @@ TEST(ResolveURIs, BadCallback)
...
+              <name>Gazebo/Grey</name>
+            </script>
+          </material>
+        </visual>
+      </link>
+    </model>
+  </sdf>)";
+
+  sdf::ParserConfig config;
+  config.SetStoreResovledURIs(true);
```
scpeters, line 520:
```suggestion
  config.SetStoreResolvedURIs(true);
```

scpeters (approved):
just some minor suggestions; looks good to me

test/integration/resolve_uris.cc
iche033: [83af965](https://github.com/gazebosim/sdformat/pull/1373/commits/83af965d2375ffe3438385e5b8e73a410db6a359)

src/Utils.cc
iche033: added comment. [83af965](https://github.com/gazebosim/sdformat/pull/1373/commits/83af965d2375ffe3438385e5b8e73a410db6a359)

src/Utils.cc
iche033: added comment. [83af965](https://github.com/gazebosim/sdformat/pull/1373/commits/83af965d2375ffe3438385e5b8e73a410db6a359)

scpeters (approved):

Issue #632, opened 2021-07-23
model_state and model alias (overwrite each other) in generated schema files


FirefoxMetzger:
The file `world.sdf` includes both `model.sdf` and `state.sdf`

https://github.com/ignitionrobotics/sdformat/blob/722b92c92e2b19c408f3d4846ea7fcb5dfa7ed58/sdf/1.8/world.sdf#L62-L68

`state.sdf` in turn includes `model_state.sdf`

https://github.com/ignitionrobotics/sdformat/blob/722b92c92e2b19c408f3d4846ea7fcb5dfa7ed58/sdf/1.8/state.sdf#L37

Both `model_state.sdf` and `model.sdf` define a `<model>` tag:

https://github.com/ignitionrobotics/sdformat/blob/722b92c92e2b19c408f3d4846ea7fcb5dfa7ed58/sdf/1.8/model.sdf#L2

https://github.com/ignitionrobotics/sdformat/blob/722b92c92e2b19c408f3d4846ea7fcb5dfa7ed58/sdf/1.8/model_state.sdf#L2

which in itself shouldn't be a problem, because they exist in different scopes, one inside `<world>` and the other inside `<state>`.

This include chain is reflected in the generated .xsd files; however, the includes are all placed at the top of their respective document, which makes both `<xsd:element name='model'>` tags appear on the same level inside `world.xsd`, i.e., the nesting of model_state's model is not preserved. This - naturally - leads to a conflict and in the current version model_state's `<model>` overwrites model's `<model>`, which is (very) undesirable. 

I'm not 100% sure how to fix this, since I am not familiar enough with the XMLSchema spec, but I assume the solution is something along the lines of renaming state tags, introducing namespaces, or ensuring that model_state's model is included inside the state element (preferred, but not sure if feasible).

FirefoxMetzger:
This appears to affect all elements that have an accompanying `*_state.xsd`, so model, light, and link,

FirefoxMetzger:
This also affects `<pose>` defined in `/sdf/v1.X/schema/types.xsd` and `<pose>` defined in `/sdf/v1.X/pose.sdf`. This is a slightly different path than the `X_state.sdf` vs `X.sdf` files, but appears to have the same root cause and a similar result.

In this case (`<pose>`) I guess the solution is to remove the entry in `types.xsd`, since `pose.sdf` is meant to supersede it?

azeey:
[...]

PR #1377, opened 2024-03-01
sdf 1.12: rename state subelements with _state suffix

scpeters:
# 🎉 New feature

Part of https://github.com/gazebosim/sdformat/issues/632

## Summary

There is some overlap in the names used for elements in the `//world/state` specification and elsewhere, for example the [model_state.sdf](https://github.com/gazebosim/sdformat/blob/sdformat14_14.0.0/sdf/1.11/model_state.sdf#L2) file [included in state.sdf](https://github.com/gazebosim/sdformat/blob/sdformat14_14.0.0/sdf/1.11/state.sdf#L37) defines an element named [model](https://github.com/gazebosim/sdformat/blob/sdformat14_14.0.0/sdf/1.11/model_state.sdf#L2). This has caused challenges with aliasing in XSD (see https://github.com/gazebosim/sdformat/issues/632 and https://github.com/gazebosim/sdformat/pull/643), and it would be simpler if some of these elements had different names (for example by appending a `_state` suffix). This pull request is a work-in-progress to append this suffix to the relevant state tags in SDFormat 1.12 with automatic conversion of state elements defined in earlier versions of the spec. I have the conversion partially working, but have run into an issue with the `Converter` class, which does not act on all sibling elements, but only to the first child element for a given `<move>` conversion.

## Checklist
- [X] Added tests

### DIFF
--- a/sdf/1.12/1_11.convert
+++ b/sdf/1.12/1_11.convert
@@ -1,2 +1,46 @@
 <convert name="sdf">
+
+  <convert name="world">
+    <!-- Rename state elements -->
+    <convert name="state">
+
+      <move>
+        <from element="light"/>
+        <to element="light_state"/>
+      </move>
+
+      <move>
+        <from element="model"/>
+        <to element="model_state"/>
+      </move>
+
+      <convert name="model">
+        <move>
+          <from element="link"/>
+          <to element="link_state"/>
+        </move>
+        <move>
+          <from element="model"/>
+          <to element="model_state"/>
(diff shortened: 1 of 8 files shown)


scpeters:
as an alternative to the partial support for backwards compatibility in this pull request, I've opened an alternative pull request in https://github.com/gazebosim/sdformat/pull/1455 that renames the state subelements as a breaking change

scpeters:
see #1455 instead

PR #1393, opened 2024-04-05
Schema accessor

### LINKED ISSUE #1376: Add schema file name accessor to each DOM type C++ class
## Desired behavior

Getting the schema file for each SDF DOM type element should be possible. 
This should make initializing an element for loading standalone DOM types easier.

Without this format. I am doing something like this in my code to initialize a DOM type element so that I can get type checking 
```
// Schema file type
template <typename T>
struct SdfTypeToSchemaFile {
  static const std::string kSchemaFile;
};

template <>
const std::string SdfTypeToSchemaFile<sdf::Surface>::kSchemaFile = "surface.sdf";
// Other definitions with template specialization.

template <typename T>
sdf::ElementPtr init_element_for_type(){
  auto element = std::make_shared<::sdf::Element>();
  sdf::initFile(SdfTypeToSchemaFile<T>::kSchemaFile, element);
  return element;
}
```

Ideally, I hope to do something like this for simplicity.
```
template <typename T>
sdf::ElementPtr init_element_for_type(){
  auto element = std::make_shared<::sdf::Element>();
  sdf::initFile(T::SchemaFile(), element);
  return element;
}
```

## Implementation suggestion

Implement a static function for each DOM type. For example for `Surface`:
```
// sdformat/include/sdf/Surface.hh
class Surface{
  // Omitted details
  public: static const std::string& SchemaFile();
};

// sdformat/src/Surface.cc

static const std::string& Surface::SchemaFile(){
  static const std::string* kSchemaFile = new std::string("surface.sdf");
  return *kSchemaFile;
}

```

## Additional context

Currently the schema file is hard coded in it's `T::ToElement()` for example:
https://github.com/gazebosim/sdformat/blob/c4bfe35bdaca8e667089bf9adf99b7930192ccaf/src/Surface.cc#L412

aagrawal05:
# 🎉 New feature

Closes #1376

## Summary
1. I've created a script format_sdf.py which reads from the /sdf/1.11 folder and generates/updates the source for all the relevant classes in the src/ and include/ directories as per the suggested implementation.
2. There were DOM elements navsat.sdf and forcetorque.sdf for which the script does not work and I had to manually edit the script because their naming convention did not match the camel-case conversion (they would need to be nav_sat.sdf and force_torque.sdf respectively to function correctly with the script) This may be undesirable from a code style perspective but may be a breaking change—this is a minor point but I just wanted to point it out.
3. The code builds successfully, I have not thoroughly tested it but it does seem to work.
4. I can add tests for the SchemaFile in all the DOM element tests but I thought it might be overkill because the names are hardcoded in the method anyways. I can add it easily with the script if needed.

## Test it
Tests can be done manually as mentioned by the OP in #1376 in his code example.
I could also add tests in the DOM element unit tests, but that might be overkill because the values are hard-coded.

## Checklist
- [x] Added tests


azeey (changes requested):
Thanks for working on this! I have a few comments that should be easy to address.

Regarding testing, I don't think we should create new tests for this, but one way we can make sure it's working is by using `SchemaFile()` instead of hard coded  file names in `ToElement` functions. Perhaps you can update your script to make that change as well?

azeey:
@aagrawal05 Looks like there are some style issues and a DCO error. Mind fixing them?

aagrawal05:
> @aagrawal05 Looks like there are some style issues and a DCO error. Mind fixing them?

Hey @azeey sure let me take a look. The style issues look fairly straightforward and I'll shortly submit changes to fix this. Please let me know if anything further is to be done in terms of functionality.

aagrawal05:
Hey @azeey in attempting to resolve the DCO error through a rebase, it appears that I have accidentally re-committed all the 23 commits from my first changes before the merge to my latest changes... 

I believe it may be a bit complicated for me to proceed accurately to revert this from this point. I feel it might be easier to just close PR and create a new PR on a new branch with a single commit (this also has the added benefit of being merged with the latest changes in sdf14).

With regards to the style I believe the errors should be resolved but I need to test with the linter using the workflow.

Please let me know how to proceed.

azeey:
@aagrawal05 yes, I think I new PR would be cleaner. It would be ideal if you can target `sdf15` as well.

azeey:
Hi @aagrawal05, any updates on this PR?

aagrawal05:
Hey Azeey sure let me just complete go ahead and open the new PR

PR #1402, opened 2024-04-16
Fix trivial warning on 24.04 for JointAxis_TEST.cc

j-rivero:
# 🦟 Bug fix

## Summary
A build of the `sdf14` on the upcoming Ubuntu Noble return 3 warnings on the `JointAxis_TEST.cc`:

```
‘multiplier’ may be used uninitialized
``` 

Adding a default value when constructing seems enough.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/JointAxis_TEST.cc
+++ b/src/JointAxis_TEST.cc
@@ -296,21 +296,21 @@ TEST(DOMJointAxis, ToElement)
   ASSERT_TRUE(errors.empty());
   EXPECT_EQ("axis2", mimicAxisName);
 
-  double multiplier = 0.0;
+  double multiplier;
   multiplier = mimicElem->Get<double>(
-      errors, "multiplier", multiplier).first;
+      errors, "multiplier", 0.0).first;
   ASSERT_TRUE(errors.empty());
   EXPECT_DOUBLE_EQ(5.0, multiplier);
 
-  double offset = 0.0;
+  double offset;
   offset = mimicElem->Get<double>(
-      errors, "offset", offset).first;
+      errors, "offset", 0.0).first;
   ASSERT_TRUE(errors.empty());
   EXPECT_DOUBLE_EQ(1.0, offset);
 
-  double reference = 0.0;
+  double reference;
   reference = mimicElem->Get<double>(
-      errors, "reference", reference).first;
+      errors, "reference", 0.0).first;
   ASSERT_TRUE(errors.empty());
   EXPECT_DOUBLE_EQ(2.0, reference);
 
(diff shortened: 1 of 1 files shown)


src/JointAxis_TEST.cc
```diff
@@ -296,19 +296,19 @@ TEST(DOMJointAxis, ToElement)
   ASSERT_TRUE(errors.empty());
   EXPECT_EQ("axis2", mimicAxisName);
 
-  double multiplier;
+  double multiplier = 0.0;
   multiplier = mimicElem->Get<double>(
       errors, "multiplier", multiplier).first;
```
scpeters, line 301:
```suggestion
  double multiplier;
  multiplier = mimicElem->Get<double>(
      errors, "multiplier", 0.0).first;
```

This third argument is the default value; I think it's little clearer to put the constant here instead of initializing `multiplier`.

src/JointAxis_TEST.cc
```diff
@@ -296,19 +296,19 @@ TEST(DOMJointAxis, ToElement)
...
   ASSERT_TRUE(errors.empty());
   EXPECT_DOUBLE_EQ(5.0, multiplier);
 
-  double offset;
+  double offset = 0.0;
   offset = mimicElem->Get<double>(
       errors, "offset", offset).first;
```
scpeters, line 307:
```suggestion
  double offset;
  offset = mimicElem->Get<double>(
      errors, "offset", 0.0).first;
```

src/JointAxis_TEST.cc
```diff
@@ -296,19 +296,19 @@ TEST(DOMJointAxis, ToElement)
...
   ASSERT_TRUE(errors.empty());
   EXPECT_DOUBLE_EQ(1.0, offset);
 
-  double reference;
+  double reference = 0.0;
   reference = mimicElem->Get<double>(
       errors, "reference", reference).first;
```
scpeters, line 313:
```suggestion
  double reference;
  reference = mimicElem->Get<double>(
      errors, "reference", 0.0).first;
```

src/JointAxis_TEST.cc
scpeters: looking at the rest of the file, I see that your proposed change matches the pattern in the rest of the file, so you can ignore this suggestion if you want to maintain consistency, and I'll fix the rest of the file in a separate pull request

scpeters (approved):

src/JointAxis_TEST.cc
j-rivero: > looking at the rest of the file, I see that your proposed change matches the pattern in the rest of the file, so you can ignore this suggestion if you want to maintain consistency, and I'll fix the rest of the file in a separate pull request

Looks good to me to have the changes under this PR [03cdb10](https://github.com/gazebosim/sdformat/pull/1402/commits/03cdb10fc857405dd83e61dbb462033fce336f0e) . Let me know if I'm missing any.

scpeters (approved):

PR #1408, opened 2024-05-06
Enable 24.04 CI, remove distutils dependency

scpeters:
# 🎉 New feature

Attempting to enable CI on 24.04

## Summary

The `python3-distutils` package is not available on 24.04, but we only use it in a codepath for old versions of cmake. Since we are already requiring cmake 3.22.1 in gz-cmake4, we can require that version in sdformat15 as well and remove code for old versions of cmake.

## Test it

Observe CI results

## Checklist
- [X] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #1413, opened 2024-05-13
(Backport) Enable 24.04 CI, remove distutils dependency (#1408)

Blast545:
* distutils is no longer required since this branch requires a new enough version of cmake.
* Remove outdated cmake code: setting python debug libraries is not needed as of cmake 3.12.
* Require cmake 3.22.1

# 🦟 Bug fix

Should fix CI on sdformat for Harmonic.
See reference job: https://build.osrfoundation.org/job/sdformat-ci-sdf14-noble-amd64/

## Summary
Backporting the removal of the distutils dependency

## Checklist
- [ ] Added tests

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
 
 find_package(sdformat14 REQUIRED)
 
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.22.1 FATAL_ERROR)
 
 find_package(sdformat14 REQUIRED)
 
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.22.1 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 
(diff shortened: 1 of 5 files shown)


Blast545:
@scpeters ABI checker is failing on the CMake version :/

scpeters:
> @scpeters ABI checker is failing on the CMake version :/

I think this is a problem with infra and not this PR; let's try to fix the infra

examples/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
```
scpeters, line 1: @azeey do you think this should be reverted as well?

python/CMakeLists.txt
```diff
@@ -1,22 +1,6 @@
...
-  # pybind11 logic for setting up a debug build when both a debug and release
-  # python interpreter are present in the system seems to be pretty much broken.
-  # This works around the issue.
-  set(PYTHON_LIBRARIES "${PYTHON_DEBUG_LIBRARIES}")
-endif()
-
-
 if(USE_SYSTEM_PATHS_FOR_PYTHON_INSTALLATION)
-  if(${CMAKE_VERSION} VERSION_LESS "3.12.0")
-    execute_process(
```
azeey, line 11: I think we should revert this. We can be sure Jammy has CMake > 3.12, but since the whole project sets the minimum version to 3.10, we can't assume that this is being built by a newer version of CMake.

examples/CMakeLists.txt
azeey: I think so for the same reason as https://github.com/gazebosim/sdformat/pull/1413#discussion_r1598975231

python/CMakeLists.txt
Blast545: All changes related to CMakeLists were removed in 47e82ba

scpeters (approved):

PR #1423, opened 2024-05-23
Update default camera instrinsics skew to 0, which matches spec

shameekganguly:
# 🦟 Bug fix

## Summary
Default camera lens intrinsics skew is 0 in the camera sdf spec: [https://github.com/gazebosim/sdformat/blob/sdf14/sdf/1.11/camera.sdf](https://github.com/gazebosim/sdformat/blob/c9779cb7f19bac92b512e213b7fd50dbcf2c8c8d/sdf/1.11/camera.sdf#L193)
But the default value in `sdf::Camera` class was set to 1. I've fixed this value to match the spec in this PR and updated the associated tests.

### DIFF
--- a/python/test/pyCamera_TEST.py
+++ b/python/test/pyCamera_TEST.py
@@ -191,7 +191,7 @@ class CameraTEST(unittest.TestCase):
       cam.set_lens_intrinsics_cy(123)
       self.assertAlmostEqual(123, cam.lens_intrinsics_cy())
 
-      self.assertAlmostEqual(1.0, cam.lens_intrinsics_skew())
+      self.assertAlmostEqual(0.0, cam.lens_intrinsics_skew())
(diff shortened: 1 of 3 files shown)


ahcorde (approved):

iche033 (approved):

azeey:
Is this change worth announcing on discourse? It could be a breaking change for some folks, but I don't have a sense of how many people will be affected.

iche033:
I think it's a bugfix that we can mention the change in Migration guide?  I imagine most people will be setting skew to 0.

azeey:
Okay. Adding to the migration guide sounds good. @shameekganguly/@iche033 mind doing that now so we don't forget next time we make a release?

iche033:
yep I'll take care of it

PR #1431, opened 2024-06-06
Deprecate camera pose

iche033:
# 🦟 Bug fix

## Summary

Deprecate `//sensor/camera/pose` as it's redundant with `//sensor/pose`, see https://github.com/gazebosim/gz-sim/pull/2433#issuecomment-2150894010.

I deprecated the relevant functions in Camera DOM class. As for deprecating the actual `//sensor/camera/pose` SDF element, I noticed that the pose element is included in the `sdf/1.12/camera.sdf` file using `<include filename="pose.sdf" required="0"/>` so I can't update it's description to say that it's being deprecated. One option is to copy the entire conent of pose.sdf into camera.sdf and update its description - let me know if this is the preferred method. **Update**: marked as deprecated by setting `required` to `-1` (`<include filename="pose.sdf" required="-1"/>`) 

## Checklist
- [ ] Added tests

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -30,6 +30,17 @@ but with improved human-readability..
       matches `"0"`, `"1"`, `"true"`, or `"false"` and returns `false`
       otherwise.
 
+### Deprecations
+
+- **sdf/Camera.hh**:
+   + The `//sensor/camera/pose` SDF element and corresponding pose functions
+     in the Camera DOM class are deprecated. Please specify camera pose using
+     the `//sensor/pose` SDF element instead.
+   + ***Deprecation:*** const gz::math::Pose3d &RawPose() const
+   + ***Deprecation:*** void SetRawPose(const gz::math::Pose3d &_pose)
+   + ***Deprecation:*** const std::string &PoseRelativeTo() const
+   + ***Deprecation:*** void SetPoseRelativeTo(const std::string &_frame)
+
 ## libsdformat 13.x to 14.x
 
 ### Additions
--- a/include/sdf/Camera.hh
+++ b/include/sdf/Camera.hh
@@ -334,27 +334,27 @@ namespace sdf
     /// \param[in] _center Distortion center or principal point.
     public: void SetDistortionCenter(const gz::math::Vector2d &_center);
 
-    /// \brief Get the pose of the camer. This is the pose of the camera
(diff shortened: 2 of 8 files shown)


scpeters:
> As for deprecating the actual `//sensor/camera/pose` SDF element, I noticed that the pose element is included in the `sdf/1.12/camera.sdf` file using `<include filename="pose.sdf" required="0"/>`

does it work to set `required="-1"` in the include element?

~~~
<include filename="pose.sdf" required="-1"/>
~~~

iche033:
> does it work to set required="-1" in the include element?

oh I didn't know about this! Just found what -1 means [in the parser code](https://github.com/gazebosim/sdformat/blob/fc84f94d147bf31fd594e17bade68946246236b3/src/parser.cc#L1640-L1641). I think that's what I need, thanks. Set `required` to `-1` in [c8f2521](https://github.com/gazebosim/sdformat/pull/1431/commits/c8f25218b242cd5ca3ef1c765a9706880776f2e0)

scpeters:
I looked back in gazebo-classic, and we used the `//camera/pose` for multicameras (such as a stereo camera) in order to specify the location of each sub-camera:

* https://github.com/gazebosim/gazebo-classic/blob/gazebo11/test/worlds/multicamera_test.world#L63-L90

do we have support for stereo-cameras yet in gz-sim? cc @azeey

iche033:
> do we have support for stereo-cameras yet in gz-sim?

Gazebo-classic uses a `multicamera` sensor so that it can guarantee the cameras are synchronized / updated in the same timestep. This is not necessary in gz-sim. To use stereo or multi-cameras in gz-sim, the users just need to create 2 (or more) cameras with the same update rate - this will guarantee that the cameras are updated in the same timestep.

iche033:
I originally thought `//camera/pose` is used for rotating the camera to optical frame but no you're right it's for `multicamera` sensors. If there are other consumers of sdformat (other than gazebo-classic and gz-sim) that support `multicamera` + `//camera/pose`, maybe we should not deprecate this. We just won't support this field in gz-sim.

iche033:
Closing this. The new Gazebo will not use this SDF element but there maybe users out there who do need `multicamera` + `//camera/pose` .

PR #1468, opened 2024-08-19
Decouple linking to shlwapi from BUILD_TESTING

scpeters:
# 🦟 Bug fix

Fixing test failures discussed in https://github.com/gazebosim/sdformat/pull/1414#issuecomment-2297021793.

## Summary

The windows builds were broken with `-DBUILD_TESTING=OFF` (see https://github.com/gazebosim/sdformat/pull/1414#issuecomment-2297021793), so move the `target_link_libraries` call outside the `if (BUILD_TESTING)` logical block.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -49,6 +49,11 @@ target_link_libraries(${PROJECT_LIBRARY_TARGET_NAME}
   PRIVATE
     TINYXML2::TINYXML2)
 
+if (WIN32)
+  target_link_libraries(${PROJECT_LIBRARY_TARGET_NAME}
+    PRIVATE shlwapi)
+endif()
+
   if (USE_INTERNAL_URDF)
     target_include_directories(${PROJECT_LIBRARY_TARGET_NAME} PRIVATE
       ${CMAKE_CURRENT_SOURCE_DIR}/urdf)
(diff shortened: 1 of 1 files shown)


scpeters:
testing with gz-sensors8 using a custom gazebodistro branch (see https://github.com/gazebo-tooling/gazebodistro/commit/e76248c03a75dfc412344b18484ee7cdcdd572ca)

* [Build Status](https://build.osrfoundation.org/view/gz-harmonic/job/gz_sensors-8-win/74/) https://build.osrfoundation.org/view/gz-harmonic/job/gz_sensors-8-win/74/

src/CMakeLists.txt
```diff
@@ -49,6 +49,11 @@ target_link_libraries(${PROJECT_LIBRARY_TARGET_NAME}
   PRIVATE
     TINYXML2::TINYXML2)
 
+if (WIN32)
+  target_link_libraries(${PROJECT_LIBRARY_TARGET_NAME}
```
azeey, line 53: Do we want to link against this library only if `GZ_ENABLE_RELOCATABLE_INSTALL` is enabled?

traversaro:
Thanks for catching this error, my bad!

src/CMakeLists.txt
scpeters: I believe `InstallationDirectories.cc` is compiled unconditionally, so I think we can't avoid linking it.

an alternative suggestion from @iche033 in https://github.com/gazebosim/sdformat/pull/1414#issuecomment-2297148801 is to avoid using `shlwapi` by copying the `joinPaths` implementation from `gz-common` and use `std::filesystem`

azeey: Yeah, that sounds like a good idea.

azeey (approved):
I'll approve this so we can merge and get CI back to green, but changing the implementation so we don't need `shlwapi` sounds good.

src/CMakeLists.txt
iche033: updated implementation in #1469

PR #1474, opened 2024-08-27
Fix symbol checking test when compiled with debug symbols

j-rivero:
# 🦟 Bug fix

## Summary

The test that checks for prefixed binary symbols was broken when compiled with DebWithRelInfo since it was checking debugging symbols that broke the heuristics used.

The commit fixes it doing a couple of actions:
  *  Include the length of the namespace sdf according to the mangling rules: `3sdf`
  *  Check only dynamic symbols being exported

## Checklist
- [ ] Added tests

### DIFF
--- a/test/integration/all_symbols_have_version.bash.in
+++ b/test/integration/all_symbols_have_version.bash.in
@@ -7,7 +7,7 @@ VERSIONED_NS=v@PROJECT_VERSION_MAJOR@
 
 # nm options:
 # -D to get only dynamic symbols exported
-# 3 before the sdf is <lenght, id> used by
+# 3 before the sdf is <length, id> used by
 # mangled symbols in C++ to check for the
 # sdf namespace
 NUM_SYMBOLS=$(nm -D $LIBPATH | grep -e "3sdf" | wc -l)
--- a/test/integration/all_symbols_have_version.bash.in
+++ b/test/integration/all_symbols_have_version.bash.in
@@ -4,7 +4,13 @@ LIBPATH=$1
 VERSIONED_NS=v@PROJECT_VERSION_MAJOR@
 
 # Sanity check - there should be at least one symbol
-NUM_SYMBOLS=$(nm $LIBPATH | grep -e "sdf" | wc -l)
+
+# nm options:
+# -D to get only dynamic symbols exported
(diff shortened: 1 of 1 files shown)


j-rivero:
Tested (pending): [Build Status](https://build.osrfoundation.org/job/sdformat15-debbuilder/529/)

scpeters:
> Tested (pending): [Build Status](https://build.osrfoundation.org/job/sdformat15-debbuilder/529/)

trying again with different branch parameters

* [Build Status](https://build.osrfoundation.org/job/sdformat15-debbuilder/530/) https://build.osrfoundation.org/job/sdformat15-debbuilder/530/

azeey:
I was also working on this and found that adding the `--extern-only` flag to `nm` fixes the issue. From what I can understand, the issue is that debbuilders enable `LTO` which causes some debug symbols to be added to the shared library. These symbols contain the file names, one of which is :

```
0000000000f9c8ce N sdformat15_get_install_prefix_impl.cc.b80149f8
```

This causes a false negative on the versioned symbol test.

I tested by adding `set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)` in the root `CMakeLists.txt` file and was able to reproduce the test failure locally.

In general, it would be good to enable LTO on our libraries so these types of issues are caught at the PR stage. I think this is the third time this LTO stuff has bit us.

scpeters:
> In general, it would be good to enable LTO on our libraries so these types of issues are caught at the PR stage. I think this is the third time this LTO stuff has bit us.

should we do this in the [action-gz-ci](https://github.com/gazebo-tooling/action-gz-ci) workflow?

azeey:
> > In general, it would be good to enable LTO on our libraries so these types of issues are caught at the PR stage. I think this is the third time this LTO stuff has bit us.
> 
> should we do this in the [action-gz-ci](https://github.com/gazebo-tooling/action-gz-ci) workflow?

I was thinking of enabling it via CMake, maybe in gz-cmake or on a per-project basis. That way local development will also use the same flags.

azeey (approved):
Tested by adding a function outside of the inline namespace and verified that it failed.

scpeters:
@Mergifyio backport sdf15 sdf14

PR #1489, opened 2024-10-14
Updating ruby script to use standalone executable

sauk2:
# 🎉 New feature

Related to https://github.com/gazebosim/gz-tools/issues/7

## Summary
Updated the ruby script to use the standalone executable. 

I also tried to resolve the issue of building the package because of the `FrameSemantics` class. You can check the error in [ci of your branch](https://github.com/gazebosim/sdformat/actions/runs/11167373622/job/31043453550?pr=1465).

I updated the CMakeLists and am not sure if the following is the right approach ...

```
  add_library(gz STATIC gz.cc ../FrameSemantics.cc)
``` 

## Test it
Test any of the SDF commands
```
  gz sdf --help
```

## Checklist
- [ ] Added tests

### DIFF
--- a/src/cmd/CMakeLists.txt
+++ b/src/cmd/CMakeLists.txt
@@ -7,7 +7,7 @@ if (NOT HAVE_GZ_TOOLS)
 endif()
 
 # Make a small static lib of command line functions
-add_library(gz STATIC gz.cc)
+add_library(gz STATIC gz.cc ../FrameSemantics.cc)
 target_link_libraries(gz
   ${PROJECT_LIBRARY_TARGET_NAME}
 )
@@ -47,14 +47,14 @@ endif()
 # Generate the ruby script for internal testing.
 # Note that the major version of the library is included in the name.
 # Ex: cmdsdformat0.rb
-set(cmd_script_generated_test 
+set(cmd_script_generated_test
   "${CMAKE_BINARY_DIR}/test/lib/$<CONFIG>/ruby/gz/cmd${PROJECT_NAME}.rb")
-set(cmd_script_configured_test 
+set(cmd_script_configured_test
   "${CMAKE_CURRENT_BINARY_DIR}/test_cmd${PROJECT_NAME}.rb.configured")
 
 # Set the library_location variable to the full path of the library file within
 # the build directory.
-set(library_location "$<TARGET_FILE:${PROJECT_NAME}>")
+set(library_location "$<TARGET_FILE:${sdf_executable}>")
 
 configure_file(
   "cmd${PROJECT_NAME_NO_VERSION_LOWER}.rb.in"
@@ -82,7 +82,7 @@ else()
   set(library_location_prefix "${CMAKE_INSTALL_LIBDIR}")
 endif()
 
-set(library_location "../../../${library_location_prefix}/$<TARGET_FILE_NAME:${PROJECT_NAME}>")
+set(library_location "../../../${CMAKE_INSTALL_LIBEXECDIR}/gz/${GZ_DESIGNATION}${PROJECT_VERSION_MAJOR}/$<TARGET_FILE_NAME:${sdf_executable}>")
 
 configure_file(
   "cmd${PROJECT_NAME_NO_VERSION_LOWER}.rb.in"
--- a/src/cmd/cmdsdformat.rb.in
+++ b/src/cmd/cmdsdformat.rb.in
@@ -14,263 +14,39 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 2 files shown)


azeey:
@sauk2 have you had a chance to push this further?

sauk2:
> @sauk2 have you had a chance to push this further?

@azeey Do take a look, I have made some changes to the tests.

src/gz_TEST.cc
```diff
@@ -40,11 +40,6 @@
 // DETAIL_GZ_CONFIG_PATH is compiler definition set in CMake.
 #define GZ_CONFIG_PATH DETAIL_GZ_CONFIG_PATH
 
-static std::string SdfVersion()
-{
-  return " --force-version " + std::string(SDF_VERSION_FULL);
-}
-
```
azeey, line 47: Any reason for removing this? The `--foce-version` argument is handled by `gz-tools` (see https://github.com/gazebosim/gz-tools/blob/2b228e5b956f1e966053dd860374670573580b41/src/gz.in#L244). We need to make sure it still works with the change to using executables.

azeey (changes requested):

src/gz_TEST.cc
sauk2: Done. I added it back and the tests should pass now!

src/gz_TEST.cc
```diff
@@ -1118,7 +1119,7 @@ TEST(print_rotations_in_degrees, SDF)
 
   // Printing with in_degrees
   output = custom_exec_str(
-      GzCommand() + " sdf -p --precision 6 " + path + " --degrees "
+      GzCommand() + " sdf -p --precision 6 " + path + " --degrees 1"
```
azeey, line 1122: Is it not possible to make `--degrees` anot take an extra argument? I'd rather not change the usage if possible.

Edit: see my comment in `sdf_main.cc`.

src/gz_TEST.cc
```diff
@@ -1066,7 +1066,8 @@ TEST(describe, SDF)
 {
   // Get the description
   std::string output =
-    custom_exec_str(GzCommand() + " sdf -d " + SdfVersion());
+    custom_exec_str(GzCommand() + " sdf -d " +
+    SDF_PROTOCOL_VERSION + SdfVersion());
```
azeey, line 1070: Can you say why this change is necessary?

src/gz_TEST.cc
azeey, line 1: I'd generally expect minimal changes in this file since the hope is to keep the usage of the CLI tool exactly the same.

src/gz_TEST.cc
sauk2: The `--describe` option requires an SDF spec argument. 

Based on your suggestions I reverted the test and instead provided a default value in [`sdf_main.cc`](https://github.com/gazebosim/sdformat/pull/1489/files#diff-cbb31fe4179fd72baff8f3fdf0e1e95b06b692506fc78170593e2bdecb63da7fR150-R151)

src/gz_TEST.cc
sauk2: Thanks for the feedback! I have reverted the tests and instead made changes in the executable to make the tests pass

src/cmd/sdf.bash_completion.sh
```diff
@@ -24,10 +24,10 @@ GZ_SDF_COMPLETION_LIST="
   -k --check
   -d --describe
   -p --print
+  -g --graph
   --inertial-stats
   -h --help
-  --force-version
-  --versions
+  -v --version
```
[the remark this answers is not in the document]
sauk2: The following test will fail if `--force-version` and `--versions` are kept in.

https://github.com/gazebosim/sdformat/blob/181c0ef9f33a56f118ab8f185178c12ddd8b0e1f/src/gz_TEST.cc#L2071-L2072

I noticed those tags were removed in [`gz-transport`](https://github.com/gazebosim/gz-transport/blame/f9cd5c5952beb19e29de2de30f6bbb6cb6d2821d/src/cmd/transport.bash_completion.sh#L73-L74) so I did the same here. Let me know what you think.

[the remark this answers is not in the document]
azeey: hmm, I think we'll need to keep it here. Otherwise, it would not show up in the bash completions. I think it'd be better to modify the `HelpVsCompletionFlags` test to make it pass in this case.

azeey (approved):
Other than the `--force-versions` issue, it LGTM!

sauk2:
> Other than the `--force-versions` issue, it LGTM!

Added dummy flags in the executable to make that `HelpVsCompletionFlags` test pass.

azeey (approved):

Issue #1482, opened 2023-12-18
//inertial/@auto: allow specifying mass instead of density


scpeters:
## Desired behavior

Currently the `//inertial/@auto` attribute enables automatic computation of inertial properties assuming uniform density based on the shape and a density parameter. Since it can be more straightforward to measure mass by weighing an object than measuring density, center of mass, and moment of inertia, we should consider allowing the user to specify the mass and auto-compute the remaining inertial properties without using the density parameter.

## Implementation suggestion

Currently automatic inertial calculations can be enabled by setting `//inertial/@auto == 'true'` and specifying collision geometries in a link:

~~~
<link name="link_with_collision_density">
  <inertial auto="true"/>
  <collision name="collision1">
    <density>...</density>
    <geometry>...</geometry>
  </collision>
  <collision name="collision2">
    <density>...</density>
    <geometry>...</geometry>
  </collision>
</link>
~~~

Note that when https://github.com/gazebosim/sdformat/pull/1335 is merged, density may also be specified in the `//inertial` instead of `//collision` (see the pull request for details on precedence if it is specified in both places):

~~~
<link name="link_with_inertial_density">
  <inertial auto="true">
    <density>...</density>
  </inertial>
  <collision name="collision1">
    <geometry>...</geometry>
  </collision>
  <collision name="collision2">
    <geometry>...</geometry>
  </collision>
</link>
~~~

If a user sets `//inertial/@auto == 'true'` and also specifies mass, inertial pose, or moment of inertia values, a warning is thrown and the user specified values are not used.

~~~
<link name="link_with_ignored_user_inertial_values">
  <inertial auto="true">
    <mass>...</mass>
    <pose>...</pose>
    <inertia>...</inertia>
  </inertial>
  <collision name="collision1">
    <geometry>...</geometry>
  </collision>
  <collision name="collision2">
    <geometry>...</geometry>
  </collision>
</link>
~~~

This feature request is proposing that if a user sets `//inertial/@auto == 'true'` and specifies `//inertial/mass`, that the density values should be ignored, and the inertial values calculated according to the density that would result in the user-specified mass:

~~~
<link name="link_with_auto_inertial_from_mass_not_density">
  <inertial auto="true">
    <mass>...</mass>
  </inertial>
  <collision name="collision1">
    <geometry>...</geometry>
  </collision>
  <collision name="collision2">
    <geometry>...</geometry>
  </collision>
</link>
~~~

scpeters:
[...]

PR #1513, opened 2024-12-09
Resolve auto inertia based on input mass

iche033:
# 🎉 New feature

Closes https://github.com/gazebosim/sdformat/issues/1482

## Summary

Support auto-inertia computation using mass *and* density. Implemented based on the suggestions in https://github.com/gazebosim/sdformat/issues/1482#issuecomment-2353689699 and https://github.com/gazebosim/sdformat/issues/1482#issuecomment-2397585273

inertia is first auto resolved from all collisions as usual. If mass is specified, we normalize the inertia to get unit inertia, then scaling is applied to match the desired mass.

**Marked as draft to get feedback on implementation and correctness of math**

## Checklist
- [x] Added tests

### DIFF
--- a/src/Link_TEST.cc
+++ b/src/Link_TEST.cc
@@ -725,7 +725,7 @@ TEST(DOMLink, ResolveAutoInertialsWithMultipleCollisions)
 }
 
 /////////////////////////////////////////////////
-TEST(DOMLink, InertialValuesGivenWithAutoSetToTrue)
+TEST(DOMLink, ResolveAutoInertialsWithMass)
 {
   std::string sdf = "<?xml version=\"1.0\"?>"
   "<sdf version=\"1.11\">"
@@ -733,6 +733,50 @@ TEST(DOMLink, InertialValuesGivenWithAutoSetToTrue)
   "   <link name='compound_link'>"
   "     <inertial auto='true'>"
   "       <mass>4.0</mass>"
+  "       <pose>1 1 1 2 2 2</pose>"
+  "       <inertia>"
+  "         <ixx>1</ixx>"
+  "         <iyy>1</iyy>"
+  "         <izz>1</izz>"
+  "       </inertia>"
+  "     </inertial>"
+  "     <collision name='box_collision'>"
+  "      <density>2.0</density>"
+  "      <geometry>"
+  "        <box>"
+  "          <size>1 1 1</size>"
+  "        </box>"
+  "        </geometry>"
+  "      </collision>"
+  "    </link>"
+  "   </model>"
+  "  </sdf>";
+
+  sdf::Root root;
+  const sdf::ParserConfig sdfParserConfig;
+  sdf::Errors errors = root.LoadSdfString(sdf, sdfParserConfig);
+  EXPECT_TRUE(errors.empty());
+  EXPECT_NE(nullptr, root.Element());
+
+  const sdf::Model *model = root.Model();
+  const sdf::Link *link = model->LinkByIndex(0);
+  root.ResolveAutoInertials(errors, sdfParserConfig);
+  EXPECT_TRUE(errors.empty());
+
+  EXPECT_DOUBLE_EQ(4.0, link->Inertial().MassMatrix().Mass());
+  EXPECT_EQ(gz::math::Pose3d::Zero, link->Inertial().Pose());
+  EXPECT_EQ(gz::math::Vector3d(0.66666, 0.66666, 0.66666),
+    link->Inertial().MassMatrix().DiagonalMoments());
+}
+
+/////////////////////////////////////////////////
+TEST(DOMLink, InertialValuesGivenWithAutoSetToTrue)
+{
+  std::string sdf = "<?xml version=\"1.0\"?>"
+  "<sdf version=\"1.11\">"
+  "  <model name='compound_model'>"
+  "   <link name='compound_link'>"
+  "     <inertial auto='true'>"
   "       <pose>1 1 1 2 2 2</pose>"
   "       <inertia>"
   "         <ixx>1</ixx>"
--- a/src/Link.cc
+++ b/src/Link.cc
@@ -211,7 +211,6 @@ Errors Link::Load(ElementPtr _sdf, const ParserConfig &_config)
       // If auto is to true but user has still provided
       // inertial values
       if (inertialElem->HasElement("pose") ||
-          inertialElem->HasElement("mass") ||
           inertialElem->HasElement("inertia"))
       {
         Error err(
@@ -681,7 +680,27 @@ void Link::ResolveAutoInertials(sdf::Errors &_errors,
       totalInertia = totalInertia + collisionInertia;
     }
 
-    this->dataPtr->inertial = totalInertia;
+    // If mass is specified, scale inertia to match desired mass
+    auto inertialElem = this->dataPtr->sdf->GetElement("inertial");
+    if (inertialElem->HasElement("mass"))
+    {
+      double mass = inertialElem->Get<double>("mass");
+      const gz::math::MassMatrix3d &totalMassMatrix = totalInertia.MassMatrix();
+      // normalize to get the unit mass matrix
+      gz::math::MassMatrix3d unitMassMatrix(1.0,
+          totalMassMatrix.DiagonalMoments() / totalMassMatrix.Mass(),
+          totalMassMatrix.OffDiagonalMoments() / totalMassMatrix.Mass());
+      // scale the final inertia to match specified mass
+      this->dataPtr->inertial = gz::math::Inertiald(
+          gz::math::MassMatrix3d(mass,
+              mass * unitMassMatrix.DiagonalMoments(),
+              mass * unitMassMatrix.OffDiagonalMoments()),
+          totalInertia.Pose());
+    }
+    else
+    {
+      this->dataPtr->inertial = totalInertia;
+    }
 
     // If CalculateInertial() was called with SAVE_CALCULATION
     // configuration then set autoInertiaSaved to true
@@ -695,7 +714,6 @@ void Link::ResolveAutoInertials(sdf::Errors &_errors,
     {
       this->dataPtr->autoInertiaSaved = true;
       // Write calculated inertia values to //link/inertial element
-      auto inertialElem = this->dataPtr->sdf->GetElement("inertial");
       inertialElem->GetElement("pose")->GetValue()->Set<gz::math::Pose3d>(
         totalInertia.Pose());
       inertialElem->GetElement("mass")->GetValue()->Set<double>(


scpeters:
this looks good! I've added some tests in https://github.com/gazebosim/sdformat/pull/1514 targeting this branch that you can merge into this branch if they seem reasonable

iche033:
> this looks good! I've added some tests in #1514 targeting this branch that you can merge into this branch if they seem reasonable

nice, test looks good, thanks!

iche033:
Merged tests from #1514 and updated Link's `ResolveAutoInertials` API doc on how auto inertia computation works if mass is specified. This PR is now ready for review.

scpeters:
I think we should also update the [auto_inertial_params_proposal proposal](http://sdformat.org/tutorials?tut=auto_inertial_params_proposal) as well, and at some point write a tutorial / documentation since the proposal has now been implemented, but I won't block this PR on either of those tasks

src/Link_TEST.cc
```diff
@@ -763,6 +763,50 @@ TEST(DOMLink, InertialValuesGivenWithAutoSetToTrue)
...
+/////////////////////////////////////////////////
+TEST(DOMLink, InertialValuesGivenWithAutoSetToTrue)
+{
+  std::string sdf = "<?xml version=\"1.0\"?>"
+  "<sdf version=\"1.11\">"
+  "  <model name='compound_model'>"
+  "   <link name='compound_link'>"
+  "     <inertial auto='true'>"
+  "       <pose>1 1 1 2 2 2</pose>"
+  "       <inertia>"
```
scpeters, line 781: the diff is a little confusing here, but I think a new test `ResolveAutoInertialsWithMass` has been added above the existing `InertialValuesGivenWithAutoSetToTrue` test, and that the existing test has been modified, with the `<mass>` removed from the line before this one

if you move the new test between `ResolveAutoInertialsCalledWithAutoFalse` and `AddCollision`, it may clarify the diff

can you confirm and either revert the change to the existing test if it's not needed or add a comment about why it is changed?

include/sdf/Link.hh
```diff
@@ -351,9 +351,12 @@ namespace sdf
...
     /// \brief Calculate & set inertial values(mass, mass matrix
-    /// & inertial pose) for the link. Inertial values can be provided
-    /// by the user through the SDF or can be calculated automatically
-    /// by setting the auto attribute to true.
+    /// & inertial pose) for the link. This function will calculate
+    /// the inertial values if the auto attribute is set to true.
+    /// If mass is not provided by the user, the inertial values will be
+    /// determined based on either link density or collision density if
+    /// explicitly set. Otherwise, if mass is specified, the inertia matrix
+    /// will be scaled to match the desired mass.
```
scpeters, line 359:
```suggestion
    /// will be scaled to match the desired mass, while respecting
    /// the ratio of density values between collisions.
```

python/test/pyLink_TEST.py
```diff
@@ -519,6 +519,47 @@ def test_inertial_values_given_with_auto_set_to_true(self):
         root.resolve_auto_inertials(errors, sdfParserConfig)
         self.assertEqual(len(errors), 0)
 
+        self.assertEqual(4.0, link.inertial().mass_matrix().mass())
+        self.assertEqual(Pose3d.ZERO, link.inertial().pose())
+        self.assertEqual(Vector3d(0.66666, 0.66666, 0.66666),
+            link.inertial().mass_matrix().diagonal_moments())
+
+    def test_inertial_values_given_with_auto_set_to_true(self):
```
scpeters, line 527: my comments on Link_TEST.cc apply here as well

include/sdf/Link.hh
iche033: [cf4d77b](https://github.com/gazebosim/sdformat/pull/1513/commits/cf4d77b136deb3a28b757686a494de3b087fa4eb)

src/Link_TEST.cc
iche033: moved test so the diff is not easier to read. I duplicated the test so that there's now one test with mass specified and one without mass. [cf4d77b](https://github.com/gazebosim/sdformat/pull/1513/commits/cf4d77b136deb3a28b757686a494de3b087fa4eb)

iche033:
> I think we should also update the [auto_inertial_params_proposal proposal](http://sdformat.org/tutorials?tut=auto_inertial_params_proposal) as well

Updated the proposal in https://github.com/gazebosim/sdf_tutorials/pull/100

src/Link_TEST.cc
scpeters: thanks! There was one expectation in this test that I thought should be updated as well, but it was outside the lines that I could add a comment on, so I just changed it directly in https://github.com/gazebosim/sdformat/pull/1513/commits/ae7685c34c85a751c9da405ec321ad67f7e27232

scpeters (approved):

scpeters:
I think we don't have a test involving multiple collisions with different density values attached to a link with auto-inertial and explicit mass. I think that would be ideal, but I won't block on it

iche033:
> I think we don't have a test involving multiple collisions with different density values attached to a link with auto-inertial and explicit mass. I think that would be ideal, but I won't block on it

added tests for auto inertial with link mass and multiple collisions in [c072a17](https://github.com/gazebosim/sdformat/pull/1513/commits/c072a17855667b2296ad099b9876e9c7a6f1a3f6) and [410d4e6](https://github.com/gazebosim/sdformat/pull/1513/commits/410d4e6ddc1452e8baa64a313d7a1e509e57add1)

azeey (commented):
Looks great overall. One of the goals for calculating the inertia based on mass was to make it possible to calculate the inertia even when `density` values are not set. Do we know if that works? Maybe we can add a test?

This also means we can stop emitting a warning if a Collision is missing its `<density>` if `mass` is provided: https://github.com/gazebosim/sdformat/blob/92da673710ad0d1b1da3f2a58be874024177bba2/src/Collision.cc#L296-L305

src/Link_TEST.cc
```diff
@@ -763,12 +766,121 @@ TEST(DOMLink, InertialValuesGivenWithAutoSetToTrue)
...
+  "       <mass>4.0</mass>"
+  "       <pose>1 1 1 2 2 2</pose>"
+  "       <inertia>"
+  "         <ixx>1</ixx>"
+  "         <iyy>1</iyy>"
+  "         <izz>1</izz>"
+  "       </inertia>"
+  "     </inertial>"
+  "     <collision name='box_collision'>"
+  "      <density>2.0</density>"
```
scpeters, line 796: @azeey I think we could delete this line in order to test without density being set

scpeters:
> > I think we don't have a test involving multiple collisions with different density values attached to a link with auto-inertial and explicit mass. I think that would be ideal, but I won't block on it
> 
> added tests for auto inertial with link mass and multiple collisions in [c072a17](https://github.com/gazebosim/sdformat/pull/1513/commits/c072a17855667b2296ad099b9876e9c7a6f1a3f6) and [410d4e6](https://github.com/gazebosim/sdformat/pull/1513/commits/410d4e6ddc1452e8baa64a313d7a1e509e57add1)

thanks; I'm did the math to verify the inertia values for that test case, and it looks right but a bit complex. I'd like to have our test expectations documented, so I'll propose a slight variation of the test geometry that places the lumped center of mass at the origin to simplify the math and then document that. I can do that in a follow-up PR if other concerns are resolved before I have a chance to write it up

src/Link_TEST.cc
```diff
@@ -763,12 +766,121 @@ TEST(DOMLink, InertialValuesGivenWithAutoSetToTrue)
...
+  EXPECT_EQ(gz::math::Vector3d(0.66666, 0.66666, 0.66666),
+    link->Inertial().MassMatrix().DiagonalMoments());
+}
+
+/////////////////////////////////////////////////
+TEST(DOMLink, ResolveAutoInertialsWithMassAndMultipleCollisions)
+{
+  // A model with link inertial auto set to true.
+  // The inertia matrix is specified but should be ignored.
+  // <mass> is speicifed - the auto computed inertial values should
```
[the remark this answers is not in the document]
iche033: fixed. [ca94a22](https://github.com/gazebosim/sdformat/pull/1513/commits/ca94a22f6327cb7d51e7b4ef912588c693bc4182)

iche033:
> Looks great overall. One of the goals for calculating the inertia based on mass was to make it possible to calculate the inertia even when density values are not set. Do we know if that works? Maybe we can add a test?

If density is not explicitly set by the user, the default value should be used. Yes I can add or change a test in a follow-up PR after @scpeters updates the test for computing auto inertia involving multiple collisions

> This also means we can stop emitting a warning if a Collision is missing its <density> if mass is provided:

removed warning [ca94a22](https://github.com/gazebosim/sdformat/pull/1513/commits/ca94a22f6327cb7d51e7b4ef912588c693bc4182)

iche033:
> I'd like to have our test expectations documented, so I'll propose a slight variation of the test geometry that places the lumped center of mass at the origin to simplify the math and then document that. I can do that in a follow-up PR if other concerns are resolved before I have a chance to write it up

thanks! I've addessed remaining comments except the one about adding a test with no `<density>` specified. Let's do that in follow-up PR(s).

azeey:
One last thing I thought of is whether this would break existing SDF files. If you have a link with the mass specified as well as densities in the collisions, previously, the mass was completely ignored. Now, the mass will override the densities which will likely result in different MOI values. Am I correct?

iche033:
> One last thing I thought of is whether this would break existing SDF files. If you have a link with the mass specified as well as densities in the collisions, previously, the mass was completely ignored. Now, the mass will override the densities which will likely result in different MOI values. Am I correct?

yes this will be a behavior change so we would likely need to retarget this to `main`. Alternatively we could add another attribute to toggle whether to use mass or density for auto-inertia, e.g. <inertial auto=true auto_use_mass=true>.

scpeters:
> One last thing I thought of is whether this would break existing SDF files. If you have a link with the mass specified as well as densities in the collisions, previously, the mass was completely ignored. Now, the mass will override the densities which will likely result in different MOI values. Am I correct?

Currently there is a warning when specifying mass with `//inertial@auto == true` saying that the mass will be ignored. This PR definitely introduces a behavior change for files like that since the mass will rescale the MOI values (though with multiple collisions the relative density still has an effect on the MOI). As to whether this is a "breaking change", this fixes a warning and honors the intent of the mass value.

We can certainly merge this to `main`. If we want to backport it to Ionic or Harmonic, we would need to clearly communicate the behavior change. Since it fixes an existing warning, I think there's a case to be made that it could be backported.

scpeters:
> > I'd like to have our test expectations documented, so I'll propose a slight variation of the test geometry that places the lumped center of mass at the origin to simplify the math and then document that. I can do that in a follow-up PR if other concerns are resolved before I have a chance to write it up
> 
> thanks! I've addessed remaining comments except the one about adding a test with no `<density>` specified. Let's do that in follow-up PR(s).

here's my update to the multi-collision / multi-density test: https://github.com/gazebosim/sdformat/pull/1515

iche033:
rebased changes to `main` and forced pushed

scpeters:
> > This also means we can stop emitting a warning if a Collision is missing its  if mass is provided:
> 
> removed warning [ca94a22](https://github.com/gazebosim/sdformat/pull/1513/commits/ca94a22f6327cb7d51e7b4ef912588c693bc4182)

It looks like the warning is unconditionally removed, regardless of whether `//inertial/mass` is set or not. It's tricky to conditionally remove the warning since `Collision::CalculateInertial` doesn't know whether the Link inertial has an explicit mass value set. I think we would need to either add a `_warnIfDensityIsUnset` flag to `Collision::CalculateInertial` method, or we could try to delete the `ELEMENT_MISSING` errors in [this loop in Link.cc](https://github.com/gazebosim/sdformat/blob/69c82c00fae5102f360b299ff6f396bab126ced3/src/Link.cc#L675-L682)

iche033:
> I think we would need to either add a _warnIfDensityIsUnset flag to `Collision::CalculateInertial` method, or we could try to delete the ELEMENT_MISSING errors in [this loop in Link.cc](https://github.com/gazebosim/sdformat/blob/69c82c00fae5102f360b299ff6f396bab126ced3/src/Link.cc#L675-L682)

I went with the option to add a `_warnIfDensityIsUnset` arg to `Collision::CalculateInertial` in [2509d55](https://github.com/gazebosim/sdformat/pull/1513/commits/2509d550c25e14757fef36b575294a1b76401f82). I think it implements what you're suggestion, please take another look, thanks.

python/test/pyLink_TEST.py
```diff
@@ -519,11 +521,112 @@ def test_inertial_values_given_with_auto_set_to_true(self):
...
+        errors = []
+        root.resolve_auto_inertials(errors, sdfParserConfig)
+        self.assertEqual(len(errors), 0)
+
+        self.assertEqual(4.0, link.inertial().mass_matrix().mass())
+        self.assertEqual(Pose3d.ZERO, link.inertial().pose())
+        self.assertEqual(Vector3d(0.66666, 0.66666, 0.66666),
+            link.inertial().mass_matrix().diagonal_moments())
+
+    def test_resolveauto_inertialsWithMassAndMultipleCollisions(self):
```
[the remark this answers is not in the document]
iche033: I think it maybe redundant but I went ahead and updated the python test in https://github.com/gazebosim/sdformat/pull/1517.

src/Link_TEST.cc
iche033: added new test without density being set in https://github.com/gazebosim/sdformat/pull/1517

python/test/pyLink_TEST.py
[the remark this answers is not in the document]
scpeters: 👍

iche033:
> I went with the option to add a _warnIfDensityIsUnset arg to Collision::CalculateInertial in [2509d55](https://github.com/gazebosim/sdformat/pull/1513/commits/2509d550c25e14757fef36b575294a1b76401f82). I think it implements what you're suggestion, please take another look, thanks.

@scpeters do the changes in [2509d55](https://github.com/gazebosim/sdformat/pull/1513/commits/2509d550c25e14757fef36b575294a1b76401f82) look good to you?

scpeters:
> > I went with the option to add a _warnIfDensityIsUnset arg to Collision::CalculateInertial in [2509d55](https://github.com/gazebosim/sdformat/pull/1513/commits/2509d550c25e14757fef36b575294a1b76401f82). I think it implements what you're suggestion, please take another look, thanks.
> 
> @scpeters do the changes in [2509d55](https://github.com/gazebosim/sdformat/pull/1513/commits/2509d550c25e14757fef36b575294a1b76401f82) look good to you?

Yes, that looks good; I want to see if we can add a test that verifies the presence / absence of the errors / warnings depending on that new flag

scpeters:
> > > I went with the option to add a _warnIfDensityIsUnset arg to Collision::CalculateInertial in [2509d55](https://github.com/gazebosim/sdformat/pull/1513/commits/2509d550c25e14757fef36b575294a1b76401f82). I think it implements what you're suggestion, please take another look, thanks.
> > 
> > 
> > @scpeters do the changes in [2509d55](https://github.com/gazebosim/sdformat/pull/1513/commits/2509d550c25e14757fef36b575294a1b76401f82) look good to you?
> 
> Yes, that looks good; I want to see if we can add a test that verifies the presence / absence of the errors / warnings depending on that new flag

see #1520

iche033:
> see https://github.com/gazebosim/sdformat/pull/1520

merged!

scpeters (approved):

iche033:
@scpeters  and @azeey, I would like to backport this PR so reviving the discussion on how to proceed as this introduces a breaking change when users have `<inertia auto=true>` and `<mass>` specified in their existing SDF files. 

So far the following options were proposed:

1. Add a new attribute to toggle behavior

> yes this will be a behavior change so we would likely need to retarget this to main. Alternatively we could add another attribute to toggle whether to use mass or density for auto-inertia, e.g. `<inertial auto=true auto_use_mass=true>`.

2. Treat this as a "fix" and backport this PR as is as this fixes a warning

> Currently there is a warning when specifying mass with //inertial@auto == true saying that the mass will be ignored. This PR definitely introduces a behavior change for files like that since the mass will rescale the MOI values (though with multiple collisions the relative density still has an effect on the MOI). As to whether this is a "breaking change", this fixes a warning and honors the intent of the mass value.

Any preferences or other thoughts?

bperseghetti:
I think viewing it as a bug fix makes sense, I vote for 2 as well.

iche033:
thanks for the input! @azeey if options 2 sounds ok to you, I'll go ahead with the backport.

azeey:
Okay, let's go with option 2. I'd rather not make such breaking changes, but I'd really like to see this backported and I don't see a clean way to make this backward compatible.  If we get complaints, we can implement a `<gz:policy>` that applies the old behavior.

iche033:
@mergifyio backport sdf15

iche033:
@mergifyio backport sdf14

PR #1515, opened 2024-12-13
Update multiple-collision test with justification

scpeters:
Targeting #1513 

Modifies the auto-inertial test with multiple collisions with different densities so that the lumped center of gravity is at the link origin and derives the expected moment of inertia values.


iche033 (approved):
looks good to me, thanks!

PR #1520, opened 2024-12-20
Add expectations about auto-inertial warnings

scpeters:
Targeting #1513.

This updates expectations for three test cases related to auto-inertial warnings by using the `ERR` enforcement policy in `ParserConfig`:

* ResolveAutoInertialsWithDifferentDensity: parse first with enforcement policy `ERR` and expect a warning about unset collision density. Keep the subsequent parsing with default enforcement policy that expects no errors but will print the warning to the console.
* InertialValuesGivenWithAutoSetToTrue: parse first with enforcement policy `ERR` and expect a warning about user-specified inertial values. Again keep the subsequent parsing with default enforcement policy that expects no errors but will print the warning to the console.
* ResolveAutoInertialsWithMassAndDefaultDensity: remove the user-specified `//inertial/pose` and `//inertial/inertia` values, parse with enforcement policy `ERR` and expect no warnings/errors.

### DIFF
--- a/src/Link_TEST.cc
+++ b/src/Link_TEST.cc
@@ -389,6 +389,24 @@ TEST(DOMLink, ResolveAutoInertialsWithDifferentDensity)
     </model>
   </sdf>)";
 
+  // Parse first with enforcement policy set to ERR to detect warnings.
+  {
+    sdf::Root root;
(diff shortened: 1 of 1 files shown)


iche033 (approved):
looks good to me, thanks!

PR #1465, opened 2024-07-29
WIP: Implement gz sdf as standalone executable

scpeters:
# 🎉 New feature

Part of https://github.com/gazebosim/gz-tools/issues/7.

## Summary

This is a work-in-progress to implement the `gz sdf` functionality in a standalone executable (similar to https://github.com/gazebosim/gz-transport/pull/216). The standalone executable has added alongside the existing approach. Once the standalone executable is complete and functioning, the old approach should be removed.

This is not targeting the first release of Gazebo Ionic.

## Test it

Use `UNIT_gz_TEST` as the guide.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #1539, opened 2025-02-07
Standalone  `gz sdf` executable

sauk2:
# 🎉 New feature

Part of https://github.com/gazebosim/gz-tools/issues/7 and a continuation of https://github.com/gazebosim/sdformat/pull/1465

## Summary
This PR introduces a standalone `gz sdf` executable following a similar approach to [`gz-transport`](https://github.com/gazebosim/gz-transport/pull/216). It is a continuation of  https://github.com/gazebosim/sdformat/pull/1465 with all the review comments addressed.

## Test it
Check for `UNIT_gz_TEST`

## Checklist
- [x] Added tests


src/cmd/sdf_main.cc
```diff
@@ -0,0 +1,201 @@
...
+    default:
+      // In the event that there is no command, display help
+      throw CLI::CallForHelp();
+  }
+}
+
+//////////////////////////////////////////////////
+void addSdfFlags(CLI::App &_app)
+{
+  auto opt = std::make_shared<SdfOptions>();
```
ahcorde, line 112: include `<memory>`

src/cmd/sdf_main.cc
```diff
@@ -0,0 +1,201 @@
...
+    SDF_PROTOCOL_VERSION ")")
+    ->expected(0, 1)
+    ->default_val(SDF_PROTOCOL_VERSION);
+
+  command->add_option_function<std::string>("-g,--graph",
+    [opt](const std::string &_graphType){
+      opt->command = SdfCommand::kSdfGraph;
+      opt->graphType = _graphType;
+    },
+    "<pose, frame> filepath  Print the PoseRelativeTo or FrameAttachedTo "
```
ahcorde, line 158:
```suggestion
    "<pose, frame> filepath Print the PoseRelativeTo or FrameAttachedTo "
```

src/cmd/sdf_main.cc
```diff
@@ -0,0 +1,201 @@
...
+  _app.callback([opt](){runSdfCommand(*opt); });
+}
+
+//////////////////////////////////////////////////
+int main(int argc, char** argv)
+{
+  CLI::App app{"Utilities for SDFormat files."};
+
+  app.add_flag_callback("-v,--version", [](){
+      std::cout << SDF_VERSION_FULL << std::endl;
```
ahcorde, line 189: include `<iostream>`

ahcorde (changes requested):

src/cmd/gz.hh
```diff
@@ -0,0 +1,64 @@
...
+
+#include <cstring>
+
+#include <sdf/sdf_config.h>
+#include "sdf/system_util.hh"
+
+/// \brief External hook to execute 'gz sdf -k' from the command line.
+/// \param[in] _path Path to the SDF file to validate.
+/// \return Zero on success, negative one otherwise.
+int cmdCheck(const char *_path);
```
azeey, line 29: Can you wrap those in the `sdf` namespace? You can follow the pattern from many of the .hh files in this library.

src/gz.hh
azeey, line 1: It would be great to use `git mv` on this file to preserve history. Currently it looks like it's being removed here and added in `cmd/gz.hh`.

src/cmd/gz.cc
```diff
@@ -139,9 +137,9 @@ extern "C" SDFORMAT_VISIBLE int cmdDescribe(const char *_version)
 }
 
 //////////////////////////////////////////////////
-extern "C" SDFORMAT_VISIBLE int cmdPrint(const char *_path,
-    int _inDegrees, int _snapToDegrees, float _snapTolerance,
-    int _preserveIncludes, int _outPrecision, int _expandAutoInertials)
+int cmdPrint(const char *_path, int _inDegrees, int _snapToDegrees,
+    float _snapTolerance, int _preserveIncludes, int _outPrecision,
+    int _expandAutoInertials)
```
azeey, line 142: `_preserveIncludes` and `_expandAutoInertials` could be `bool`s.

src/cmd/gz.cc
```diff
@@ -22,7 +22,7 @@
 #include <string.h>
 #include <vector>
 
-#include "sdf/config.hh"
+#include "sdf/sdf_config.h"
```
azeey, line 25: I think we actually want `sdf/config.hh`.  https://github.com/gazebosim/sdformat/blob/33b8c91cbbd0ad6f7f8e35d6693ab120192c17c7/include/sdf/sdf_config.h#L21-L23

src/cmd/gz_TEST.cc
```diff
@@ -37,9 +37,6 @@
   #define pclose _pclose
 #endif
 
-// DETAIL_GZ_CONFIG_PATH is compiler definition set in CMake.
-#define GZ_CONFIG_PATH DETAIL_GZ_CONFIG_PATH
```
azeey, line 41: I believe this was done intentionally so it's clear where the definition came from. That is: Define `DETAIL_GZ_CONFIG_PATH` in CMake and add the alias here with a comment.

src/cmd/sdf_main.cc
```diff
@@ -0,0 +1,203 @@
...
+    case SdfCommand::kSdfGraph:
+      cmdGraph(_opt.graphType.c_str(), _opt.filepath.c_str());
+      break;
+    case SdfCommand::kSdfPrintInertialStats:
+      cmdInertialStats(_opt.filepath.c_str());
+      break;
+    case SdfCommand::kSdfPrint:
+      cmdPrint(_opt.filepath.c_str(), _opt.degrees, *_opt.snapToDegrees,
+               _opt.snapTolerance, _opt.preserveIncludes, *_opt.precision,
+               _opt.expandAutoInertials);
```
azeey, line 102: What would happen if the optional parameters are not specified by the user? Would calling `operator*` throw? This might be where the windows warning is coming from. It might be best to just initialize these with default values instead of using `std::optional`.

azeey (changes requested):

src/cmd/gz.cc
```diff
@@ -34,328 +34,329 @@
...
+    {
+      result = -1;
+    }
 
-  if (!sdf::filesystem::exists(_path))
-  {
-    std::cerr << "Error: File [" << _path << "] does not exist.\n";
-    return -1;
-  }
+    if (!sdf::filesystem::exists(_path))
```
ahcorde, line 86: include `<filesystem>`

ahcorde (changes requested):

src/gz.hh
sauk2: I tried to do this but subsequent modifications made the file too different for git to recognize it

ahcorde (approved):

sdf/CMakeLists.txt
```diff
@@ -47,5 +47,7 @@ if (GZ_PROGRAM)
       COMMENT "Generating full description for spec ${desc_ver}"
       VERBATIM)
   endforeach()
-  add_custom_target(sdf_descriptions DEPENDS ${description_targets} ${PROJECT_LIBRARY_TARGET_NAME})
```
j-rivero, line 50: Was the removal of `PROJECT_LIBRARY_TARGET_NAME` from DEPENDS intended?

sdf/CMakeLists.txt
```diff
@@ -47,5 +47,7 @@ if (GZ_PROGRAM)
...
+  add_custom_target(sdf_descriptions DEPENDS ${description_targets})
+  # Add a dependency on the gz-sdformat-sdf target which is created in in ../cmd/CMakeLists
+  add_dependencies(sdf_descriptions gz-sdformat-sdf)
```
j-rivero, line 52: This block is under the `if (GZ_PROGRAM)` check. Do we need to have the `gz` tool generate the standalone gz-sdformat-sdf` tool?

src/cmd/CMakeLists.txt
```diff
@@ -1,15 +1,54 @@
...
+      gz-utils${GZ_UTILS_VER}::gz-utils${GZ_UTILS_VER}
+      ${EXTRA_TEST_LIB_DEPS}
+    ENVIRONMENT
+      SDF_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
+  )
+
+  if(TARGET UNIT_gz_TEST)
+    target_compile_definitions(UNIT_gz_TEST
+      PRIVATE
+        "GZ_PATH=\"${GZ_PROGRAM}\""
```
j-rivero, line 33: Are we using the new standalone tool to run the gz_TEST? Is the new executable being tested in the test suite?

sdf/CMakeLists.txt
sauk2: Thanks for bringing this up! I have removed the dependency on `GZ_PROGRAM` and instead find the executable of `gz-sdformat-sdf` target (51e848570a02f31031316922a9561e4ecc493cdd)

src/cmd/CMakeLists.txt
sauk2: Yes, we are. The tests run with [this ruby script invoking the executable](https://github.com/sauk2/sdformat/blob/51e848570a02f31031316922a9561e4ecc493cdd/src/cmd/CMakeLists.txt#L40-L60).

j-rivero: ah thanks! it was a bit hidden.

sdf/CMakeLists.txt
j-rivero: @sauk2 this final question and we are good to go !

sauk2: Oh sorry, I forgot to reply. This change was made by @azeey as a part of this commit (https://github.com/gazebosim/sdformat/pull/1489/commits/615048c27784459f12594da214836a9d709a6e4d). He'd be the best person to clarify!

src/cmd/sdf_main.cc
```diff
@@ -0,0 +1,205 @@
...
+      break;
+    case SdfCommand::kNone:
+    default:
+      // In the event that there is no command, display help
+      throw CLI::CallForHelp();
+  }
+}
+
+//////////////////////////////////////////////////
+void addSdfFlags(CLI::App &_app)
```
j-rivero, line 114: The current message is:
```
Usage: ./bin/gz-sdformat-sdf [OPTIONS] [filepath]
```

I think that it is not very usable since the expected schema is more something like:
```
gz-sdformat-sdf [command] [command-options] [filepath]
```
Somehow: specify that command is the main first argument and that options are really affecting the way that commands work and probably that they don't work without a command. Also great if command can appear before the options and the filepath.

j-rivero (commented):
Final comments from the runtime check. Let me know @sauk2 if you need help implementing them.

src/cmd/sdf_main.cc
```diff
@@ -0,0 +1,205 @@
...
+    [opt](const std::string &_version){
+      opt->command = SdfCommand::kSdfDescribe;
+      opt->version = _version;
+    },
+    "Print the aggregated SDFormat spec description. Latest version ("
+    SDF_PROTOCOL_VERSION ")")
+    ->expected(0, 1)
+    ->default_val(SDF_PROTOCOL_VERSION);
+
+  command->add_option_function<std::string>("-g,--graph",
```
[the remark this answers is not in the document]
sauk2: Sorry for the confusion. This option needs a following `pose` or `frame` text as argument. I have changed the description to reflect this (https://github.com/gazebosim/sdformat/pull/1539/commits/b15e7679025bc5209faea5ee2135e0e546550ce8).

I have also added a validation that enforces the argument to be either `pose` or `frame`.

Now it looks like this in the help output
```
-g [--graph] TEXT:{pose,frame}
                                Print the PoseRelativeTo or FrameAttachedTo graph by following
                                with either pose or frame as argument respectively.
                                (WARNING: This is for advanced use only and the output may change
                                without any promise of stability)
                                Requires: filepath

```

src/cmd/sdf_main.cc
```diff
@@ -0,0 +1,205 @@
...
+{
+  CLI::App app{"Utilities for SDFormat files."};
+
+  app.add_flag_callback("-v,--version", [](){
+      std::cout << SDF_VERSION_FULL << std::endl;
+      throw CLI::Success();
+    },
+    "Print the current library version");
+
+  // Dummy flags handled by gz-tools
```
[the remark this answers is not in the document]
sauk2: These flags are handled directly by [`gz-tools`](https://github.com/gazebosim/gz-tools/blob/2d2cf690aae6533d46cc9264f3347d8986c79037/src/gz.in#L200-L203) but removing them from this executable will cause the following test to fail!

https://github.com/gazebosim/sdformat/blob/71fe5048d533ea2f422fc6f0b0b98f75e22e0720/src/gz_TEST.cc#L2085-L2086

So in [this discussion](https://github.com/gazebosim/sdformat/pull/1489#discussion_r1909136354) it was decided to keep them in as dummy flags and let `gz-tools` handle them.

src/cmd/sdf_main.cc
sauk2: The usage message seems to be generated by [`gz-utils`](https://github.com/gazebosim/gz-utils/blob/fd0a40c09be4fd7b2a40d877bb56d2fa1c7607e3/cli/include/vendored-cli/gz/utils/cli/Formatter.hpp#L93). I could not find any setting I could change in the executable to reflect this change.

Do you think it would be a good idea to make changes in `gz-utils` side to bring about this change?

src/cmd/sdf_main.cc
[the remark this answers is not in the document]
j-rivero: Thanks. The user experience is still a bit weird not providing any feedback or output but we can live with it by now.

src/cmd/sdf_main.cc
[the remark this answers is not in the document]
j-rivero: Is still true the message that says "Requires: filepath" ?

src/cmd/sdf_main.cc
j-rivero: I think that the relevant changes needs to go into the `GzFormatter` in https://github.com/gazebosim/gz-utils/blob/fd0a40c09be4fd7b2a40d877bb56d2fa1c7607e3/cli/include/gz/utils/cli/GzFormatter.hpp#L42-L65

It would be great to have them in but it is out of the scope for this PR I would say. Feel free to create an issue or a PR for it.

j-rivero (approved):
Another great contribution @sauk2 , thanks !

j-rivero:
ops, I think that we need your approval @azeey here.

sdf/CMakeLists.txt
azeey: Since we're adding a dependency on `gz-sdformat-sdf` below, which is ultimately what is used to generate these files, this target no longer needs to depend on `PROJECT_LIBRARY_TARGET_NAME`

src/cmd/sdf_main.cc
azeey: @j-rivero I'm not sure what you mean. The `gz-sdformat-sdf` is not intended to be used directly, but through the `gz` tool. When you run `gz sdf`, it'll be running `gz-sdformat-sdf` under the hood. So as long as the usage is `gz sdf [OPTIONS] ...`, which is what the current usage (before this PR) is, I think we're good.

src/cmd/gz.hh
```diff
@@ -0,0 +1,73 @@
...
+  //
+
+  /// \brief External hook to execute 'gz sdf -k' from the command line.
+  /// \param[in] _path Path to the SDF file to validate.
+  /// \return Zero on success, negative one otherwise.
+  int cmdCheck(const char *_path);
+
+  /// \brief External hook to read the library version.
+  /// \return C-string representing the version. Ex.: 0.1.2
+  char *gzVersion();
```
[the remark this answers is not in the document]
azeey: I'm okay with removing it.

src/gz.cc
azeey, line 1: Can you use `gz mv` to move this file to `cmd/gz.cc` to preserve history? Right now, it looks like this is removed and a new file is being added in `cmd/gz.cc`.

src/cmd/sdf_main.cc
```diff
@@ -0,0 +1,206 @@
...
+
+//////////////////////////////////////////////////
+void addSdfFlags(CLI::App &_app)
+{
+  auto opt = std::make_shared<SdfOptions>();
+
+  auto filepathOpt =
+    _app.add_option("filepath", opt->filepath,
+                    "Path to an SDFormat file.");
+  _app.add_flag("-i,--preserve-includes", opt->preserveIncludes,
```
azeey, line 121: Many of these options are relevant when the `-p` flag is active. If you see how the  usage is printed in the original `gz sdf` version, you'll see the relationship:

```
  -g [ --graph ] <pose, frame> arg  Print the PoseRelativeTo or FrameAttachedTo graph. (WARNING: This is for advanced
                                    use only and the output may change without any promise of stability)
  -p [ --print ] arg                Print converted arg. Note the quaternion representation of the
                                    rotational part of poses and unit vectors will be normalized.
      -i [ --preserve-includes ]    Preserve included tags when printing converted arg (does not preserve merge-includes).
      --degrees                     Pose rotation angles are printed in degrees.
      --snap-to-degrees arg         Snap pose rotation angles to this specified interval in degrees. This value must be
                                    larger than 0, less than or equal to 360, and larger than the defined snap tolerance.
      --snap-tolerance arg          Used in conjunction with --snap-to-degrees, specifies the tolerance at which snapping
                                    occurs. This value must be larger than 0, less than 360, and less than the defined
                                    degrees value to snap to. If unspecified, its default value is 0.01.
      --inertial-stats  arg         Prints moment of inertia, centre of mass, and total mass from a model sdf file.
      --precision arg               Set the output stream precision for floating point numbers. The arg must be a positive integer.
  -h [ --help ]                     Print this help message.

```

Is there a way to add this type of dependency in CLI11?

src/gz.cc
sauk2: I tried to do this, but git does not recognize the files as moved due to successive modifications. Locally, I am able to see 'renamed' when I drop the similarity index to 10% but I don't know how to enforce this on GitHub.
Same with `gz.hh`

src/cmd/sdf_main.cc
sauk2: Done! I added a dependence on `--print`.

The usage looks like this now.

```
Positionals:
  filepath TEXT               Path to an SDFormat file.
                              Required by: --print
                              Required by: --inertial-stats
                              Required by: --graph
                              Required by: --check
                              

Options:
  -h [--help]                 Print this help message and exit
                              
  -v [--version]              Print the current library version
                              
  --force-version             Use a specific library version.
                              
  --versions                  Show the available versions.
                              
  -i [--preserve-includes]    Preserve included tags when printing converted arg (does not preserve merge-includes).
                              Requires: --print
                              
  --degrees                   Printed pose rotations are will be in degrees.
                              Requires: --print
                              
  --expand-auto-inertials     Auto-computed inertial values will be printed.
                              Requires: --print
                              
  --precision INT             Set the output stream precision for floating point numbers.
                              Requires: --print
                              
  --snap-to-degrees INT       Printed rotations are snapped to specified degree intervals.
                              Requires: --print
                              
  --snap-tolerance FLOAT      Printed rotations are snapped if they are within this specified tolerance.
                              Requires: --print
                              
[Option Group: command]
  Command to be executed.
  Options:
    -k [--check]                Check if an SDFormat file is valid.
                                Requires: filepath
                                
    -d [--describe] TEXT=1.12   Print the aggregated SDFormat spec description. Latest version (1.12)
                                
    -g [--graph] TEXT:{pose,frame}
                                Print the PoseRelativeTo or FrameAttachedTo graph by following
                                with either pose or frame as argument respectively.
                                (WARNING: This is for advanced use only and the output may change
                                without any promise of stability)
                                Requires: filepath
                                
    --inertial-stats            Prints moment of inertia, centre of mass, and total mass from a model sdf file.
                                Requires: filepath
                                
    -p [--print]                Print converted arg. Note the quaternion representation of the
                                rotational part of poses and unit vectors will be normalized.
                                Requires: filepath
                                Required by: --snap-tolerance
                                Required by: --snap-to-degrees
                                Required by: --precision
                                Required by: --expand-auto-inertials
                                Required by: --degrees
                                Required by: --preserve-includes

```

src/cmd/sdf_main.cc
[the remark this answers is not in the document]
sauk2: Yes, it still shows a dependency on `filepath` because the `--graph` requires it as [this](https://github.com/sauk2/sdformat/blob/0f40dd57ef29673bf6d3c4f4fd4a271bdc018ffb/src/cmd/gz.cc#L189) function is called to process that flag.

scpeters:
Thanks for finishing this! Apologies that I didn't help more with the review

PR #1542, opened 2025-02-11
Handle invalid mesh inertia produced by custom mesh inertia calculator

iche033:
# 🦟 Bug fix

## Summary

The custom mesh inertia calculator can produce invalid / null inertias, see [code in MeshInertiaCalculator](https://github.com/gazebosim/gz-sim/blob/4b6445faa1a060eec548a71034b904e802e0b3d5/src/MeshInertiaCalculator.cc#L239-L242). SDFormat currently [logs en error](https://github.com/gazebosim/sdformat/blob/2b583cbc3f40ac4f1af27613a49b475f2d0d99b7/src/Collision.cc#L323-L328) but proceeds to return an empty `gz::math::Inertiald` object with 0 mass and inertia matrix. This sdf error leads gz-sim to [shutdown immediately](https://github.com/gazebosim/gz-sim/blob/4b6445faa1a060eec548a71034b904e802e0b3d5/src/Server.cc#L201-L210). This PR prevents collision inertial calculation from returning zero inertial values, and instead it will now return default inertia if an invalid inertia is detected. 

The other option is to keep existing behavior and return an sdf error but update gz-sim to add a special case for continuing sim with presence of inertial sdf errors, which is also not ideal because 0 mass and inertia matrix could cause weird behavior in physics sim.

I went with the current approach because this behavior is similar to the case when [no custom mesh inertial calculator exists](https://github.com/gazebosim/sdformat/blob/2b583cbc3f40ac4f1af27613a49b475f2d0d99b7/src/Mesh.cc#L379-L393), in which case the default inertial values are also returned.

## Checklist
- [x] Added tests

### DIFF
--- a/src/Collision.cc
+++ b/src/Collision.cc
@@ -324,10 +324,7 @@ void Collision::CalculateInertial(
   {
     _errors.push_back({ErrorCode::LINK_INERTIA_INVALID,
         "Inertia Calculated for collision: " +
-        this->dataPtr->name + " is invalid, using default inertial values."});
-    _inertial = gz::math::Inertiald(
-        gz::math::MassMatrix3d(1, gz::math::Vector3d::One,
-        gz::math::Vector3d::Zero), gz::math::Pose3d::Zero);
+        this->dataPtr->name + " is invalid."});
   }
   else
   {
--- a/src/Collision.cc
+++ b/src/Collision.cc
@@ -324,7 +324,10 @@ void Collision::CalculateInertial(
   {
     _errors.push_back({ErrorCode::LINK_INERTIA_INVALID,
         "Inertia Calculated for collision: " +
-        this->dataPtr->name + " is invalid."});
+        this->dataPtr->name + " is invalid, using default inertial values."});
+    _inertial = gz::math::Inertiald(
+        gz::math::MassMatrix3d(1, gz::math::Vector3d::One,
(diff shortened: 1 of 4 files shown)


scpeters:
> The custom mesh inertia calculator can produce invalid / null inertias, see [code in MeshInertiaCalculator](https://github.com/gazebosim/gz-sim/blob/4b6445faa1a060eec548a71034b904e802e0b3d5/src/MeshInertiaCalculator.cc#L239-L242).

from looking at [MeshInertiaCalculator.hh](https://github.com/gazebosim/gz-sim/blob/4b6445faa1a060eec548a71034b904e802e0b3d5/src/MeshInertiaCalculator.hh#L70-L72), I see the phrase "it works on triangle water-tight meshes for simple polyhedron"

for meshes using convex decomposition, I think we could guarantee that the convex hulls of these meshes are water-tight?

iche033:
I can extend the MeshInertialCalculator to work with convex decomposed meshes as it currently doesn't. 

That can be done in parallel. I think we still do need to handle the case that the mesh inertial calculator fails though, i.e. the changes proposed in this PR.

scpeters:
instead of changing the default behavior, I'd prefer added a policy enum to ParserConfig like in https://github.com/gazebosim/sdformat/pull/1543

iche033:
closing in favor or https://github.com/gazebosim/sdformat/pull/1543

PR #1589, opened 2025-09-10
Optimize load time and memory consumption by reusing element descriptions

azeey:
# 🦟 Bug fix

Toward https://github.com/gazebosim/jetty_demo/issues/3

## Summary
As far as I can tell, element descriptions are not mutated, so there is no reason to clone them. But I'm not 100% sure if this has unintended consequences.

I created a small test where a large sdf is loaded to show the before and after memory consumption results:
**Before**
```bash
$ time build/sdf_memory 3k_shapes.sdf
Successfully loaded file
8647.16MB
build/sdf_memory 3k_shapes.sdf  6.77s user 3.13s system 99% cpu 9.904 total
```

**After**
```bash
$ time build/sdf_memory 3k_shapes.sdf
Successfully loaded file
289.069MB
build/sdf_memory 3k_shapes.sdf  0.40s user 0.19s system 97% cpu 0.611 total
```

This is a significant speedup and reduction in memory usage, so it would be worth figuring out if we can truly do this without any negative consequences.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Generated-by: Remove this if GenAI was not used.

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

PR #1664, opened 2026-06-11
Backport load optimization without deprecation

scpeters:
This is a partial backport of #1589 without the deprecations.

## Original PR description:

Optimize load time and memory consumption by reusing element descriptions (#1589)

In most usages of `GetElementDescription`, the returned `sdf::ElementPtr` is not mutated, thus it is possible to implement a copy-on-write mechanism to avoid unnecessary clones of element descriptions. This PR removes clones of element description during the loading of an SDFormat file and introduces APIs to get immutable element descriptions. It also provides mutable element descriptions which clone the element description when called.

* Do not clone element descriptions
* Add MutableElementDescription and deprecate GetElementDescription
* Add tests
* Narrower use of GZ_UTILS_WARN_* macros
* Reword comments, Migration guide
* Remove duplicate line in python test
* Add / adjust //// lines

### DIFF
--- a/include/sdf/Element.hh
+++ b/include/sdf/Element.hh
@@ -20,8 +20,10 @@
 #include <any>
 #include <map>
 #include <memory>
+#include <optional>
 #include <set>
 #include <string>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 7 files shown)


ahcorde (approved):
awesome! Thank you for the backport @scpeters

scpeters:
https://github.com/Mergifyio backport sdf14

Issue #1692, opened 2026-08-07
resolvePose leaves a sub-epsilon residual when a frame is resolved to itself


samirbhattarai135:
## Environment

* OS Version: macOS 15 (arm64). The residual is observable on arm64; on amd64 the same computation appears to cancel exactly.
* Source or binary build? Source, via gz-sim10.

## Description

`resolvePose()` resolves both the frame and the resolve-to frame all the way to the graph root, then composes them:

https://github.com/gazebosim/sdformat/blob/cb1179326fab25ba4b44dfecf1f9f22bdd909346/src/FrameSemantics.cc#L1791-L1813

```cpp
Errors errors = resolvePoseRelativeToRoot(_pose, _graph, _frameVertexId);

if (_resolveToVertexId != gz::math::graph::kNullId)
{
  gz::math::Pose3d poseR;
  Errors errorsR = resolvePoseRelativeToRoot(poseR, _graph, _resolveToVertexId);
  ...
  _pose = poseR.Inverse() * _pose;
}
```

When `_frameVertexId == _resolveToVertexId`, `poseR` and `_pose` are the product of the *same* edge chain, so the last line is an inverse-times-itself round trip. That is the identity in exact arithmetic, but in doubles the quaternion inverse and multiply leave a residual on the order of 2^-56.

This is reachable from ordinary SDF. A joint axis declared without `xyz_expressed_in`:

```xml
<axis><xyz>0 0 1</xyz></axis>
```

resolves the axis frame to itself, because `JointAxis::ResolveXyz` sets `axisExpressedIn` and `resolveTo` to the same `xmlParentName` when both are empty:

https://github.com/gazebosim/sdformat/blob/main/src/JointAxis.cc#L451-L493

## Steps to reproduce

Downstream in gz-sim, `gz model -m vehicle_blue --joint` against `test/worlds/static_diff_drive_vehicle.sdf` prints:

```
  - Axis unit vector [ XYZ ]:
    [0 1.26714e-17 1]
```

where `[0 0 1]` is declared in the SDF. Verified on a GitHub `macos-15` arm64 runner:
https://github.com/samirbhattarai135/gz-sim/actions/runs/31218579236

Every other value in that output — poses, inertial matrices, masses — matched the expected values byte for byte. The same magnitude of residual is present in the pose values too; it is simply invisible there because those are printed with fixed 6-decimal formatting.

Context and discussion: https://github.com/gazebosim/gz-sim/issues/3602

## Suggested fix

[...]

PR #1693, opened 2026-08-08
Return identity directly when resolving a frame against itself

samirbhattarai135:
🎉 New feature

Part of #1692

## Summary

`resolvePose` computes the pose of a frame relative to a resolve-to frame as `poseR.Inverse() * _pose`. When both are the same vertex, that composes an edge chain with its own inverse — the identity in real arithmetic, but not guaranteed in floating point. This returns the identity directly for that case instead of deriving it.

The first `resolvePoseRelativeToRoot` call is kept, so validation and error diagnostics for the frame vertex are unchanged.

## What this changes

`resolvePose` leaves a sub-epsilon residual when a frame is resolved against itself, and this removes it. Confirmed by @scpeters on macOS arm64 by reverting the `FrameSemantics.cc` change and running the new test:

```
pose.Rot().X()  = -6.3357220245782474e-18   (wheel, wheel_joint)
xyz.Y()         =  1.2671444049156495e-17   (wheel_joint)
```

That `xyz.Y()` value is exactly what gazebosim/gz-sim#3602 prints for the joint axis.

**Correction to my earlier claim in this PR and in #1692.** I originally wrote that the test passed with and without the change. That was wrong, and the reason is worth recording: my CI runs used GitHub's `macos-latest`, which now resolves to the `macos-26-arm64` image. So the architecture was right, but the toolchain was not the one that exhibits the residual — the newer Apple clang on macOS 26 apparently contracts the quaternion multiply such that the round trip cancels exactly. Same source, same architecture, different codegen, opposite result.

Worth knowing for anyone testing this area: reproducing it needs the right compiler, not just an arm64 machine, and CI on `macos-latest` will not show it.

## Test

`FrameSemantics.resolveAgainstOwnFrameIsExact` checks the identity case at the `resolvePose` level and through `JointAxis::ResolveXyz` for an axis declared without `xyz_expressed_in`.

`EXPECT_EQ` cannot be used here: `Pose3d` and `Vector3d` compare through tolerance-based `operator==`, so a residual near `1e-17` compares equal. Components are checked directly.

The fixture uses `-1.5707`, mirroring gz-sim's `static_diff_drive_vehicle.sdf`. Note that `model_frame_relative_to_joint.sdf` uses only `0` and exact `pi/2` rotations and round trips exactly on some toolchains, so a test written against that fixture can pass whether or not the behaviour is present.

## Checklist

- [x] Added tests
- [ ] Consider updating Python bindings (as needed)
- [x] `codecheck` passed (See [contributing](https://github.com/gazebosim/sdformat/blob/main/CONTRIBUTING.md#contributing-code))
- [ ] All tests passed (See [test coverage](https://github.com/gazebosim/sdformat/blob/main/CONTRIBUTING.md#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+archived%3Afalse+user%3Agazebosim+-author%3A%40me+sort%3Aupdated-desc+) to support the maintainers

I have not run the full suite locally — I do not have a working sdformat build on this machine, so I have been relying on the fork's CI. Flagging that rather than ticking the box.

**Note:** Assisted by Claude (Anthropic), per the GenAI disclosure policy; commits carry an `Assisted-by:` trailer. All analysis and measurements above were verified against real CI runs rather than taken on trust.

### DIFF
--- a/src/FrameSemantics.cc
+++ b/src/FrameSemantics.cc
@@ -1798,6 +1798,20 @@ Errors resolvePose(gz::math::Pose3d &_pose,
   // If the resolveTo is empty, we're resolving to the Root, so we're done
   if (_resolveToVertexId != gz::math::graph::kNullId)
   {
+    // A frame resolved against itself is the identity by definition, so
+    // return it directly instead of deriving it. Computing it as
+    // poseR.Inverse() * _pose composes an edge chain with its own inverse,
+    // which is exact only in real arithmetic: in floating point the
+    // quaternion round trip can leave a residual on the order of 2^-56.
+    if (_frameVertexId == _resolveToVertexId)
+    {
+      if (errors.empty())
+      {
+        _pose = gz::math::Pose3d::Zero;
+      }
+      return errors;
+    }
+
     gz::math::Pose3d poseR;
     Errors errorsR =
         resolvePoseRelativeToRoot(poseR, _graph, _resolveToVertexId);
--- a/src/FrameSemantics_TEST.cc
+++ b/src/FrameSemantics_TEST.cc
@@ -328,7 +328,7 @@ TEST(FrameSemantics, resolveAgainstOwnFrameIsExact)
   ASSERT_TRUE(sdf::buildPoseRelativeToGraph(graph, model).empty());
   graph = graph.ChildModelScope(model->Name());
 
-  for (const std::string &frame :
+  for (const auto &frame :
        {"__model__", "chassis", "wheel", "wheel_joint"})
   {
     gz::math::Pose3d pose;
@@ -346,7 +346,7 @@ TEST(FrameSemantics, resolveAgainstOwnFrameIsExact)
 
   // The consumer that motivated this: a joint axis on a link rotated by a
   // value that is not exactly representable.
-  for (const std::string &jointName : {"wheel_joint", "wheel_joint_exact"})
+  for (const auto &jointName : {"wheel_joint", "wheel_joint_exact"})
   {
     const sdf::Joint *joint = model->JointByName(jointName);
     ASSERT_NE(nullptr, joint) << jointName;
--- a/src/FrameSemantics_TEST.cc
+++ b/src/FrameSemantics_TEST.cc
@@ -24,6 +24,8 @@
 #include "sdf/Element.hh"
 #include "sdf/Frame.hh"
 #include "sdf/Filesystem.hh"
+#include "sdf/Joint.hh"
+#include "sdf/JointAxis.hh"
 #include "sdf/Model.hh"
 #include "sdf/Root.hh"
 #include "sdf/SDFImpl.hh"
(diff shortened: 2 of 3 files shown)


scpeters:
> To be explicit: this test passes with and without the change.

I manually reverted the changes to FrameSemantics.cc from this branch and then build and ran the new test, and it fails locally for me on a macOS arm64 machine:

~~~
[ RUN      ] FrameSemantics.resolveAgainstOwnFrameIsExact
/Users/scpeters/ws/sdformat_math/src/sdformat/src/FrameSemantics_TEST.cc:338: Failure
  0.0
    Which is: 0
  pose.Pos().Y()
    Which is: 2.0816681711721685e-17
wheel
/Users/scpeters/ws/sdformat_math/src/sdformat/src/FrameSemantics_TEST.cc:342: Failure
  0.0
    Which is: 0
  pose.Rot().X()
    Which is: -6.3357220245782474e-18
wheel
/Users/scpeters/ws/sdformat_math/src/sdformat/src/FrameSemantics_TEST.cc:338: Failure
  0.0
    Which is: 0
  pose.Pos().Y()
    Which is: 2.0816681711721685e-17
wheel_joint
/Users/scpeters/ws/sdformat_math/src/sdformat/src/FrameSemantics_TEST.cc:342: Failure
  0.0
    Which is: 0
  pose.Rot().X()
    Which is: -6.3357220245782474e-18
wheel_joint
/Users/scpeters/ws/sdformat_math/src/sdformat/src/FrameSemantics_TEST.cc:360: Failure
  0.0
    Which is: 0
  xyz.Y()
    Which is: 1.2671444049156495e-17
wheel_joint
[  FAILED  ] FrameSemantics.resolveAgainstOwnFrameIsExact (23 ms)
~~~

so I don't know why you weren't able to reproduce the problem (maybe testing with amd64 instead of arm64?), but this looks good to me

samirbhattarai135:
Thank you for reverting and running it — that resolves it, and it means my "could not reproduce" note in #1692 was wrong. I have corrected the PR description.

It was not amd64, which makes it more interesting. My runs used sdformat's existing `macos.yml`, which is `runs-on: macos-latest`, and that now resolves to the `macos-26-arm64` image:

```
Image Release: .../runner-images/releases/tag/macos-26-arm64%2F20260728.0273
```

So same architecture as your machine, same source, opposite result. The remaining difference is the toolchain — the Apple clang on the macOS 26 image evidently contracts the quaternion multiply so the round trip cancels exactly, while yours does not. That also fits the original amd64-passes / arm64-fails split in gazebosim/gz-sim#3602 better than architecture alone does: it looks like FP contraction in codegen rather than anything about the ISA.

Practical consequence worth flagging: **this will not reproduce on GitHub's `macos-latest` runners**, so CI there cannot be used to verify it. Your local machine and the OSRF Jenkins arm64 builders can.

I will correct my earlier comments on #1692 and gazebosim/gz-sim#3602 accordingly — with the residual confirmed real in `resolvePose`, your original point about the trivial case stands, and the LCA change would indeed skip the calculation that induces it.

I have also pushed a benchmark on a separate branch comparing current `resolvePose` against the cost of a `LowestCommonAncestor` lookup across nesting depths, since `LowestCommonAncestor` allocates an `unordered_set` and a vector per call and I wanted the trade measured rather than assumed before attempting (2). I will open that separately once I have numbers.

scpeters:

scpeters (approved):

samirbhattarai135:
@scpeters heads up on the red `sdformat-ci-pr_any-resolute-amd64` check — it is not a test failure. All 1341 tests passed; the build went `UNSTABLE` on the compiler-warning quality gate:

```
[GNU C Compiler (gcc)] - [Total (any severity)]: <Unstable> - (Actual value: 2, Quality gate: 1.00)
```

Both warnings were mine, in the new test: I bound string literals to `const std::string &` in two range-for loops, so GCC 15 reported `-Wrange-loop-construct` for the per-iteration temporary. Clang did not warn, which is why homebrew-arm64 stayed green.

Fixed in 3081c36 by letting the loop variable deduce, matching the existing idiom at `src/parser.cc:1770`. No change to the test's behaviour or to `FrameSemantics.cc`.

scpeters:
thanks for the contribution!

scpeters:
https://github.com/Mergifyio backport sdf16

scpeters:
https://github.com/Mergifyio backport sdf15

scpeters:
https://github.com/Mergifyio backport sdf14

scpeters:
https://github.com/Mergifyio backport sdf12

PR #1699, opened 2026-08-12
Capture this explicitly in the InterfaceAPI find callback

caguero:
# 🦟 Bug fix

Part of gazebosim/gz-cmake#580

## Summary

`[=]` implicitly captured `this`, which is deprecated in C++20. The lambda only uses `this->modelDir`, so the capture is now explicit. No behavior change, and it is valid C++17, so it can land before any standard bump.

Before:
```
test/integration/interface_api.cc:151:9: warning: implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20 [-Wdeprecated]
```

## Backport Policy
- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests
- [x] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Assisted-by: Claude Opus 5

### DIFF
--- a/test/integration/interface_api.cc
+++ b/test/integration/interface_api.cc
@@ -148,7 +148,7 @@ class InterfaceAPI : public ::testing::Test
     this->modelDir = sdf::testing::TestFile("integration", "model");
 
     this->config.SetFindCallback(
-        [=](const std::string &_file)
+        [this](const std::string &_file)
         {
           return sdf::filesystem::append(modelDir, _file);
         });
--- a/test/integration/interface_api.cc
+++ b/test/integration/interface_api.cc
@@ -150,7 +150,7 @@ class InterfaceAPI : public ::testing::Test
     this->config.SetFindCallback(
         [this](const std::string &_file)
         {
-          return sdf::filesystem::append(modelDir, _file);
+          return sdf::filesystem::append(this->modelDir, _file);
         });
   }
   public: void CheckFrameSemantics(const sdf::World *world);


test/integration/interface_api.cc
```diff
@@ -148,7 +148,7 @@ class InterfaceAPI : public ::testing::Test
     this->modelDir = sdf::testing::TestFile("integration", "model");
 
     this->config.SetFindCallback(
-        [=](const std::string &_file)
+        [this](const std::string &_file)
         {
           return sdf::filesystem::append(modelDir, _file);
```
luca-della-vedova: Should we change this to `this->modelDir` as well? So it's clear at a glance why `this` is necessary here.

luca-della-vedova (approved):

test/integration/interface_api.cc
caguero: Good point, done in c2f24b87. It also matches the style used elsewhere in this fixture.

caguero: Yes! Added in [c2f24b8](https://github.com/gazebosim/sdformat/pull/1699/commits/c2f24b87e90e4fe384492c6ba2b5c15ff805e64c)

scpeters:
is there a reason not to backport to Jetty?

caguero:
> is there a reason not to backport to Jetty?

We could if we think we'd eventually use C++20 with Jetty (which perhaps we do it when compiling it as a vendor package). It's only fixing a warning when using C++20.

scpeters:
> > is there a reason not to backport to Jetty?
> 
> We could if we think we'd eventually use C++20 with Jetty (which perhaps we do it when compiling it as a vendor package). It's only fixing a warning when using C++20.

this is exactly my thinking, since the Jetty vendor packages are using c++20 with Lyrical

at any rate, it seems low risk to backport

caguero:
@mergifyio backport sdf16

PR #1658, opened 2026-05-12
Unhide warning when schema test is skipped

scpeters:
# 🦟 Bug fix

Unhides a warning added in #1655

## Summary

The schema integration test fails when run with very new versions of `xmllint`. This was discovered when adding CI with Ubuntu 26.04 in #1655, and logic to disable the test with a warning was added in that PR. Unfortunately, the warning is not currently being reported due to the order in which `gz_build_warning` is processed in `gz_configure_build` even though the test is skipped:

* [Build Status](https://build.osrfoundation.org/job/sdformat-ci-main-resolute-amd64/2/) https://build.osrfoundation.org/job/sdformat-ci-main-resolute-amd64/2/

This unhides the warning by switching to `message(WARNING "")`, in a similar manner to https://github.com/gazebosim/gz-transport/pull/565.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

Issue #1656, opened 2026-05-05
INTEGRATION_schema_test fails for new versions of xmllint


scpeters:
## Environment
* OS Version: 26.04
* Source or binary build? source build of https://github.com/gazebosim/sdformat/pull/1655/commits/76713ad762c697f928e280ed8b783b99c5afea6d from https://github.com/gazebosim/sdformat/pull/1655
  
  

## Description
* Expected behavior: INTEGRATION_schema_test should pass
* Actual behavior: INTEGRATION_schema_test fails

## Steps to reproduce

1. Install `xmllint` from `libxml2-utils` on Ubuntu or `libxml2` on brew, with a version at least 2.15.1. On Ubuntu 26.04 and homebrew, the default version is higher than this
2. Compile `sdformat` and run `ctest -VV -R INTEGRATION_schema_test`

## Output

from CI in https://github.com/gazebosim/sdformat/pull/1655: https://github.com/gazebosim/sdformat/actions/runs/25358850325/job/74487225142?pr=1655

~~~
[ RUN      ] SDFSchemaGenerator.TestDoublePendulum
  I/O warning : failed to load "http://sdformat.org/schemas/types.xsd": No such file or directory
  /github/workspace/build/sdf/1.12/root.xsd:8: element include: Schemas parser error : Element '{http://www.w3.org/2001/XMLSchema}include': Failed to load the document 'http://sdformat.org/schemas/types.xsd' for inclusion.
  WXS schema /github/workspace/build/sdf/1.12/root.xsd failed to compile
  CMD[xmllint --noout --schema /github/workspace/build/sdf/1.12/root.xsd /github/workspace/test/integration/model/double_pendulum.sdf]
  /github/workspace/test/integration/schema_test.cc:40: Failure
  Failed
  Fail in parsing the model
  [  FAILED  ] SDFSchemaGenerator.TestDoublePendulum (2 ms)
  [ RUN      ] SDFSchemaGenerator.TestPR2Model
  I/O warning : failed to load "http://sdformat.org/schemas/types.xsd": No such file or directory
  /github/workspace/build/sdf/1.12/root.xsd:8: element include: Schemas parser error : Element '{http://www.w3.org/2001/XMLSchema}include': Failed to load the document 'http://sdformat.org/schemas/types.xsd' for inclusion.
  WXS schema /github/workspace/build/sdf/1.12/root.xsd failed to compile
[...]

PR #1702, opened 2026-08-14
Disable schema test if xmllint is too new

scpeters:
# 🦟 Bug fix

Mitigation for #1656, manual backport of #1658 and part of https://github.com/gazebo-tooling/release-tools/issues/1503

## Summary

As noted in #1656, `INTEGRATION_schema_test` fails with versions of `xmllint` newer than 2.15. This disables the test if the `xmllint` version is too new. It also fixes the warning when `xmllint` is not found.

## Backport Policy

- [ ] This is safe to backport to the following versions:
    - [ ] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

### DIFF
--- a/test/integration/CMakeLists.txt
+++ b/test/integration/CMakeLists.txt
@@ -71,9 +71,33 @@ endif()
 
 find_program(XMLLINT_EXE xmllint)
 if (EXISTS ${XMLLINT_EXE})
-  set (tests ${tests} schema_test.cc)
+  # get xmllint version
+  execute_process(
+    COMMAND ${XMLLINT_EXE} --version
+    ERROR_VARIABLE XMLLINT_VERSION_OUTPUT
+  )
+  # typical output: "/usr/bin/xmllint: using libxml version 20913\ncompiled with: ..."
+  # version is formatted as a single integer, with 20913 representing 2.9.13
+  string(REGEX MATCH "using libxml version [0-9]+" XMLLINT_VERSION_STRING "${XMLLINT_VERSION_OUTPUT}")
+  string(REPLACE "using libxml version " "" XMLLINT_VERSION "${XMLLINT_VERSION_STRING}")
+  if("${XMLLINT_VERSION}" STREQUAL "")
+    message(WARNING "Unable to identify xmllint version. schema_test won't be run")
+  elseif("${XMLLINT_VERSION}" LESS 10000)
+    message(WARNING "Old version of xmllint (${XMLLINT_VERSION_STRING}) detected. schema_test won't be run")
+  elseif("${XMLLINT_VERSION}" LESS 21500)
+    # schema test is broken with very new versions of xmllint
+    # https://github.com/gazebosim/sdformat/issues/1656
+    # enable schema test if xmllint is older than 2.15.0
+    set (tests ${tests} schema_test.cc)
+  else()
+    # TODO: convert to WARNING when https://github.com/gazebosim/sdformat/issues/1656
+    # is resolved on supported platforms. It currently fails on Ubuntu 26.04
+    message(STATUS  "WARNING: xmllint version (${XMLLINT_VERSION}) is too new; "
+                    "schema_test won't be run. "
+                    "See https://github.com/gazebosim/sdformat/issues/1656")
+  endif()
 else()
-  gz_build_warning("xmllint not found. schema_test won't be run")
+  message(WARNING "xmllint not found. schema_test won't be run")
 endif()
 
 gz_build_tests(TYPE ${TEST_TYPE}


test/integration/CMakeLists.txt
```diff
@@ -71,9 +71,33 @@ endif()
...
+  )
+  # typical output: "/usr/bin/xmllint: using libxml version 20913\ncompiled with: ..."
+  # version is formatted as a single integer, with 20913 representing 2.9.13
+  string(REGEX MATCH "using libxml version [0-9]+" XMLLINT_VERSION_STRING "${XMLLINT_VERSION_OUTPUT}")
+  string(REPLACE "using libxml version " "" XMLLINT_VERSION "${XMLLINT_VERSION_STRING}")
+  if("${XMLLINT_VERSION}" STREQUAL "")
+    message(WARNING "Unable to identify xmllint version. schema_test won't be run")
+  elseif("${XMLLINT_VERSION}" LESS 10000)
+    message(WARNING "Old version of xmllint (${XMLLINT_VERSION_STRING}) detected. schema_test won't be run")
+  elseif("${XMLLINT_VERSION}" LESS 21500)
```
caguero, line 87: Apparently there are versions (2.12-2.14 compiled without HTTP). We could potential add another check before the `if`s:
```
string(REGEX MATCH "compiled with:[^\n]* HTTP[ \n]" XMLLINT_HAS_HTTP "${XMLLINT_VERSION_OUTPUT}")
```
And verify it has HTTP support before the 21500 check:
```
elseif("${XMLLINT_HAS_HTTP}" STREQUAL "")
  message(WARNING "xmllint built without HTTP support detected. schema_test won't be run")
elseif("${XMLLINT_VERSION}" LESS 21500)
  set (tests ${tests} schema_test.cc)
```

test/integration/CMakeLists.txt
```diff
@@ -71,9 +71,33 @@ endif()
...
+    # schema test is broken with very new versions of xmllint
+    # https://github.com/gazebosim/sdformat/issues/1656
+    # enable schema test if xmllint is older than 2.15.0
+    set (tests ${tests} schema_test.cc)
```
caguero, line 91: This branch checks the xmllint found by find_program, but `schema_test.cc` runs plain `xmllint` from PATH, so the test can potentially execute a different binary than the one that was version checked.

test/integration/CMakeLists.txt
```diff
@@ -71,9 +71,33 @@ endif()
...
+  else()
+    # TODO: convert to WARNING when https://github.com/gazebosim/sdformat/issues/1656
+    # is resolved on supported platforms. It currently fails on Ubuntu 26.04
+    message(STATUS  "WARNING: xmllint version (${XMLLINT_VERSION}) is too new; "
```
caguero, line 95: This is the only skip path that fires on Ubuntu Resolute, and STATUS output is hidden from the CMake warning stream, so the tests disappear there with no visible signal, unlike the sibling paths that use message(WARNING).

test/integration/CMakeLists.txt
```diff
@@ -71,9 +71,33 @@ endif()
 
 find_program(XMLLINT_EXE xmllint)
 if (EXISTS ${XMLLINT_EXE})
-  set (tests ${tests} schema_test.cc)
+  # get xmllint version
+  execute_process(
```
caguero, line 75: Nit: Consider adding RESULT_VARIABLE, OUTPUT_QUIET, and a TIMEOUT here so a broken `xmllint` does not create an unexpected configuration.

test/integration/CMakeLists.txt
```diff
@@ -71,9 +71,33 @@ endif()
...
+    COMMAND ${XMLLINT_EXE} --version
+    ERROR_VARIABLE XMLLINT_VERSION_OUTPUT
+  )
+  # typical output: "/usr/bin/xmllint: using libxml version 20913\ncompiled with: ..."
+  # version is formatted as a single integer, with 20913 representing 2.9.13
+  string(REGEX MATCH "using libxml version [0-9]+" XMLLINT_VERSION_STRING "${XMLLINT_VERSION_OUTPUT}")
+  string(REPLACE "using libxml version " "" XMLLINT_VERSION "${XMLLINT_VERSION_STRING}")
+  if("${XMLLINT_VERSION}" STREQUAL "")
+    message(WARNING "Unable to identify xmllint version. schema_test won't be run")
+  elseif("${XMLLINT_VERSION}" LESS 10000)
```
caguero, line 85: Do we need this branch? It looks unreachable, since every libxml2 reports at least 20000 and even the last libxml1 release encodes as 10817.

test/integration/CMakeLists.txt
caguero, line 87: Consider adding a cache variable to force enable schema_test, so developers can still run it when validating a fix for issue #1656 or once all supported platforms ship xmllint 2.15 or newer.

test/integration/CMakeLists.txt
```diff
@@ -71,9 +71,33 @@ endif()
...
+    message(WARNING "Old version of xmllint (${XMLLINT_VERSION_STRING}) detected. schema_test won't be run")
```
caguero, line 86: Nit: Should we use XMLLINT_VERSION instead?

test/integration/CMakeLists.txt
scpeters: ah, nice catch. I didn't notice that HTTP support was explicitly removed in 2.15.0 but that we can detect it directly by looking for `compiled with:.* HTTP ` instead of the version. I will update the logic in `main` and then backport

test/integration/CMakeLists.txt
scpeters: please read the comment above this line

test/integration/CMakeLists.txt
scpeters: I added this logic in response to your comment in https://github.com/gazebosim/sdformat/pull/1658#discussion_r3778952690

test/integration/CMakeLists.txt
scpeters: > HTTP support was explicitly removed in 2.15.0

reference:

https://github.com/GNOME/libxml2/blob/v2.14.6/xmllint.c#L2723-L2724

https://github.com/GNOME/libxml2/blob/v2.15.0/xmllint.c#L2285

test/integration/CMakeLists.txt
scpeters: > please read the comment above this line

I'll rephrase since I think my tone could have been kinder:

using `STATUS` is intentional, as the check currently fails on 26.04, so using `STATUS` instead of `WARNING` is suppressing the failure to allow CI to pass, while still providing some notice in the console log

scpeters:
gemini suggested how to fix this test using xml catalogs, and @azeey suggested that we just disable this test for now, so I will close this and open a simpler disabling PR

scpeters:
> gemini suggested how to fix this test using xml catalogs, and @azeey suggested that we just disable this test for now, so I will close this and open a simpler disabling PR

I meant to say these are two reasons to not put more effort into cleaning up this PR

PR #1704, opened 2026-09-01
Install types.xsd for all spec versions

scpeters:
# 🦟 Bug fix

Part of #1656 

## Summary

Gemini noticed that the `types.xsd` schema files that are present in each sdf spec folder (such as [sdf/1.12/schema/types.xsd](https://github.com/gazebosim/sdformat/blob/main/sdf/1.12/schema/types.xsd)) are not installed alongside the other `.sdf` and `.xsd` files. This installs all of those files.

Gemini also noticed that the `.xsd` files generated by the xmlschema.py script do not have a dependency declared on that script. This adds that dependency to ensure that all `.xsd` files will be regenerated when the xmlschema.py script is modified.

## Backport Policy

- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

There will be conflicts when backporting to collections without SDFormat 1.12

## Checklist
- [ ] Added tests
- [x] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Assisted-by: Gemini 3.6 Flash

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

### DIFF
--- a/sdf/1.12/CMakeLists.txt
+++ b/sdf/1.12/CMakeLists.txt
@@ -78,10 +78,21 @@ foreach(FIL ${sdfs})
     OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${FIL_WE}.xsd"
     COMMAND ${Python3_EXECUTABLE} ${CMAKE_SOURCE_DIR}/tools/xmlschema.py
     ARGS --sdf-dir ${CMAKE_CURRENT_SOURCE_DIR} --input-file ${ABS_FIL} --output-dir ${CMAKE_CURRENT_BINARY_DIR}
-    DEPENDS ${ABS_FIL}
+    DEPENDS ${ABS_FIL} ${CMAKE_SOURCE_DIR}/tools/xmlschema.py
     COMMENT "Running xml schema compiler on ${FIL}"
     VERBATIM)
 endforeach()
+
+add_custom_command(
+  OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/types.xsd"
+  COMMAND ${CMAKE_COMMAND} -E copy
+    "${CMAKE_CURRENT_SOURCE_DIR}/schema/types.xsd"
+    "${CMAKE_CURRENT_BINARY_DIR}/types.xsd"
+  DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/schema/types.xsd"
+  COMMENT "Copying types.xsd to binary dir"
+  VERBATIM)
+
+list(APPEND SDF_SCHEMA "${CMAKE_CURRENT_BINARY_DIR}/types.xsd")
 
 add_custom_target(schema1_12 ALL DEPENDS ${SDF_SCHEMA})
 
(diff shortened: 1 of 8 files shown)


sdf/1.12/CMakeLists.txt
```diff
@@ -78,11 +78,22 @@ foreach(FIL ${sdfs})
...
 endforeach()
 
+add_custom_command(
+  OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/types.xsd"
+  COMMAND ${CMAKE_COMMAND} -E copy
+    "${CMAKE_CURRENT_SOURCE_DIR}/schema/types.xsd"
+    "${CMAKE_CURRENT_BINARY_DIR}/types.xsd"
+  DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/schema/types.xsd"
+  COMMENT "Copying types.xsd to binary dir"
+  VERBATIM)
```
caguero: Could this be `configure_file(... COPYONLY)`? It does the same copy in one line, and CMake tracks the source file for us.

Also, perhaps it's worth a comment about why we copy into the build dir. As far as I can tell it's for the catalog in #1705. Otherwise, installing straight from `schema/` should be enough.

sdf/1.12/CMakeLists.txt
```diff
@@ -78,11 +78,22 @@ foreach(FIL ${sdfs})
...
+
+list(APPEND SDF_SCHEMA "${CMAKE_CURRENT_BINARY_DIR}/types.xsd")
```
caguero: I think that the generated schemas still load `types.xsd` from `http://sdformat.org/schemas/`, so the installed copy is only used if a catalog redirects that URL. The catalog in #1705 points at the build dir, so installed schemas still need network access. Maybe a follow-up to install a catalog as well?

sdf/1.5/CMakeLists.txt
```diff
@@ -69,11 +69,22 @@ foreach(FIL ${sdfs})
...
 endforeach()
 
+add_custom_command(
+  OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/types.xsd"
+  COMMAND ${CMAKE_COMMAND} -E copy
+    "${CMAKE_CURRENT_SOURCE_DIR}/schema/types.xsd"
+    "${CMAKE_CURRENT_BINARY_DIR}/types.xsd"
+  DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/schema/types.xsd"
+  COMMENT "Copying types.xsd to binary dir"
+  VERBATIM)
```
caguero: Optional but this block is now repeated in many files, next to the generation loop that was already repeated. Consider a small shared helper CMake function.

sdf/1.12/CMakeLists.txt
scpeters: I will look at installing a catalog. I do think it's worth installing the `types.xsd` for completeness even if it is not directly used without a catalog

